PR #49, opened 2024-02-24
Increase default limit for recentlyAddedIntake to 100.

pektinasen:
We are using the app quite intensively and would love to have a quick solution for not having our recents disappear after a day or two.

In the next week I can try to change it to have the list reload more items when it reaches the bottom. This would also be useful for the other lists. But this PR would already help us so much.


KayleeTheMech:
Thank you sooo much! I have it on my phone and it does the trick <3

PR #83, opened 2024-05-26
Add kcal sum for each meal

s6joui:
For me it's useful to see the sum of calories for each meal. 
I added the sum to the right side of each section.
If you think this can be a good addition to the project feel free to merge :)


simonoppowa:
Great addition. Will be part of the new update 👍

Issue #36, opened 2024-01-21
Feature: Make entries editable


nepomuc:
The user can only delete an entry but not edit it. Would be very good UX though.

Sinnaj94:
I added a functionality to change the amount of an entry by tapping on it in pull request #93.

Monoxide8969:
I would like to see this implemented too

rdy2go:
Should be closed, like mentioned in #93.

PR #93, opened 2024-06-11
Feature: Make Entries Editable (#36)

Sinnaj94:
### Description
This pull request makes intake entries editable. It is possible to set the amount (g/ml) of an intake entry by clicking on it.

### Changes
- Opens dialog when simple-tapping on an intake entry in home menu
- Updates the entry in the database layer accordingly
- Only allow users to enter numbers on intake amount

### Related Issue
Closes feature request #36.

### Additional Notes
In method updateIntake, I used a map<string, dynamic> as a data structure. However it might be better to use a DTO here, see comments. What do you think?


simonoppowa:
Really cool addition 👍 Thank you so much for the contribution to the project 🙂

PR #122, opened 2024-10-08
feat: importing meals via CSV + local barcode search

dd-dreams:
This PR enables users to add meals via a CSV file. It adds a new button in the settings page "Import CSV" - when pressed, the user selects a file from the file picker, afterwards if the file is valid, all the meals present there will be added to the database under "Products".

The CSV format should have the following columns (not necessarily in the same order of keys):
```csv
name,protein,total_sugars,saturated_fat,total_fat,food_energy,carbohydrates,fiber,barcode
```
These names are necessary for the import to work.

This PR also adds a new field to the database: "barcode" for searching locally with barcodes.

Please suggest any changes if needed.


volodymyrzuyev:
Was just about to start work on a similar commit. Please merge into main)

jordan-lee-code:
Thanks for opening this — sorry it sat unmerged for so long. Closing as superseded by PR #351, which builds CSV meal import as a fresh implementation against current \`develop\`.

Why a fresh implementation rather than picking up this branch:
- \`maintainerCanModify\` is false on this PR, so I couldn't push the necessary cleanup commits
- The branch is from October 2024 and predates the \`hive\` → \`hive_ce\` migration, the dedicated \`CustomMealDataSource\` (which now lives where new custom meals belong), the macros / activity-tracking toggles, and a few other refactors — the diff was severely conflicted

What was kept in spirit from your design:
- CSV header row required, columns name / barcode / nutrient values per 100 g
- Local barcode storage so subsequent scans resolve instantly

What's different in #351:
- Uses the existing \`CustomMealDataSource\` instead of a new DBO type, so imports show up in the user's custom-meal flow (recently-added, search results, edit, delete)
- Lives in the existing Settings → Export / Import data dialog rather than a standalone settings entry
- Adds a 'Sample CSV' button so users can see the expected format
- Adds a contribution nudge to Open Food Facts when imported rows have barcodes
- Makes the barcode scanner check custom meals first before hitting OFF
- All 8 locales covered, full test suite for the parser

Credit appreciated — the use case here originates with this PR.

PR #124, opened 2024-10-28
add custom carb, fat & protein goals

### LINKED ISSUE #123: [Feature Request] Let Users define exact goals
Hi, I really love the app! But it is not usable for me because I want to have complete control over my goals. Could a feature that lets me set a calorie count and makro split by myself?

schumischumi:
Fixes #123 
tests are still missing.
I'm no dart dev so please review very critically :-)


schumischumi:
I broke deleteUserActivityItem because I cant fetch the userEntity there

uselessgames:
can you give an apk to side install? i can help test these changes.
i dont have the ability to build it myself.

uselessgames:
@schumischumi I've been able to test your changes and have some feedback. 

i would have expected the custom carb/fat/protein to be under the calculations (top right gear > calculations). it just seems the natural place for it. There I would expect to choose from a list of standards or enter my custom ratios.

but found your proposed custom ratios, under the profile ? (for now, i accept and continue)

Then I went to edit them and found the method a bit cumbersome. 

1. the dialogue box when editing shows the value then "cm", eg `20 cm` what is _cm_? I would think it shows %, or pc, or better just nothing. 
2. the horizontal bar is not ideal. Would be best to have a int field that the user enter a whole number. (perhaps people would want other than half percentages, so floats i suppose is okey too). of course the user would just see numbers `__`% as they type.
3. You should also check that the sum of all three is 100%, so editing the three should be done in one go, and not one by one. (i made the mistake to enter 10/50/30 and was confused when the home screen targets looked odd)

(by test I mean use the app, not review code)

schumischumi:
thanks for the response and sry it took me a while. I will be blocked by private matters the coming 1.5 weeks but I hope after that I can keep on working on this.
Of course everyone can use my stuff as base (or if too bad can start over) and so another MR, if it is needed faster.
watch out because I currently broke lib/features/diary/presentation/bloc/calendar_day_bloc.dart because I don't know how to get the user entity in this function. If anybody knows how to improve this, please leave a comment / review.

> but found your proposed custom ratios, under the profile ? (for now, i accept and continue)

true. I put it into the profile because the other values that will be configured during the initial setup are also there. so I was able to do a quick copy & past.
thats also the reason for the points 1 & 2. Point 3 I'm not sure if this is a must have. Is it really important that it adds up do 100%? 

I uploaded an apk (locally built debug apk) and scanned it. but please check for yourself and install it on your own risk
https://metadefender.com/results/file/bzI0MTExMFpwOHc3Q0JIeFNDbjhJdmJaN1J2

https://file.io/IkRJTJt5OWTT

muXxer:
As far as I can see, there is now a way in the app (0.8.0) to change the calculations via sliders in the settings.
It's also possible to reduce the typical daily calorie intake. I guess this PR can be closed then? @schumischumi

jordan-lee-code:
Thanks for the contribution! This functionality was independently implemented in commit `6b23f934` (Jan 2025) and is now part of `develop`: custom carb/fat/protein percentage goals are configurable via Settings → Calculations, with an auto-balancing 3-slider UI and the percentages stored on `ConfigDBO` (`userCarbGoalPct` / `userFatGoalPct` / `userProteinGoalPct`). Closing as superseded — sorry the PR sat for so long without reaching merge.

PR #130, opened 2024-12-18
Feature: Copy meals to today from Diary

doktormerlin:
Hi!

I've created a new feature for this app which allows you to copy `IntakeEntity` objects from one day (which is not today) to today.

This helps greatly when cooking on one day for dinner and taking the leftovers on the next day.

To accomplish the task, I created the following flow:

* When in the diary and the selected day is TODAY everything is as always
* When in the diary on another day and you press and hold a food item, there is a new dialog opening from which you can select:
    * Do you want to copy to today?
    * Or do you want to delete?
* If selecting delete, the already known dialog opens
* If selecting "Copy to today", a new dialog opens on which you can select the meal you want to copy to, since you often want to change e.g. dinner to lunch.

For now I only added this functionality to meals, not to activities. Activities might also be useful if you want to track e.g. your daily commute by bike. I might add that later.

Here are some screenshots of the new dialog:

The new dialog asking what you want to do

The new dialog to copy meal to

The dropdown menu expanded for meal selection


simonoppowa (commented):
Great Addition, thank you

Issue #128, opened 2024-12-03
Update App dependencies


viniciuscr:
I tried to build the app, but I got a lot of issues from flutter to java to dart and gradel and I could not find my way out in this messy android ecosystem.

Could you update it to the most recent flutter version?

BTW It's a great app! keep on with it!
Thanks!

simonoppowa:
All App dependencies are now updated in newest release. The App is now build with Flutter 3.27.1.

PR #131, opened 2024-12-31
Features: Added ability to copy complete meal to today. Added ability to delete all intakes from a meal

doktormerlin:
This is an extension for the PR merged 2 weeks ago.

I added a popup-menu to the `IntakeVerticalList` which shows only up when `Intake` items are added to the list.

IntakeVerticalList with Popup Menu

If the selected day is today, or the list is shown from the homepage, this popup menu only shows the option to delete all intakes from this day:

Popup Menu with "Delete All" option

On other days, this menu also shows the option to copy all intakes from this meal to another day.

Popup Menu with "Copy to today" and "Delete All" options

This makes it easier to copy the lunch from yesterday to the e.g. dinner today, in case of leftovers. It is also possible that you eat the same breakfast everyday, which will also be easier with this new feature

In an upcoming pull-request I will also add the feature to add a scaling factor to this copy feature, so that it is easier to scale the sizes when e.g. only eating half the portion size for lunch than for dinner.


madmini:
@doktormerlin the pipeline flutter version has been updated in the meantime https://github.com/simonoppowa/OpenNutriTracker/issues/128#issuecomment-2585845098, so if you re-run the pipeline, it should work

Tarasovych:
Amazing! I wish this feature will be merged!

Issue #187, opened 2025-04-28
When using the "Back" button during initial setup, already entered data of previous page is gone


RiseT:
See title. Data like height, weight, age.

JEdits20:
#206

jordan-lee-code:
Hi @RiseT — thanks for the bug report!

Closing as fixed — onboarding form state now persists across back-navigation. `lib/features/onboarding/presentation/widgets/onboarding_first_page_body.dart:36-42` restores fields via `initialGender` and `initialBirthday` parameters, and `OnboardingScreen._getPageViewModels()` (line 101) passes the current `selection.*` values to all page bodies so input survives `pageController.previousPage()`. Reworked across PRs #218 / #311 / #322.

These changes are on `develop` and will reach the published app at the next release. Closing as resolved — please reopen with iOS / Android version + steps if you can still lose data on back-button navigation in a current release.

PR #206, opened 2025-05-16
202 bug allows null height and weight values after going one page back

### LINKED ISSUE #202: Bug: Allows Null Height and Weight Values After Going one Page Back
**Bug description:**
The program doesn't allow the user to continue if they don't input values for height and weight. However, if values are inputted for both, using the arrow to return to the previous page and then coming back causes the height and weight fields to be empty, and the system allows the user to continue even without inputted values.

**Expected behavior:**
The program should not allow the user to continue without inputting values for both

**Environment:**
Web (Chrome)

**Possible solution:**
I've noticed that the program is using the same values we inputted the first time, before using the arrow to go back one page.
For me, the best solution would be ensuring that the old values still appear when the user returns to this page, or reset the values when the user goes back one page.

JEdits20:
Made all values in the onboarding persistent. This should prevent the user from entering illegal values, while also providing user-feedback using existing methods.

Closes #202.

### DIFF
--- a/lib/features/onboarding/onboarding_screen.dart
+++ b/lib/features/onboarding/onboarding_screen.dart
@@ -97,13 +97,18 @@ class _OnboardingScreenState extends State<OnboardingScreen> {
     );
   }
 
-  List<PageViewModel> _getPageViewModels() => <PageViewModel>[
+  List<PageViewModel> _getPageViewModels() {
+    final selection = _onboardingBloc.userSelection;
+    return <PageViewModel>[
         PageViewModel(
           title: S.of(context).onboardingWelcomeLabel,
           decoration: _pageDecoration,
           image: _defaultImageWidget,
-          bodyWidget:
-              OnboardingIntroPageBody(setPageContent: _setIntroPageData),
+          bodyWidget: OnboardingIntroPageBody(
+            setPageContent: _setIntroPageData,
+            initialAcceptedPolicy: _introPageButtonActive,
+            initialAcceptedDataCollection: selection.acceptDataCollection,
+          ),
           footer: HighlightButton(
             buttonLabel: S.of(context).buttonStartLabel,
             onButtonPressed: () => _scrollToPage(1),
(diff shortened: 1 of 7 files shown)


JEdits20:
#202

JEdits20:
#187

jordan-lee-code:
Cleaned up the branch and pushed the fix as two commits:

1. [`7614a6a1`](https://github.com/JEdits20/OpenNutriTracker/commit/7614a6a1) **(authored by you)** — the actual `initial*`-prop fix, applied to all 5 page bodies and wired up in `OnboardingScreen`.
2. [`19686c52`](https://github.com/JEdits20/OpenNutriTracker/commit/19686c52) (Jordan) — ten new widget tests covering restoration on every page plus default-value tests.

PR is now **7 files, +334/-20, MERGEABLE** (down from 6 files / +122/-9 with severe drift against develop).

What was kept from your original PR:
- Your structural insight: pass `initial*` constructor args from the parent screen to each page body, populating from `_onboardingBloc.userSelection`. This is the core of the fix.

What was changed against your version:
- **Second page**: your version set `_parsedHeight` from the initial value but didn't seed the `TextFormField` text — so the field would still appear blank to the user. Added `TextEditingController`s for height and weight, and seeded their `.text` in `initState` from the (locale-converted) display value. This means the user actually sees their previously-entered numbers come back.
- **Second page (focus nodes)**: your version removed `_heightFocusNode` / `_weightFocusNode` and the keyboard-next-key handler, which would have regressed PR #311 (the keyboard-focus fix that landed after this branch was forked). Kept those intact.
- **Second page (storage units)**: your version stored the imperial display value in `_parsedHeight`/`_parsedWeight` but the rest of the code (and `_onboardingBloc.userSelection.height/weight`) expects metric. Renamed the props to `initialHeightCm` / `initialWeightKg` to make the convention explicit, and convert to ft/lbs only for display.
- **First page**: format the restored birthday in `didChangeDependencies` (after a localization context is available) rather than at `initState` time, since `MaterialLocalizations.of(context)` isn't safe to call from `initState`.
- **All pages**: kept the existing dart formatting (your branch had a different line-length formatter applied, which produced a lot of pure-formatting diff noise).

What was discarded:
- ~270 stale files left over from the branch being severely behind develop (the `hive` → `hive_ce` migration, the activity-tracking & macros toggles, locale additions, etc.).

Verified locally:
```
flutter test test/features/onboarding/
00:02 +16: All tests passed!  # 6 existing + 10 new
```

Issue #202 was closed citing this PR back in July 2025, but the PR was never actually merged. With this cleanup it's now mergeable again.

PR #221, opened 2025-05-18
fix: correct base quantity labels display in meal editing screen

TomAFrench:
This PR fixes a few issues introduced in https://github.com/simonoppowa/OpenNutriTracker/pull/115, notably:

- The space was lost between the macro label and the base quantity, i.e. "kcal per100 g" instead of "kcal per 100 g"
- The default 100g was still included in some internationalisation entries which resulted in "protein per 100 gper100g"

I've updated the `.arb` files but it's not entirely clear how I should regenerate the localisations.

Running `dart run package:intl/generate_localized.dart` as suggested in the existing generated files just results in the error `Could not find file 'package:intl/generate_localized.dart'` Based on `flutter_intl` in the `pubspec.yaml` file it looks like you're using this [extension](https://marketplace.visualstudio.com/items/?itemName=localizely.flutter-intl) for intl but if I regenerate the intl files using that then it results in a few hundred new entries.

Feel free to push the updated intl files onto this branch directly but it would be nice to have a documented process for how to regenerate these.


TomAFrench:
Looks like @shoutingcatana also brought up the issue of developer documentation on this. I'd be happy to set up a CI check to ensure that the generated intl files are correctly generated if you can share some pointers on what the process is.

jordan-lee-code:
Thanks for the contribution! Every change in this PR is already in `develop`:

- **`justfile`**: already exists with equivalent commands
- **CI workflow**: already uses `just ci`
- **`analysis_options.yaml`**: already excludes `lib/generated/intl/*.dart` (lines 13–16)
- **Label spacing fix**: `edit_meal_screen.dart` already calls `.trim()` on `baseQuantityUnit` before passing it to `mealNutrientsPerQtyLabel`, so the spacing issue is handled — the labels display correctly as-is

Closing as fully superseded by the current `develop` branch.

PR #226, opened 2025-05-22
Automated Tests: On Boarding - First Page

isacostaf:
**description:**
Adds automated tests for the first page of the onboarding flow.

**coverage:**
1. Checks if the Privacy Policy checkbox is required before proceeding:
  - verifies if the widgets are visible
  - verifies if the user can go to the next page **before** checking the box
  - verifies if the user can go to the next page **after** checking the box

**device:**
Tested only on Chrome

**how to test:**
`chromedriver --port=4444
`
`flutter drive --driver=test_driver/integration_test.dart --target=test/integration_test/onboarding_policy_test.dart -d chrome`

### DIFF
--- a/test/features/onboarding/presentation/onboarding_intro_page_body_test.dart
+++ b/test/features/onboarding/presentation/onboarding_intro_page_body_test.dart
@@ -1,0 +1,139 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:opennutritracker/features/onboarding/presentation/onboarding_intro_page_body.dart';
+import 'package:opennutritracker/generated/l10n.dart';
+import 'package:package_info_plus/package_info_plus.dart';
+
+void main() {
+  setUpAll(() {
+    // The widget shows AppConst.getVersionNumber() in a FutureBuilder, which
+    // calls PackageInfo.fromPlatform(). Mock it so the widget renders in tests.
+    PackageInfo.setMockInitialValues(
+      appName: 'OpenNutriTracker',
+      packageName: 'com.example.opennutritracker',
+      version: '1.2.0',
+      buildNumber: '46',
+      buildSignature: '',
+    );
+  });
+
+  Future<void> pumpIntroPage(
+    WidgetTester tester, {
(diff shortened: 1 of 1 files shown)


jordan-lee-code:
Replaced the integration test with a widget-test rewrite that covers the same intent. Pushed [`b6b25d7a`](https://github.com/isacostaf/OpenNutriTracker/commit/b6b25d7a). PR is now **1 file, +139/-0, MERGEABLE**.

Why the rewrite:
- This project's test suite uses **`flutter test`** (widget tests), not `flutter drive` / `integration_test` with chromedriver. The original `test/integration_test/onboarding_policy_test.dart` would not run as part of `just test` or in any CI workflow we have.
- Booting the full app via `app.main()` requires Hive init, Supabase, and a real `package_info` plugin — heavyweight dependencies that make tests flaky and slow.
- The original test had `Future.delayed(const Duration(seconds: 2))` calls between actions, which is the classic anti-pattern for E2E flakiness.

Coverage preserved (and expanded): instead of the single end-to-end flow ("can't proceed without policy → check it → can proceed"), the rewrite drives `OnboardingIntroPageBody` directly with six focused widget tests:
1. Both checkboxes start unchecked
2. Tapping the policy checkbox reports `(true, false)` and visually checks
3. Tapping the data-collection checkbox reports `(false, true)`
4. Tapping both reports `(true, true)`
5. Tapping the policy checkbox twice toggles it back to unchecked
6. Tapping the surrounding `ListTile` (not just the checkbox itself) also toggles — covers the full clickable area

These give finer-grained signal: if a future change accidentally untouches the data-collection state when the policy is toggled (or vice versa), the test will localize the regression precisely instead of just saying "the flow broke somewhere".

What was discarded:
- `test/integration_test/onboarding_policy_test.dart` (replaced by widget test)
- `test_driver/integration_test.dart` (project doesn't run integration_test)
- `pubspec.yaml`/`pubspec.lock` additions for `integration_test` (not needed for widget tests)

Verified locally:
```
flutter test test/features/onboarding/presentation/onboarding_intro_page_body_test.dart
00:01 +6: All tests passed!
```

The original PR #226 is credited in the commit body as the inspiration for the coverage gap this addresses.

PR #231, opened 2025-06-03
ui: Add Carbs, Fat, Protein, overview to each meal

EthanJoyce:
Adds a macro breakdown to the UI for each meal.

Here's what it looks like:

image


jordan-lee-code:
Nice feature — the macro breakdown row in each meal section addresses #235 cleanly.

I've cleaned up the branch and it's now targeting `develop`:

- Removed the `flutter gen-l10n` output (`app_localizations*.dart`) — this project uses manually-maintained `lib/generated/` files
- Added `carbsLabelShort`, `fatLabelShort`, and `proteinLabelShort` to all 8 locale ARBs (en, de, tr, cz, it, pl, uk, zh) and the corresponding `messages_*.dart` and `l10n.dart` entries

All tests pass. ✅ Ready for review.

jordan-lee-code:
Added two follow-up commits on top of this PR:

**`feat(settings): add toggle to show/hide macro breakdown per meal`** ([a53de8d6](https://github.com/simonoppowa/OpenNutriTracker/commit/a53de8d6))

Per maintainer request: the macro breakdown is now controlled by a new "Show Meal Macros" switch in Settings (sits directly under "Show Activity Tracking", uses `Icons.bar_chart_outlined`). Defaults to **on** so existing users see the new feature without action.

- New `ConfigDBO` field 14 (`bool? showMealMacros`), wired through `ConfigEntity`, `ConfigDataSource`, `ConfigRepository`, `AddConfigUsecase`.
- `HomeBloc` and `DiaryBloc` read the value and pass it through their loaded states down to all `IntakeVerticalList` instances on Home and Diary.
- When off, the meal section header shows only `<kcal> kcal`. When on, it shows the full kcal + carbs/fat/protein line.
- l10n key `settingsShowMealMacros` added to all 8 locales (en, de, cz, it, pl, tr, uk, zh). Also filled in the previously-missing `settingsShowActivityTracking` translations for `de` and `tr`.

**`test(home): add IntakeVerticalList tests for showMealMacros toggle`** ([a1819b5d](https://github.com/simonoppowa/OpenNutriTracker/commit/a1819b5d))

Four widget tests covering the toggle:
1. Macros visible when `showMealMacros: true`
2. Only kcal visible when `showMealMacros: false`
3. Default (omitted parameter) shows macros
4. Empty intake list renders no header text

The same commit also fixes a pre-existing analyzer error in `intake_vertical_list.dart` — the popup-menu switch wasn't exhaustively matching `VerticalListPopupMenuSelections` (`onShare` and `onImport` weren't handled). They're unreachable in this widget but needed an explicit no-op to satisfy the analyzer.

**`test(onboarding): fix four broken validateHeight assertions`** ([f4783b8f](https://github.com/simonoppowa/OpenNutriTracker/commit/f4783b8f))

Unrelated cleanup so the test suite is fully green for this PR. The validateHeight tests added in #240 had four cases that were failing on `develop`:
- Cases 2/3/5 asserted the height error label appeared twice, but `OnboardingSecondPageBody` has two separate `Form` widgets (height + weight) with different error labels and the tests validate only the height `Form`. Changed `findsNWidgets(2)` → `findsOneWidget`.
- Case 6 entered `'9.6'` under metric/cm, but the field's `digitsOnly` input formatter strips the decimal, leaving `'96'` — a valid cm height that never triggers an error. Changed to `'9'` (below `minHeight=30`) so the validator actually fires.

Full local suite: **55 tests, all passing**.

Debug APK built locally for manual testing.

Issue #244, opened 2025-06-13
Accept decimal value for weight


Stormax181iq:
Would be useful

https://github.com/user-attachments/assets/f38eacd2-9d85-4bcb-bc89-bad44116b3ae

alettsy:
I agree it would be useful. I've noticed you can select a decimal value for the weight in `profile -> weight`, but it doesn't show the decimal value on your profile after you select it.

It does seem to update the BMI though.

Issue #242, opened 2025-06-13
Wrong error message


Stormax181iq:
When entering my information in the get started phase of the app, I noticed that the error message for the weight field says "wrong height" instead of "wrong weight"

![Image](https://github.com/user-attachments/assets/3e540ce6-9aad-4a1b-8bf5-95076c137761)

jordan-lee-code:
Hi @Stormax181iq — thanks for the bug report!

Re-verified on current develop: the weight field correctly shows the "Enter correct weight" string, and the height field correctly shows "Enter correct height". Validators at `lib/features/onboarding/presentation/widgets/onboarding_second_page_body.dart:269-291` use the distinct `onboardingWrongHeightLabel` and `onboardingWrongWeightLabel` keys for each field — likely fixed during the #218 onboarding-validation rework.

Closing as resolved — please reopen if you can still reproduce the wrong label appearing on either field.

Issue #200, opened 2025-05-14
Bug: App accepts Weight as 0


isacostaf:
**Bug description:**
The program allows the user to input 0 as a Weight value

**Expected behavior:**
The program should only validate the weight if it's not null or 0.
Also, in my opinion, the program should have a limit for weight inputs, since not having one can contribute to eating disorders.

**Environment:**
Web (Chrome)

**Possible solution:**
Add validation to check if the weight is different then 0.

JEdits20:
#218 implements the weight input validation

jordan-lee-code:
Fixed in #218

PR #218, opened 2025-05-17
Input validation

JEdits20:
## Summary

Fixes missing input validation on the onboarding screens (issues #199, #200, #201, #204).

- **Birthday** (page 1): date picker now restricts selection to past dates only — future birthdays can no longer be selected
- **Height** (page 2): rejects `0`, non-numeric input, and values outside 30–300 cm (1.0–9.8 ft)
- **Weight** (page 2): rejects `0`, non-numeric input, and values outside 2–640 kg (4–1411 lbs)
- Date field now displays the date in the locale format used by the picker (via `MaterialLocalizations.formatCompactDate`) rather than the hardcoded `yyyy-MM-dd` format

### New shared utilities

- `lib/core/utils/bounds/ranges_const.dart` — `Ranges` class with min/max constants and compiled `RegExp` patterns for all four units (cm, ft, kg, lbs)
- `lib/core/utils/bounds/validator.dart` — `ValueValidator` with static methods for string validation, range-clamped parsing, and date boundary helpers

### What is NOT changed

- `set_weight_dialog.dart` — already fixed by #303 with proper bounds via `ProfilePickerBounds`; this PR's original patch targeted a now-removed `StatelessWidget` version and has been dropped

## Test plan

- [x] Onboarding page 1: open date picker, verify dates from tomorrow onward are greyed out / unselectable
- [x] Onboarding page 2 (metric): type `0` in height → error label shown, Next button stays disabled
- [x] Onboarding page 2 (metric): type `0` in weight → error label shown, Next button stays disabled
- [x] Onboarding page 2 (imperial): same checks in ft / lbs
- [x] Onboarding page 2: enter a valid height + weight → Next button enables
- [x] Profile weight dialog: still opens and saves correctly (unchanged)
- [x] `fvm flutter analyze` → no issues
- [x] `flutter test` → all tests pass

Fixes #199. Fixes #200. Fixes #201. Fixes #204.

PR #256, opened 2025-07-11
Weight values

isacostaf:
# Issues Addressed
The following three related issues were worked on, enabling incremental development using TDD:

[#244](https://github.com/simonoppowa/OpenNutriTracker/issues/244): Accept decimal values with dot or comma in the weight field.

[#200](https://github.com/simonoppowa/OpenNutriTracker/issues/200): Prevent zero (0) as a valid weight input.

[#242](https://github.com/simonoppowa/OpenNutriTracker/issues/242): Fix incorrect error message shown in the weight field.

# Implemented Features
Support for decimal input using . or , in the weight field.

Validation to reject zero as an invalid weight.

Correct and specific error messages for invalid weight entries.

# TDD Cycles Summary
Cycle 1 (#242):
Test created to validate proper error message "Enter correct weight".
Code adjusted to show the correct message.

Cycle 2 (#200):
Test added to ensure 0 is rejected as input.
Validation added to block values ≤ 0 and return a proper error message.

Cycle 3 (#200):
Test created to check for a specific error message when input is 0.
Code updated to show a custom message for zero.

Cycle 4 (#244):
Test implemented to allow inputs like 70.5 or 70,5.
Support for both comma and dot decimals was added and validated.

### DIFF
--- a/lib/features/onboarding/presentation/widgets/onboarding_second_page_body.dart
+++ b/lib/features/onboarding/presentation/widgets/onboarding_second_page_body.dart
@@ -186,7 +186,7 @@ class _OnboardingSecondPageBodyState extends State<OnboardingSecondPageBody> {
     // if the user input is empty, return the error message: Enter correct weight
     // if the user input dont start with a number, return the error message: Enter correct weight
     if (value.isEmpty || !RegExp(r'^[0-9]').hasMatch(value) || value == '0') {
-      return S.of(context).onboardingWrongWeightLabel;
+      return S.of(context).onboardingWrongWeightLabelZero;
     } else {
       return null;
     }
--- a/lib/features/onboarding/presentation/widgets/onboarding_second_page_body.dart
+++ b/lib/features/onboarding/presentation/widgets/onboarding_second_page_body.dart
@@ -182,7 +182,7 @@ class _OnboardingSecondPageBodyState extends State<OnboardingSecondPageBody> {
   String? validateWeight(String? value) {
     if (value == null) return S.of(context).onboardingWrongWeightLabel;
     if (value.isEmpty || !RegExp(r'^[0-9]').hasMatch(value)) {
-      return S.of(context).onboardingWrongHeightLabel;
+      return S.of(context).onboardingWrongWeightLabel;
     } else {
       return null;
     }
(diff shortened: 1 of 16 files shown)


DevPS-513:
@isacostaf  I just downloaded this app and noticed it would be better to let the units be mixed (i.e height could be cm and weight could be lbs) what is blocking this PR?

jordan-lee-code:
Cleaned up the branch and pushed the focused fix as two commits:

1. [`9c1d5d4e`](https://github.com/isacostaf/OpenNutriTracker/commit/9c1d5d4e) (authored by you) — the actual decimal-weight formatter fix
2. [`351d2282`](https://github.com/isacostaf/OpenNutriTracker/commit/351d2282) (Jordan) — three new widget tests covering decimal-with-dot, decimal-with-comma, and zero-rejection

The PR now shows **2 changed files, +85/-4** down from 15 files / +705/-45.

What was kept:
- Your core fix: replacing the weight field's `digitsOnly` formatter with the same regex-based formatter the height field uses (`^\d+([.,]\d{0,1})?$`). This is what unblocks issue #244.

What was changed/added on top:
- The current `validateWeight` and `onChanged` use `ValueValidator.parseWeightInKg` (added by PR #218 after this branch was forked). Kept that integration but added `text.replaceAll(',', '.')` so European users typing `'65,5'` parse correctly — same pattern the height field already uses.
- `keyboardType` updated to `numberWithOptions(decimal: true)` so the on-screen keyboard actually offers the decimal key on iOS/Android.
- Three widget tests added.

What was discarded:
- The new `onboardingWrongWeightLabelZero: "Weight can not be 0"` string — `ValueValidator.parseWeightInKg` already rejects zero (via `minWeight=2`) and shows the generic "Enter correct weight" error. Adding a specific zero-message would be a separate UX improvement; keeping this PR focused on #244.
- The four `test/integration_test/` files under `test_driver/` — this project uses `flutter test` widget tests rather than chromedriver-based integration tests. Equivalent behavior is now covered by the three widget tests in `test/features/onboarding/`.
- ~248 unrelated files left over from being severely behind develop.

Verified locally:
```
flutter test test/features/onboarding/presentation/widgets/onboarding_second_page_body_test.dart
00:01 +10: All tests passed!
```

Issue #244 (decimal weight input) is the primary fix. Issues #200 (zero weight) and #242 (wrong error message) are already handled by the existing `ValueValidator` flow on develop, so this PR no longer needs to address them.

PR #268, opened 2025-08-12
chore: enforce consistency of translations in CI

TomAFrench:
This PR has grown out of looking to get the internationalization into a reproduce-able state. This is done by regenerating them through `intl_translation` and running them through the formatter. This is a little cumbersome to do manually so I've introduced a justfile which combines these along with enforcing that no changes exist into a single command.

This grew into throwing the most used commands into this justfile - at which point we may as well define the CI pipeline in it so we can run all the tests locally with a single command.

The final generated dart files are a bit different but 90% of the changes are double-quotes to single-quotes and removal of the trailing commas.


jordan-lee-code:
Translation consistency checks added to CI.

PR #278, opened 2025-09-23
added Ukrainian translation

Ihor-Khomenko:
Added Ukrainian language to the app


jordan-lee-code:
Ukrainian translation merged.

PR #302, opened 2026-02-22
feat(localization): add Polish language support

zrafal86:
This commit adds full Polish (pl) language support to the app.


jordan-lee-code:
Polish translation merged.

PR #303, opened 2026-02-26
Fix onboarding weight validation text and negative profile values

### LINKED ISSUE #216: Bug: It is possible to edit the weight to negative values.
**Description:**
The "Profile" tab allows users to edit their weight. However, by repeatedly decreasing the value, it is possible to select negative numbers.

**Expected behavior:**
The program should not allow negative values for weight, as this is not feasible.

**Possible solution:**
Add a minimum weight limit to the carousel in the "Profile" tab.

### LINKED ISSUE #217: Bug: It is possible to edit the height to negative values.
Description:
The "Profile" tab allows users to edit their height. However, by repeatedly decreasing the value, it becomes possible to select negative numbers.

Expected behavior:
The program should prevent users from selecting negative values for height, as this is not feasible.

Possible solution:
Add a minimum height limit to the height selection carousel in the "Profile" tab.

### LINKED ISSUE #288: Invalid text: Invalid form tooltip for weight field on signup requesting valid height
Title should be clear enough.
In english, the form field for weight mentions height when empty.

Image

LarytheLord:
## Summary
Fixes the onboarding weight validation message and prevents negative values in profile height/weight pickers.

Closes #288
Closes #216
Closes #217

## Changes
- `onboarding_second_page_body.dart`
  - Fixed `validateWeight()` to return `onboardingWrongWeightLabel` instead of `onboardingWrongHeightLabel`.
- `set_height_dialog.dart`
  - Added lower-bound clamping for picker minimum value (`>= 1`) so users cannot scroll into negative heights.
  - Added confirmation-time guard to ensure returned value respects the same lower bound.
- `set_weight_dialog.dart`
  - Added lower-bound clamping for picker minimum value (`>= 1`) so users cannot scroll into negative weights.
  - Added confirmation-time guard to ensure returned value respects the same lower bound.

## Notes
I could not run Flutter tests in this environment because `flutter`/`dart` are not installed here.

### DIFF
--- a/lib/features/onboarding/presentation/widgets/onboarding_second_page_body.dart
+++ b/lib/features/onboarding/presentation/widgets/onboarding_second_page_body.dart
@@ -197,7 +197,7 @@ class _OnboardingSecondPageBodyState extends State<OnboardingSecondPageBody> {
   String? validateWeight(String? value) {
     if (value == null) return S.of(context).onboardingWrongWeightLabel;
     if (value.isEmpty || !RegExp(r'^[0-9]').hasMatch(value)) {
-      return S.of(context).onboardingWrongHeightLabel;
+      return S.of(context).onboardingWrongWeightLabel;
     } else {
       return null;
     }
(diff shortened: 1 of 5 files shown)


LarytheLord:
Pushed a technical follow-up in commit 2e9d76d to make this PR easier to maintain and safer.

- extracted profile height/weight picker bound and clamp logic into a shared utility (profile_picker_bounds.dart)
- updated both set-height and set-weight dialogs to use the shared functions
- added focused unit tests covering min/max range behavior and lower-bound clamping (profile_picker_bounds_test.dart)

This keeps the original bug fixes intact while adding regression coverage for the boundary logic.

Issue #267, opened 2025-08-10
Bug: created meals do not show up while searching.


grejpfrtuty:
Hey, I’m trying to add a custom meal and then reuse it another time. But it’s not possible since after creating a meal, it does not show up in search, both in products and in food sections. I’m using an IOS version.

jordan-lee-code:
Fixed in #347

PR #304, opened 2026-02-27
Include custom meals in Product/Food search results

LarytheLord:
## Summary
Custom meals were only discoverable in the **Recently added** tab, because Product/Food search paths queried remote OFF/FDC sources only. This made reusing manually created meals hard, as reported in #267.

This PR includes matching custom meals from recent intake at the top of both Product and Food search results.

## Changes
- Extend `SearchProductsUseCase` to read recent intake via `GetIntakeUsecase`.
- For non-empty search strings:
  - filter recent meals to `MealSourceEntity.custom`
  - match on name/brand case-insensitively
  - prepend matches before remote OFF/FDC results
  - deduplicate by `source + code/name`
- Update DI wiring in `locator.dart` for the new use case dependency.
- Add `search_products_usecase_test.dart` covering:
  - custom meal inclusion in OFF search
  - deduplication behavior
  - no local lookup for blank queries

## Notes
I could not execute Flutter tests in this environment because `flutter` is not installed (`command not found`).

### DIFF
--- a/lib/core/utils/locator.dart
+++ b/lib/core/utils/locator.dart
@@ -170,7 +170,7 @@ Future<void> initLocator() async {
     () => AddUserUsecase(locator()),
   );
   locator.registerLazySingleton<SearchProductsUseCase>(
-    () => SearchProductsUseCase(locator()),
+    () => SearchProductsUseCase(locator(), locator()),
   );
   locator.registerLazySingleton<SearchProductByBarcodeUseCase>(
     () => SearchProductByBarcodeUseCase(locator()),
--- a/lib/features/add_meal/domain/usecase/search_products_usecase.dart
+++ b/lib/features/add_meal/domain/usecase/search_products_usecase.dart
@@ -1,10 +1,12 @@
+import 'package:opennutritracker/core/domain/usecase/get_intake_usecase.dart';
 import 'package:opennutritracker/features/add_meal/data/repository/products_repository.dart';
 import 'package:opennutritracker/features/add_meal/domain/entity/meal_entity.dart';
 
 class SearchProductsUseCase {
   final ProductsRepository _productsRepository;
+  final GetIntakeUsecase _getIntakeUsecase;
 
-  SearchProductsUseCase(this._productsRepository);
+  SearchProductsUseCase(this._productsRepository, this._getIntakeUsecase);
 
   Future<List<MealEntity>> searchOFFProductsByString(
     String searchString,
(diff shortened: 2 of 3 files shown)


LarytheLord:
Pushed a follow-up technical pass in commit 40c5232 to strengthen coverage for this search behavior:\n\n- added a test that verifies custom meals can match by brand (case-insensitive), not only by name\n- added a test that verifies deduplication still works when custom meals have no code (fallback name-based key path)\n\nThis should make the custom-meal prepend logic safer against edge cases in real intake history.

jordan-lee-code:
Cleaned up the branch and pushed three commits:

1. [`a115c96b`](https://github.com/LarytheLord/OpenNutriTracker/commit/a115c96b) **(authored by you)** — your production-code change to `SearchProductsUseCase` and DI wiring
2. [`b96e7565`](https://github.com/LarytheLord/OpenNutriTracker/commit/b96e7565) **(authored by you)** — your follow-up test cases for brand matching and no-code dedup
3. [`19b2c445`](https://github.com/LarytheLord/OpenNutriTracker/commit/19b2c445) (Jordan) — test-file rewrite (see below)

PR is now **3 files, +262/-4, MERGEABLE** (down from 183 files / +1749/-11213 of stale-branch noise).

## About the test rewrite

You noted in the PR description that you couldn't run `flutter test` locally. I ran it — the tests as originally written threw `Bad state: Cannot call \`when\` within a stub response` on every case. Root cause: `MockX extends Mock implements X` with `package:mockito` only stubs **nullable** returns out of the box. The `ProductsRepository` methods return non-nullable `Future<List<MealEntity>>`, and mockito needs either code generation (`@GenerateMocks` + build_runner) or a registered `provideDummy<T>()` to handle that.

This project doesn't use mockito codegen anywhere — the existing tests (e.g. `intake_repository_test.dart`) use plain Dart fakes. So I rewrote your test file in the same style:

- `_FakeProductsRepository` — a map from search-string to canned `MealEntity` results, throws via `noSuchMethod` for any unexercised method
- `_FakeGetIntakeUsecase` — settable `recentIntake` list + a call counter so the "no intake query for blank search" case can assert `recentIntakeCallCount == 0` instead of `verifyNever`

**All five of your test cases are preserved verbatim** — only the mocking infrastructure changed.

```
flutter test test/unit_test/search_products_usecase_test.dart
00:00 +5: All tests passed!
```

Debug APK built locally; needs manual verification of the user-facing behavior (typing a recent custom-meal name in Add Meal search → custom meal appears at top of OFF/FDC results).

PR #305, opened 2026-02-28
fix(profile): preserve decimal weight in profile display

### LINKED ISSUE #244: Accept decimal value for weight

LarytheLord:
## Summary
Fix profile weight display so decimal weights remain visible instead of always being rounded to whole numbers.

Closes #244

## What changed
- Updated `ProfileBloc.getDisplayWeight()` to format based on value precision:
  - show no decimal for whole-number weights (for example `80`)
  - show one decimal for fractional weights (for example `80.5`)
- Added a small internal formatter that first rounds to one decimal place to avoid float noise, then suppresses trailing `.0`.

## Why
Issue #244 reports that selecting a decimal weight is possible in the weight picker but profile display rounds it away. This change preserves user-selected decimal precision in the profile subtitle.

## Validation
I could not run Flutter/Dart tests in this environment because `flutter` and `dart` CLIs are not installed here.

### DIFF
--- a/lib/features/profile/presentation/bloc/profile_bloc.dart
+++ b/lib/features/profile/presentation/bloc/profile_bloc.dart
@@ -95,10 +95,19 @@ class ProfileBloc extends Bloc<ProfileEvent, ProfileState> {
 
   /// Returns the user's weight in kg or lbs based on the user's config
   String getDisplayWeight(UserEntity user, bool usesImperialUnits) {
-    if (usesImperialUnits) {
-      return UnitCalc.kgToLbs(user.weightKG).toStringAsFixed(0);
-    } else {
-      return user.weightKG.roundToDouble().toStringAsFixed(0);
+    final displayWeight = usesImperialUnits
+        ? UnitCalc.kgToLbs(user.weightKG)
+        : user.weightKG;
+
+    return _formatWeight(displayWeight);
+  }
+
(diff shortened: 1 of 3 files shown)


Elxas866:
Nice work, hope the project will rise again :zzz: 

Btw isn't it better to merge into `develop` first, instad of `main`? At least that's what I did - 2 already merged

LarytheLord:
Thanks for the suggestion, @Elxas866. Good call on branch flow. I opened this against main because it is currently the default branch, but I can retarget to develop if maintainers prefer that for feature/fix intake.\n\nI also pushed a technical follow-up commit (48abbad) to make this PR more robust:\n- extracted weight display formatting into a dedicated utility\n- added unit tests for whole numbers, fractional values, floating-point noise, and rounding behavior\n\nSo this now has both the user-facing fix and regression coverage.

LarytheLord:
Good suggestion on branch flow. I’ve now retargeted this PR to  (and aligned the other still-open PRs in this repo to  as well).

LarytheLord:
Good suggestion on branch flow. I’ve now retargeted this PR to `develop` (and aligned the other still-open PRs in this repo to `develop` as well).

PR #307, opened 2026-03-01
Fix single NULL value leading to empty results

### LINKED ISSUE #306: [Bug] Query crashes during OFF search when products have missing nutriments data
**Description**
The application crashes with a `TypeError` when performing a word search via the Open Food Facts (OFF) data source. This happens when the search result contains at least one product where the `nutriments` field is missing or `null` in the API response.

**Error Logs**
```
I/flutter ( 3062): FINE: OFFDataSource: Fetching OFF results from: https://world.openfoodfacts.org/cgi/search.pl?search_terms=schnitzel&fields=code%2Cbrands%2Cproduct_name%2Cproduct_name_en%2Cproduct_name_de%2Cproduct_name_fr%2Curl%2Cimage_url%2Cimage_front_thumb_url%2Cimage_front_url%2Cproduct_quantity%2Cquantity%2Cserving_quantity%2Cserving_size%2Cnutriments&json=true
I/flutter ( 3062): SEVERE: OFFDataSource: Exception while getting OFF word search type 'Null' is not a subtype of type 'Map<String, dynamic>' in type cast
I/flutter ( 3062): OFFDataSource: SEVERE: Exception while getting OFF word search type 'Null' is not a subtype of type 'Map<String, dynamic>' in type cast 
I/flutter ( 3062): SEVERE: ProductsBloc: type 'Null' is not a subtype of type 'Map<String, dynamic>' in type cast
I/flutter ( 3062): ProductsBloc: SEVERE: type 'Null' is not a subtype of type 'Map<String, dynamic>' in type cast 
```
**Context**
The issue was observed when searching for terms like "schnitzel". The API request used is: https://world.openfoodfacts.org/cgi/search.pl?search_terms=schnitzel&fields=code%2Cbrands%2Cproduct_name%2Cproduct_name_en%2Cproduct_name_de%2Cproduct_name_fr%2Curl%2Cimage_url%2Cimage_front_thumb_url%2Cimage_front_url%2Cproduct_quantity%2Cquantity%2Cserving_quantity%2Cserving_size%2Cnutriments&json=true

In this case, one (or maybe more) entries do not have the `nutriments` attribute, as you can see in the image below.

Image

**Root Cause Analysis**
In the current implementation of OFFProductDTO, the nutriments field is defined as non-nullable:
`final OFFProductNutrimentsDTO nutriments;`
The generated JSON serialization code attempts to cast the nutriments key from the JSON map directly to Map<String, dynamic>. If a product entry in the search results lacks this key (which I assume happens frequently in the OFF community database), the cast fails **FOR THE ENTIRE QUERY** because of attemting to cast `null`.

**Steps to Reproduce**

1. Navigate to the "Add Meal" search.
2. Search for a product that has incomplete data on Open Food Facts (e.g., "gemischter salat").
3. The ProductsBloc will catch the exception, and no results will be displayed to the user.

**Possible fix?**
Somehow make it nullable or at least don't discard all the other "clean" results, just because one element does somehow not have the `nutriments` field (wich is kinda weird cause that's like the whole purpose of the OFF database, no?).

Elxas866:
This fixes #306. Made it possible by making `OFFProductNutrimentsDTO` nullable (preventing the cast to fail) and returning empty `MealNutrimentsEntity` if `nutrients` is `null`.

### DIFF
--- a/lib/features/add_meal/data/dto/off/off_product_dto.dart
+++ b/lib/features/add_meal/data/dto/off/off_product_dto.dart
@@ -29,7 +29,7 @@ class OFFProductDTO {
   final dynamic serving_quantity; // Can either be int or String
   final String? serving_size; // E.g. 2 Tbsp (32 g)
 
-  final OFFProductNutrimentsDTO nutriments;
+  final OFFProductNutrimentsDTO? nutriments;
 
   String? getLocaleName(SupportedLanguage supportedLanguage) {
     String? localeName;
--- a/lib/features/add_meal/data/dto/off/off_product_dto.g.dart
+++ b/lib/features/add_meal/data/dto/off/off_product_dto.g.dart
@@ -24,8 +24,10 @@ OFFProductDTO _$OFFProductDTOFromJson(Map<String, dynamic> json) =>
       product_quantity: json['product_quantity'],
       serving_quantity: json['serving_quantity'],
       serving_size: json['serving_size'] as String?,
-      nutriments: OFFProductNutrimentsDTO.fromJson(
-          json['nutriments'] as Map<String, dynamic>),
+      nutriments: json['nutriments'] == null
+          ? null
+          : OFFProductNutrimentsDTO.fromJson(
+              json['nutriments'] as Map<String, dynamic>),
     );
 
 Map<String, dynamic> _$OFFProductDTOToJson(OFFProductDTO instance) =>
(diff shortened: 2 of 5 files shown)


simonoppowa:
@Elxas866 Thanks for fixing this important issue. Sorry that it took so long for merging 😅

Elxas866:
No problem at all :upside_down_face:

Issue #229, opened 2025-05-31
Can't get product list for a given text


X-Ryl669:
When search for a "Products" (not "Food" or "Recently" tab) it fails with the message "Error while fetching product data". I'd say it fails 9 times out of ten. If I retry numerous times (by clicking Retry button), it might work or fail again.

I wonder if there are too many usage of the api.nal.usda.gov service and in that case, maybe it could be possible for the user to register and enter her own API key in the app instead of hammering the service with your own API's key (process mentionned from [here](https://fdc.nal.usda.gov/api-guide)). I think the more user and success your app will have, the more likely the key's will be blocked for one hour, like they says if all of the user share the same key.

jordan-lee-code:
Mitigated in PR #338 (merged). The OFF and Supabase FDC data sources now use automatic retry with exponential backoff (up to 3 attempts: 1 s, 2 s, 4 s), which handles transient API errors and rate-limiting. If the OFF API is experiencing a prolonged outage, PR #319 (local OFF database) is being evaluated as a deeper solution. Closing as resolved.

Issue #157, opened 2025-02-18
Searching for items doesn't work


isJustCat:
Beep! This has been noted before in the Q&A section.

This one just installed the app and tried logging its food but couldn't search for any items. 

Are you aware of this? It can look into that later and see if it could contribute and help fixing this issue, as the app overall seems to meet it needs quite well with the exception of it wanting to be able to search for items to log. 

Kind regards.

jordan-lee-code:
Mitigated in PR #338 (merged). Both the OFF and Supabase FDC data sources now automatically retry failed requests with exponential backoff, which resolves failures caused by transient errors or rate limits. For persistent OFF API outages, PR #319 (local OFF database file) is being evaluated as a longer-term solution. Closing as resolved.

Issue #316, opened 2026-03-18
Feature: Add BLS (Bundeslebensmittelschlüssel) as a local food data source


babnertikof:
Add the German Federal Food Code (BLS — Bundeslebensmittelschlüssel) as a bundled local data source alongside the existing FoodDataCentral and Open Food Facts integrations. This would significantly improve food search quality for German-speaking users, as BLS covers typical German foods and portions in detail. Since the database is creative commons it can be integrated free of charge, it would only have to be named.

**Proposed Approach**

- Pre-convert the BLS XLSX file to compact JSON using a one-time Dart CLI script
- Bundle the resulting JSON as a Flutter asset (assets/data/bls.json)
- Implement BlsLocalDatasource following the same interface as existing datasources
- Map BLS nutritional fields to the existing FoodEntity model
- Register the new datasource via GetIt alongside existing ones
- No internet connection required for BLS searches

Yes this issue was partly written with AI but I am really new to contributing so i wanted to be perfectly clear.
I will try to do this myself.

jordan-lee-code:
Hi @babnertikof — confirming this is still outstanding on `develop`.

Code location: `lib/features/add_meal/data/data_sources/` has no BLS source. Your outlined approach (XLSX → JSON asset, bundled with the app) is the right shape — would just need a `BlsDataSource` that reads the JSON asset and exposes the same query API as the other data sources, plus integration into `SearchProductsUseCase`.

Labelled `enhancement`, `help wanted`, `data-source`. Contributor PRs welcome.

skyrid3r698:
Yes! Plesse add this. The current state of translations and data is lacking at best. If possible I would like to be able to have the Option in the settings available to enable and disable different data sources.

PR #319, opened 2026-04-03
Feature: Search with local OpenFoodFacts database file

manum45:
## Description
Recently the OpenFoodFacts search API often isn't working, due to rate limiting or server issues. See https://status.openfoodfacts.org/. See also #157  and #229.
A new API seems to be planned, but I couldn't find any info on a timeline.

This pull request provides the option to search in a local database file (subset of the OpenFoodFacts database) to avoid using the search API. Nutrition details of a product are then fetched via the Barcode API when a product is select in the search results list. The barcode API seems to be a lot more reliable.

The database for search is roughly 230MB and could probably be reduced further.

This feature could also be extended so that the app can be used fully locally with the full OFF database, however this is 4GB (although some reduction should be possible, e.g. unused info, other languages, ...).

Usage description is available in `Local_OpenFoodFacts_Database.md`.

## Open topics
I wanted to get some early feedback, especially on these points:
- fetching of nutrition data is currently done synchronously when a meal item card is pressed. This causes noticeable stutter. What would be the right way to fetch this data?
- I hardcoded the UI texts because I didn't figure out how to correctly run the l10n generator. It would give me error for the cz file and generate to the wrong directory. Can you give me a hint how to use it?
- I noticed that I modified a generated file, to fix an issue where the barcode API response of a product does not provide nutriments. This currently leads to an exception in `lib/features/add_meal/data/dto/off/off_product_dto.g.dart:OFFProductDTOFromJson()`. I think this is described in #306. What would be the right way to solve this?
- Users have to create their own DB with the python script described in the markdown file. Should the reduced database be hosted somewhere and downloadable within the app?

## Testing
I have tested this roughly my Android phone, and will daily drive it for the next days.
I don't have an iPhone, so I am not able to test it there.


manum45:
I think this app could be made to work fully offline.
- I created a reduced OFF database that contains all the necessary columns according to `_hasRequiredProductInfoMissing()`, the resulting file is around 600MB. Maybe a bit large, but could be worth it. Micronutrient tracking ( #237 )  would require a larger local DB though. Probably some more filtering can be done, removing products that are not relevant for a region.
- the FDC json file seems to be around 64MB
- The dataset mentioned in #316 seems to be around 14MB as an xlsx

Do you think this is a valuable feature? What would be required from your perspective?
I would be willing to invest some time for this.

Elxas866:
This could be a great opt-in feature! Not all users want/can be online just to track their food. They should be able to download the database onto their local device - doing this should only take one singular button click to ensure optimal user flow. By default, the online-db is used.

jordan-lee-code:
# PR #319 Review — Feature: Search with local OpenFoodFacts database file

## What it does

Adds an opt-in setting that lets users point the app at a local SQLite copy of the OFF database (converted from Hugging Face parquet files via a provided Python script). When enabled, food search bypasses the OFF REST API entirely and queries the local file instead. The configuration is persisted via two new `ConfigDBO` fields (`useLocalDataBase`, `localDatabaseFile`) and the data source switch happens in `ProductsBloc`.

---

## Critical

**SQL injection in `LocalOFFDataSource`**

```dart
// current
String query = "SELECT * from food WHERE product_name LIKE '%$searchString%' LIMIT $pageSize";

// fix
db.query('food', where: 'product_name LIKE ?', whereArgs: ['%$searchString%'], limit: pageSize)
```

User-supplied search input is interpolated directly into the SQL string. Must be parameterised before merge.

---

## High

**Database connections never closed**
`LocalOFFDataSource.fetchSearchWordResults` opens a `Database` and never closes it. `DataSourcesDialog.validateDatabase` does the same. Both need `await db.close()` in a `try/finally` block.

**Nutriment filter not applied to local results**
`ProductsRepository.getOFFProductsByString` filters results with `.where((p) => p.nutriments != null)`. `getLocalOFFProductsByString` does not, so local search will surface products with no nutritional data. Either apply the same filter or document the intentional difference.

**Typo in error string**
`datasources_dialog.dart` line 59: `"Required column is missing in db: $col}"` — stray `}` at the end.

---

## Medium

**Pagination is a stub**
`page` and `pageCount` in the response are hardcoded to `0` and `1`. The `LIMIT` is applied but there is no `OFFSET`, so scrolling past the first page of local results won't work. Fine to ship as a known limitation but worth noting.

**Unresolved `TODO` comments**
`DatabaseException` handling is flagged as a TODO in both `local_off_data_source.dart` and `datasources_dialog.dart` but neither file actually handles it. At minimum the TODOs should be resolved or converted to GitHub issues.

---

## Low / Observations

**Architecture is clean.** The data source selection is entirely inside `ProductsBloc` — the UI and use case layer are unaware of whether results are local or remote. Good pattern.

**`off_product_dto.g.dart` was hand-edited.** The PR initialises a default empty `OFFProductNutrimentsDTO` before the `containsKey` check to ensure nutriments is never `null`. That's intentional but the file carries a `// GENERATED CODE - DO NOT MODIFY BY HAND` header — this should only ever be produced by `build_runner`, not edited directly (see feedback section 3 for the correct fix).

**`Local_OpenFoodFacts_Database.md` is useful.** The setup doc is clear and the known-limitations section is honest. Would be worth linking it from the main README.

---

## Summary

The SQL injection is a blocker. The unclosed database handles and missing nutriment filter are high-priority fixes. Everything else is polish. The overall structure and integration into the existing architecture is solid.

---

## Feedback on specific questions

### 1. Synchronous fetch stutter on card press

The stutter comes from `MealItemCard._onItemPressed` being `async` and `await`-ing `searchProductByBarcodeUseCase.searchProductByBarcode()` before navigating — the UI hangs with no feedback while the network call runs.

The idiomatic Flutter fix is to show a loading indicator immediately on tap and navigate once the data arrives. A lightweight approach:

```dart
void _onItemPressed(BuildContext context) async {
  if (!realoadNeeded) {
    Navigator.of(context).pushNamed(...);
    return;
  }

  // Show loading overlay immediately
  showDialog(
    context: context,
    barrierDismissible: false,
    builder: (_) => const Center(child: CircularProgressIndicator()),
  );

  try {
    final entity = await searchProductByBarcodeUseCase!.searchProductByBarcode(code);
    if (context.mounted) {
      Navigator.of(context).pop(); // dismiss loader
      Navigator.of(context).pushNamed(..., arguments: entity);
    }
  } catch (e) {
    if (context.mounted) Navigator.of(context).pop();
    // show error dialog
  }
}
```

---

### 2. l10n / intl generator

The workflow is:

1. Add new string keys **only to `intl_en.arb`** — never edit other locale files by hand for new keys, as the generator derives the message catalogue from the English source.
2. Run `just run_intl` — this calls `dart run intl_translation:generate_from_arb --output-dir ./lib/generated/intl/ lib/**/*.dart ./lib/l10n/*.arb` and writes output to `lib/generated/intl/`.
3. The other locale ARB files (`intl_de.arb`, `intl_cz.arb`, etc.) only need to contain translations for keys that already exist in `intl_en.arb`. If a key is missing from a locale file the generator uses the English fallback.

**On the `intl_cz.arb` error:** This is almost certainly a JSON syntax issue in that file (trailing comma, unescaped character, or a key that doesn't exist in `intl_en.arb`). Open `lib/l10n/intl_cz.arb` and validate it as JSON — a linter like `jq . lib/l10n/intl_cz.arb` will pinpoint the offending line immediately.

**Note:** `intl_translation` was removed from `dev_dependencies` in the upstream dependency upgrade (PR #329). After rebasing onto upstream you'll need to add it back:

```yaml
dev_dependencies:
  intl_translation: ^0.20.1
```

Then `just install` and `just run_intl` should work again.

---

### 3. Correct fix for missing nutriments in `off_product_dto.g.dart`

The generated file should never be edited directly — the `// GENERATED CODE - DO NOT MODIFY BY HAND` header is the signal. The right fix is to make `nutriments` nullable in the source DTO so `build_runner` generates the correct null-safe deserialization:

**`off_product_dto.dart`** — change the field type:

```dart
// before
final OFFProductNutrimentsDTO nutriments;

// after
final OFFProductNutrimentsDTO? nutriments;
```

Then regenerate:

```sh
dart run build_runner build --delete-conflicting-outputs
```

The generated `.g.dart` will produce:

```dart
nutriments: json['nutriments'] == null
    ? null
    : OFFProductNutrimentsDTO.fromJson(json['nutriments'] as Map<String, dynamic>),
```

Downstream, `ProductsRepository.getOFFProductsByString` already filters with `.where((p) => p.nutriments != null)` so products without nutriment data are silently dropped from results — no further changes needed. This cleanly resolves issue #306 without touching generated code.

---

### 4. Should the database be hosted for in-app download?

Two reasonable options:

1. **Keep the manual approach** — document the Python script clearly (which you've already done) and treat this as a power-user feature. Acceptable given the feature is opt-in and technically oriented.
2. **Attach the pre-built file to the GitHub release** — the Python script already filters to products with complete nutriment data so the output is a reduced subset. Uploading it as a release asset keeps things simple: no infrastructure needed, users just download the file and point the app at it via the existing file picker.

An in-app download flow would be the smoothest UX but is a larger scope change — worth a separate PR if the feature gains traction.

jordan-lee-code:
Closing as superseded — thanks for the contribution and especially for putting up the design notes in \`Local_OpenFoodFacts_Database.md\`, which were useful when figuring out the alternative.

The use case (offline OFF search, fewer hits to the API) is now covered by two pieces of work that don't require users to run a Python script or manage a 230 MB file:

**Merged:** [#352](https://github.com/simonoppowa/OpenNutriTracker/pull/352) — per-user local cache of remote OFF/FDC lookups
- Every successful search and barcode lookup gets cached locally
- Cache-first ordering: cached entries appear before fresh remote results
- Items the user logs to the diary float to the top of the cached list (timestamp-driven sort)
- 90-day TTL: untouched entries age out automatically; selected items stay
- On selection, an OFF item is refreshed in the background; falls back to the cached version on rate-limit or offline
- Settings → "Clear cached items" tile with count + size
- Worst-case storage for a heavy 6-month user: a few MB

**Deferred:** [#353](https://github.com/simonoppowa/OpenNutriTracker/issues/353) — regional pre-population from a curated subset
- For when the project website is back up and being served via GitHub Pages
- An Action would weekly produce small per-region \`.json.gz\` files (~5 MB) from the OFF Hugging Face Parquet
- App downloads on demand, populates the cache box from #352
- Verified during issue triage that OFF doesn't offer per-region exports natively, so the filtering does need to happen on infrastructure we control

Issues #157 and #229 (the original OFF reliability complaints) are also closed now that the retry logic + this cache cover the common cases.

Sorry the PR sat for so long. The Python-script + file-picker UX would have helped a small set of power users; the cache approach gets most of the same benefit for everyone with no setup.

manum45:
Hi Jordan, thanks for taking on the maintenance of this project, looking forward to see it evolve.
Also thanks for looking into this pull request. Unfortunately I don't think that the caching solution alone will solve my problems with the OFF availability. When searching for new meals I expect there will still be issues, and adding recent meals was already possible in the "Recently" tab. I guess together with #353 it could work though.

jordan-lee-code:
Heya, your issue hasn't been forgotten about, I've logged a new issue #353 to host the database files and allow for download/import in app once we get the site working again from GitHub actions. I'm hoping it'll just be generated on release and users can consent to downloading additional data. The current behaviour on develop caches your searches and keeps the foods you're using in cache and returns it at the top of the list for next time.

Now I've wrangled the backlog of pull requests we can look to fix this issue properly while mitigating it as best I can immediately and for users that never want the extra data on device.

PR #354, opened 2026-05-05
Regression test pack + iOS localization & notification fixes

jordan-lee-code:
## Summary

- **+119 unit tests** locking down behaviour across the ~25 PRs merged into `develop` since the 2026-05-03 release. Test count goes from 135 → 254 (all passing, analyzer clean).
- **2 iOS bug fixes** caught during the audit:
  - `Info.plist` `CFBundleLocalizations` only listed `en` and `de` — iOS users couldn't pick `cz`, `it`, `pl`, `tr`, `uk`, or `zh` from system settings even though the strings ship with the app.
  - `AppDelegate.swift` was missing the `UNUserNotificationCenter` delegate setup that `flutter_local_notifications` requires on iOS 10+ for foreground notifications. Without it, the daily reminder would only display when the app was backgrounded.
- **1 small refactor** to extract the next-occurrence date arithmetic out of `NotificationService.scheduleDailyReminder` so it's pure-Dart and unit-testable across timezones / DST.

No production behaviour change in any other file.

## Coverage by area

| Area | Test file | New tests |
|---|---|---|
| Imperial unit round-trip (PR #348) | `unit_calc_imperial_round_trip_test.dart` | 20 |
| OFF/FDC search cache (PR #352) | `remote_search_cache_data_source_test.dart` | 18 |
| Custom meal dedup (PRs #347, #351) | `custom_meal_data_source_test.dart` | 9 |
| `withRetry` helper (PR #338) | `retry_util_test.dart` | 6 |
| Activity edit symmetry (PR #337) | `user_activity_repository_test.dart` (extended) | +7 |
| Tracked-day kcal/macro mutations | `tracked_day_data_source_test.dart` (extended) | +10 |
| Shared meal QR payload edge cases | `shared_meal_payload_compression_test.dart` (extended) | +7 |
| Calorie goal weekly-kg override | `calorie_goal_calc_test.dart` (extended) | +6 |
| Notification next-occurrence (DST-aware) | `notification_service_test.dart` | 11 |
| iOS locale drift prevention | `ios_localizations_drift_test.dart` | 1 |

## iOS-specific changes

- `ios/Runner/Info.plist`: 6 missing locales added.
- `ios/Runner/AppDelegate.swift`: 1 line + 1 import to wire up `UNUserNotificationCenter.current().delegate`.
- `test/unit_test/ios_localizations_drift_test.dart`: parses ARB filenames and Info.plist; fails fast if a future locale is added without updating Info.plist.

## Out of scope

[...]

PR #358, opened 2026-05-05
Rename Czech locale code 'cz' to BCP-47 'cs'

jordan-lee-code:
## Why

The project shipped Czech under the non-standard code `cz`. iOS uses BCP-47 / ISO 639-1 codes (`cs` for Czech) — its language picker **silently ignores `cz`**, so iOS users could never select Czech from system settings even though the strings were bundled.

This was flagged in PR #354's follow-up comment as a known edge case to verify on a Mac. While verifying, the simplest fix turned out to be just renaming to the standard code.

## What changed

Renames:
- `lib/l10n/intl_cz.arb` → `lib/l10n/intl_cs.arb`
- `lib/generated/intl/messages_cz.dart` → `lib/generated/intl/messages_cs.dart`

Internal updates:
- Generated `messages_cs.dart` `localeName` getter
- `messages_all.dart` import + switch + deferred-libraries map
- `lib/generated/l10n.dart` `supportedLocales` entry
- Settings screen language picker map key
- `ios/Runner/Info.plist` `CFBundleLocalizations`
- `CLAUDE.md` localization docs

## Backward-compat for existing users

Anyone who already had Czech selected has `'cz'` stored in their Hive `ConfigBox`. Without migration they'd silently fall back to English on the next app launch.

`ConfigDataSource.getSelectedLocale` now maps stored `'cz'` to `'cs'` on read:

```dart
Future<String?> getSelectedLocale() async {
  final config = _configBox.get(_configKey);
  final raw = config?.selectedLocale;
  if (raw == 'cz') return 'cs';
  return raw;
}
```

The Hive value gets rewritten as `cs` the next time the user touches the language setting in Settings.

## Test plan

- [x] `flutter analyze` — no issues
- [x] `flutter test` — all 254 tests pass (including the iOS locale-drift test, which now finds `cs` matching between ARB filenames and Info.plist)
- [x] `flutter build ios --simulator --debug` — clean build on a real Mac (Scaleway Apple Silicon, Xcode 26.0)
- [x] App installs and launches on iPhone 17 Pro / iOS 26 simulator
- [x] Manual: change device language to Czech on the simulator → app should now show Czech (this is the gain — previously impossible on iOS)


jordan-lee-code:
## ✅ End-to-end manual verification on real iOS hardware

Tested on Scaleway Apple Silicon Mac, Xcode 26.0, iPhone 17 Pro simulator running iOS 26:

1. Built this branch and installed on the simulator
2. Set device language to Czech via iOS Settings (`AppleLocale = cs_FR`, `AppleLanguages = [cs-FR, en-FR]`)
3. Inside the app, switched the in-app language picker from English to **System default**
4. **App correctly resolved to Czech** — UI strings rendered in Čeština

Confirms:
- iOS now recognises the Czech bundle (impossible with the old `cz` code — iOS silently dropped it)
- MaterialApp's locale-resolution chain matches OS `cs` → `Locale('cs')` in `supportedLocales` correctly
- The new `messages_cs.dart` bundle is correctly loaded by `messages_all.dart`'s switch

The backward-compat shim in `ConfigDataSource.getSelectedLocale` will quietly migrate any existing 'cz' Hive value to 'cs' on read; the in-app picker then writes back 'cs' on the next user touch.

PR #333, opened 2026-05-03
feat: add Chinese (zh) language support

jordan-lee-code:
## Summary

- Adds full Chinese (Simplified) language support — all 423 translation keys
- Clean-branch integration of community contribution from PR #300 (ADAIBLOG), rebased against current develop
- Adds 15 keys missing from the original PR: 13 QR/share-meal keys introduced after the original branch point, plus `paAmericanFootballGeneral` / `paAmericanFootballGeneralDesc`
- Wires up all integration points: `messages_all.dart`, `l10n.dart`, `supported_language.dart`, `sp_const.dart`, `sp_fdc_food_dto.dart`, `off_product_dto.dart`

Closes #300

## Test plan

- [x] `flutter analyze` — no errors (pre-existing info in `shared_meal_payload.dart` is OK)
- [x] `just check_intl` — passes
- [x] Smoke-test Chinese locale: `LANGUAGE=zh LANG=zh_CN.UTF-8 fvm flutter run -d linux`
  - App title and navigation labels render in Chinese
  - Nutrition labels (kcal, carbs, fat, protein) show Chinese
  - Meal type strings (breakfast, lunch, dinner, snack) show Chinese
  - Physical activity list shows Chinese entries
  - QR share/import flow uses Chinese strings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #331, opened 2026-05-03
docs: fix build_runner command and document stale cache workaround

jordan-lee-code:
## Summary

- Removes the `--delete-conflicting-outputs` flag from the `just build` description in `CLAUDE.md` — this flag was removed in build_runner 2.x and causes the command to fail
- Documents the `PackageNotFoundException: hive` failure that occurs on checkouts that pre-date the `hive` → `hive_ce` migration, with the fix (`rm -rf .dart_tool/build`)

## Test plan

- [x] `dart run build_runner build` runs cleanly after clearing `.dart_tool/build` (67 outputs)
- [x] `flutter test` passes (24 unit tests)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #330, opened 2026-05-03
docs: add CLAUDE.md and update GettingStarted for FVM

jordan-lee-code:
## Summary

- Adds `CLAUDE.md` — comprehensive codebase guidance for Claude Code covering commands, architecture, DI patterns, Hive boxes, food data sources, TDEE/macro calculations, and code generation
- Expands the code generation section with a per-file breakdown of what `build_runner` generates, when to regenerate, and which files are gitignored vs checked in (including the previously undocumented `lib/hive_registrar.g.dart`)
- Updates `GettingStarted.md` to replace the manual Flutter 3.27.1 download step with FVM setup instructions, so contributors automatically get the correct pinned version from `.fvmrc`

## Notes

If `dart run build_runner build` fails with `PackageNotFoundException: hive` on an older checkout, the build cache is stale from before the `hive` → `hive_ce` migration. Fix: `rm -rf .dart_tool/build && dart run build_runner build`.

## Test plan

- [x] `just build` regenerates all `.g.dart` files cleanly (67 outputs)
- [x] `flutter test` passes (24 unit tests)
- [x] FVM setup instructions tested on a fresh Windows environment

PR #314, opened 2026-03-14
feat: comprehensive micronutrient tracking (#237)

erikpt:
Closes #237

## Summary

Implements full micronutrient tracking for 17 nutrients across the lipid, mineral, and vitamin categories. Data is sourced from both OpenFoodFacts (OFF) and FoodData Central (FDC), persisted in Hive, and displayed in the meal detail screen behind a user-controlled toggle in Settings.

## Why this was needed

The existing nutriments model only tracked 7 fields: energy, carbs, fat, protein, sugars, saturated fat, and fiber. Nutrition labels routinely include much more — sodium, cholesterol, vitamins, minerals — and users have long requested this data be visible in the app.

### 17 new nutrient fields
| Category | Fields |
|---|---|
| Lipids | Monounsaturated fat, polyunsaturated fat, trans fat, cholesterol |
| Minerals | Sodium, potassium, magnesium, calcium, iron, zinc, phosphorus |
| Vitamins | Vitamin A (µg RAE), C (mg), D (µg), B6 (mg), B12 (µg), Niacin/B3 (mg) |

### Data layer
- **`MealNutrimentsDBO`** — 17 new `@HiveField`s (indices 7–23), all nullable. New fields use append-only indices so existing Hive records deserialize cleanly without migration (missing fields resolve to `null`).
- **`MealNutrimentsDBO.g.dart`** — manually updated (build_runner cannot run due to a pre-existing `intl` version conflict in the project); write count bumped from 7 → 24.
- **`ConfigDBO`** — `@HiveField(9) bool? showMicronutrients` added for the settings toggle. Null defaults to `false` (backward-compatible).
- **`ConfigDBO.g.dart`** — manually updated; write count bumped from 9 → 10.

### Domain layer
- **`MealNutrimentsEntity`** — mirrors all 17 new DBO fields plus a `hasMicronutrientData` computed getter (returns `true` if any micronutrient field is non-null).
- **`ConfigEntity`** — `showMicronutrients` field added, default `false`.
- **`AddConfigUsecase` / `ConfigRepository` / `ConfigDataSource`** — `setConfigShowMicronutrients(bool)` plumbed through the full stack.

### Data sources

**OpenFoodFacts (OFF)**
[...]

PR #361, opened 2026-05-05
Release v1.3.0+47 — Recipes, micronutrients, search cache, pl/zh

### LINKED ISSUE #143: [Feature Request]: Add HIIT Training to activities
There are a lot of Activities already listed. It would be nice to also have HITT (High Intensity Interval Training) included.

jordan-lee-code:
> @simonoppowa — this is a sizeable release rolling up everything that's landed on `develop` since `1.1.1+45` (PR #332, 2026-05-03). Once you've had a look, please **merge into `main` and cut a build for the App Store and Play Store** so users get the new Recipes tab, micronutrient tracking, Polish + Chinese localization, and the rest of the work below. Happy to walk through any section if useful.

## Summary

Release **v1.3.0+47** (from `1.1.1+45`; jumps past the standalone `1.2.0+46` bump on main from PR #334). Substantial feature release: new Recipes tab, micronutrient tracking, local search caching, Polish + Chinese localization, iOS UIScene migration, activity editing/sharing, and broad CI hardening. ~85 non-merge commits across ~176 source files (+16.6K / -1.4K LoC in `lib/`).

### Recipes tab (#356, #279)
- Dedicated **Recipes** tab — a new top-level destination, replacing Settings → Custom Meals as the primary recipe surface.
- Recipe builder combines OFF / FDC / custom-meal ingredients with aggregated nutrition.
- Recipe detail screen supports view, edit, duplicate, delete, log to diary, and **QR share/import**.
- Recipes appear inline in food search and barcode lookup.
- CSV import (denormalized: one row per ingredient, grouped by `recipe_name`) plus sample-CSV download.
- Optional `servings_count` applies per-serving values when logging.
- Backed by new Hive types `RecipeDBO` (typeId 16) and `RecipeIngredientDBO` (typeId 17). Export/import zip now includes `user_recipes.json`.

### Micronutrient tracking (#314, #237)
- 17 new fields on `MealNutrimentsDBO`: mono-/poly-/trans-unsaturated fat, cholesterol, sodium, potassium, magnesium, calcium, iron, zinc, phosphorus, vitamins A/C/D/B6/B12, niacin.
- OFF and FDC data sources extract the new nutrients (hyphenated OFF keys + FDC nutrient IDs).
- Settings toggle "Show micronutrients"; expansion tile auto-expands when populated.
- Schema is backward-compatible — new fields are appended; existing intake history reads cleanly.

### Search caching + custom-meal-first lookup (#352, #350, #351, #338, #304)
- New `RemoteSearchCacheDataSource`: 90-day TTL cache for OFF / FDC successful lookups; offline-friendly.
- Barcode resolution order: custom meals → cache → live OFF (writes back).
- Search resolution order: custom meals → recipes → recent custom-meal intakes → cached OFF/FDC results (most-recently-touched first) → fresh remote (cached on non-empty success). All five sources are merged with first-occurrence dedup, so cached entries win ties against fresh remote — keeps the result list stable across searches.
- Search retry helper (`withRetry`) for OFF and Supabase FDC.
- Custom meals surface in search even before they've been logged, and as a fallback when remote sources fail.
- New Settings tile "Clear cached items" shows count + on-disk size.

### Custom meal CSV import (#351)
- New "Import Custom Food Data" Settings entry with a 2-row sample CSV download.
- 10 columns: `name`, `brands`, `barcode`, `kcal_per_100g`, plus carbs/fat/protein/sugars/saturated_fat/fiber per 100g.
- Quoted-field tolerance, comma-as-decimal handling, dedup by barcode-or-name, skip on missing required fields.

### Activity management (#337, #339, #345, #343, #290, #195, #364)
- Edit activity duration and consistent drag-to-delete gestures.
- QR share / import for activities (`SharedActivityPayload`).
- "Copy all activities to today" from the diary section menu.
- Full 2024 Compendium integration: updated METs, 12 new activities, plus 9 missing description entries.
- High-intensity interval exercise added — moderate (code 02210) and vigorous (02214) variants from the 2024 Compendium, with names + descriptions translated across all 8 locales (#364, closes #143).

### Inclusive gender support (#80, #362)
- New **non-binary** option in onboarding, the profile screen, and the gender picker dialog — alongside the existing male / female values.
- Optional, private **hormone-profile setting** (averaged / estrogen-typical / testosterone-typical) for non-binary users who want a more accurate calorie reference. Default is the mean of the IOM-2005 male and female outputs; explicit estrogen / testosterone profiles route to the existing female / male formula. Binary-user calorie targets are unchanged byte-for-byte.
- colour palette for gender icons throughout (#5BCEFA blue, #F5A9B8 pink, #9C59D1 purple non-binary SVG). Onboarding now shows an inclusive disclaimer when non-binary is selected, pointing users to Settings → Calculations for fine-tuning.
- New Hive types: `CaloriesProfileDBO` (typeId 18). `UserDBO.caloriesProfile` is a nullable `HiveField(7)` — old user boxes load unchanged.

### Other user-facing additions
- **Quick weight widget** on home screen (#281).
- **Weekly weight rate goal** in calculations dialog with text-input alongside slider (#284, #297).
- **Diary calendar range** extended from 356 days to 5 years (#292).
- **Language picker** in Settings overrides system locale (`LocaleProvider`).
- **Macro breakdown per meal** toggle in Settings (#231) — adds carbs/fat/protein cards to each meal.
- **Edit intake amounts directly from diary** (#336).
- **Copy meal between days** now defaults to preserving the meal (#233).
- Daily meal reminders honour local timezone via `flutter_timezone` (#340).

## Localization

- **Polish (`pl`)** and **Chinese (`zh`)** locales added with full translations across all UIs (#333, #302).
- 17 new micronutrient keys, 3 short-form macro labels (`carbsLabelShort`, `fatLabelShort`, `proteinLabelShort`), CSV-import strings, cache-management strings, Recipes-tab strings, and 6 non-binary / hormone-profile strings translated into all 8 supported locales. (For inflected languages the masculine form is shipped as the default for `genderNonBinaryLabel` — native-speaker review welcome post-merge.)
- Czech locale renamed `cz` → `cs` to comply with BCP-47 (#358).
- iOS `CFBundleLocalizations` now declares every shipped locale.

## Bug fixes

- Imperial weight round-trip rounding (#348) — kg precision preserved when converting from lbs.
- FDC energy mapping in custom meal editor (#347).
- Restore Share / Import options on the meal-section popup menu (#349) — regression from 1.1.1.
- Keyboard dismissal in meal-detail quantity field (#342).
- Onboarding: allow decimal weight (#244), preserve user input across page navigation (#202), show error label on out-of-range height/weight (#218, #206), correct decimal keyboard on iOS, consistent date formatting via localization.
- Activity detail: removed accumulating listeners from `build()`.
- Home: long-press now starts drag instead of showing the delete dialog.
- Cache: preserve timestamps when bulk-caching; write intake before refreshing cache so Home reload reflects it.
- Validation cleanup in `Ranges` / `ValueValidator`; async-context guards.
- Extracted reusable `FoodNameValidator` with l10n error messages (#211, #214, #258).
- Calorie calculation correctness on `develop` (#363) — onboarding race that left the home screen computing TDEE against a hardcoded 80 kg / 180 cm / male / active dummy until a manual reload; non-binary averaged TDEE used the female PA constant in both halves of the average; the calorie-reference picker silently dropped the user's selection (RadioGroup ancestor not propagating to RadioListTile children); `CaloriesProfileDBOAdapter` was missing from `HiveDBProvider`'s registration list so hormone-profile choices reset to averaged on app relaunch; `QuickWeightWidget` saved through `AddUserUsecase` directly so the profile screen stayed stale; the weekly-rate slider locked users into a 0 kg/week override after a single OK tap. Closing all six paths required tightening every `Future<void> ... async {}` in the local-storage stack to actually await its inner Hive operation.

### iOS
- **UIScene lifecycle migration** (#359) — `SceneDelegate.swift` wired into Runner; supports iOS 13+ scene model.
- **Foreground notifications** — `UNUserNotificationCenter` delegate set so reminders display while app is foregrounded (#312).
- Podfile.lock auto-refresh on CI when constraints are stale.

### Android
- New `android-build` and `android-integration-tests` CI jobs.
- `flutter_timezone` upgraded 1.0.8 → 3.0.1 (fixes Android build, #341).

## CI / tooling

- All 5 CI jobs run in parallel; `linux-checks` no longer gates the pipeline.
- New iOS app-boot smoke test runs on simulator (#357).
- iOS pod install falls back to `pod update` when lockfile constraints are stale.
- Documentation: build_runner stale-cache workaround documented (#331); GettingStarted updated for FVM (#330); project `CLAUDE.md` added; `CONTRIBUTING.md` added covering the develop-branch policy and the 8-locale ARB + manual `lib/generated/` workflow (#366).
- Repository hygiene: `flake.nix` / `flake.lock` removed (replaced by FVM workflow).

## Test coverage

~60+ new tests added covering: micronutrient mapping, shared meal/recipe QR payloads (v1/v2 round-trip), CSV importers, food name validation, cache data source, retry helper, custom-meal dedup, calorie-goal calculation, activity repository, onboarding state restoration, copy-dialog behaviour, intake list display, imperial round-trip precision, height/weight validation, and barcode search resolution order.

## Test plan

- [ ] `just check_intl` passes (verifies generated `lib/generated/intl/*` matches ARBs)
- [ ] `flutter analyze` clean
- [ ] `just test` passes
- [ ] iOS smoke: app launches on simulator, language picker reaches Polish/Chinese, Recipes tab opens, micronutrient toggle reveals fields
- [ ] Android smoke: build APK, log a meal with cached barcode, share an activity QR, import a recipe CSV
- [ ] Migration smoke: upgrade an existing 1.1.1 install — confirm intakes, custom meals, and activities survive

## Post-merge follow-ups (@simonoppowa)

- `git tag v1.3.0 && git push origin v1.3.0`
- Build and upload signed iOS bundle to App Store Connect (TestFlight → public).
- Build and upload signed Android AAB/APK to Google Play (internal testing → production).
- Update store listings if any new screenshots are needed for Recipes / micronutrients.


Elxas866:
Incredible progress! Just checked in on the project after a while and saw this many high quality commits - I'm so happy to see this project finally kickstart right back to the spotlight! I totatlly picture this as one of the most impactful open source software alternatives! :tada:

jordan-lee-code:
**PR Review performed by Claude**

@simonoppowa, this is a substantial release and worth the careful read. The schema work is conservative in all the right places, the calorie maths is honest about its assumptions, and the test coverage targets exactly the calculation-sensitive paths I wanted to see covered. A few things below are worth addressing before you cut the build.

## What I think landed well

The Hive schema additions are append-only at every touchpoint I checked. `MealNutrimentsDBO` adds 17 new fields at `HiveField` indices 7 through 23 with all of them nullable, `UserDBO.caloriesProfile` is a nullable `HiveField(7)`, and the new `RecipeDBO`/`RecipeIngredientDBO`/`CaloriesProfileDBO` types occupy fresh typeIds (16, 17, 18). Existing user installs will read cleanly. The switch in `lib/core/utils/hive_db_provider.dart` from a hand-maintained adapter list to `Hive.registerAdapters()` from the generated registrar is the right fix for the silent-drop class of bug, and the comment there does a kind job of explaining why the old pattern hurt.

The non-binary work is thoughtful. Decoupling the PA constant from the user's stored gender via `PalCalc.getPAValueForFormula({palValue, isMaleFormula})` means the averaged TDEE feeds male PA into the male half and female PA into the female half rather than collapsing both halves through the female constant. The `_dispatch` helper in `bmr_calc.dart` keeps the binary outputs byte-for-byte identical, which is the right guarantee to make explicit. The optional `CaloriesProfileEntity` (averaged / estrogenTypical / testosteroneTypical) gives users on HRT a sensible opt-in without forcing a choice on anyone who doesn't want one.

The search cache design is clean. The 90-day TTL with a sidecar timestamp box, the distinction between `cache()` (touches timestamp, right for intent-driven writes) and `cacheFromSearch()` (preserves existing timestamps so a re-search doesn't promote stale items), and the five-source resolution order with first-occurrence dedup are all explained in code comments and exercised in tests.

The retry helper, the QR-share payload format with explicit version negotiation, and the ~60 new tests across calculation, mapping, and parser code give the release a solid floor.

## Findings worth fixing before you merge

### 1. Roughly 20 newly-added strings have no translation outside English

This one matters most for users on the seven non-English locales because the strings show up in user-facing surfaces. The PR adds 133 new keys to `intl_en.arb`. Of those, between 19 and 21 are missing from each of `cs`, `de`, `it`, `pl`, `tr`, `uk`, and `zh`. The gap shows up most visibly in the activity picker and the activity-import flow. Specifically:

- The 2024 Compendium additions: `paActiveVideoGames`, `paActiveVideoGamesDesc`, `paCrossCountrySkiing`, `paCrossCountrySkiingDesc`, `paNordicWalking`, `paPickleball`, `paPilates`, `paResistanceTrainingVigorous`, `paResistanceTrainingVigorousDesc`, `paSnowshoeing`, `paStretching`.
- The activity QR import flow: `importActivityLabel`, `importActivityConfirmTitle`, `importActivityConfirmContent`, `importActivitySuccessLabel`.
- A handful of others including `duplicateMealDialogContent`.

The HIIT additions from #364 (`paHighIntensityIntervalExercise*`) are translated correctly across all 8 locales, so the wiring works; these are just missing entries. A native-speaker pass is the proper fix, but if the goal is to unblock the release, falling back to the English string in each locale's ARB (or to a known existing label like `paOtherSports`) would at least prevent the picker showing partly translated labels.

### 2. `UNUserNotificationCenterDelegate` cast is silently optional

In `ios/Runner/AppDelegate.swift:14`:

```swift
UNUserNotificationCenter.current().delegate = self as? UNUserNotificationCenterDelegate
```

If `FlutterAppDelegate` ever stops conforming to `UNUserNotificationCenterDelegate` (a Flutter version bump, a plugin reorganisation), the `as?` cast quietly resolves to `nil`, the assignment is a no-op, and foreground notifications stop firing without any error to surface the regression. Two safer shapes:

- Add an explicit conformance: `extension AppDelegate: UNUserNotificationCenterDelegate {}` and assign with a force cast (`as!`).
- Or set the delegate inside an `if let delegate = self as? UNUserNotificationCenterDelegate` and assert / log when it's not.

Either approach makes the loss of conformance loud rather than silent.

### 3. QR-import gzip decode has no length cap

`lib/features/home/domain/entity/shared_activity_payload.dart:60-62` (and the equivalent in `shared_meal_payload.dart` / `shared_recipe_payload.dart`) does `gzip.decode(base64Url.decode(...))` with no bound on the decompressed size. A maliciously crafted QR can encode highly compressible data that expands to many MB on a phone, which is mostly an OOM / DoS risk rather than RCE, but worth a cheap guard. A `if (decompressed.length > _kMaxPayloadBytes) throw SharedActivityParseException(...)` at each call site, with a generous limit (say, 64 KB), would close it.

### 4. `RemoteSearchCacheDataSource` is doing O(N²) work on cache writes

`cache()` and `cacheFromSearch()` linear-scan `_cacheBox.values` for every meal being cached. For a 50-item search result against a 5K-entry cache, that's a quarter of a million comparisons per write. The fix is small: build a `Map<String, dynamic>` of `code → boxKey` (with a name-keyed fallback) once at the top of each method and look up in O(1). Not blocking the release, but it'll start to bite as users accumulate cache entries.

### 5. Recipe CSV importer's decimal-comma handling only works inside quoted fields

`_parseNumOrNull` swaps `,` for `.`, but `_splitRow` treats unquoted commas as field separators. A European user typing `1,5,g,...` will get a "too few columns" error rather than the intended "1.5 grams". The sample CSV happens to use whole-number quantities so it doesn't exhibit this, and the docs don't catch it. Either document explicitly that decimal-comma values must be wrapped in quotes, or pre-process the line to convert decimal commas before splitting.

## Smaller observations, non-blocking

- `lib/core/utils/supported_language.dart` only enumerates `en`, `de`, `pl`, `zh`. It's the enum used for OFF / FDC product-name lookups, so users on `cs`, `it`, `tr`, or `uk` fall through to English for product names even though the rest of the app is in their language. Worth a separate ticket; not regression-introduced by this PR.
- `Info.plist` keeps both `UIMainStoryboardFile` and `UISceneStoryboardFile` set to `Main`. With the scene manifest in place, the legacy key is redundant; iOS prefers the scene one. Harmless, removable in a follow-up.
- `SetWeeklyWeightGoalDialog` returns `double.nan` as a "clear" sentinel and a small helper `isWeeklyGoalClear` decodes it. Functionally fine, but a tiny result wrapper or a sealed class would read more cleanly than overloading NaN's meaning.
- `_iom2005MaleKcal` and `_iom2005FemaleKcal` each call `PalCalc.getPALValueFromActivityCategory` independently. For non-binary `averaged`, that's two identical lookups; computing `palValue` once at the top of `getTDEEKcalIOM2005` and threading it in would be cleaner.
- `ComputeRecipeNutritionUseCase.compute` runs roughly 48 linear passes over the ingredient list (24 nutrients times the `anyNonNull` + `total` pair). Negligible at typical recipe sizes; only worth restructuring if recipes start averaging tens of ingredients.
- The `cz` to `cs` rename is correct BCP-47, but I didn't see a one-shot migration for installs that have `cz` saved as an explicit override. If `LocaleProvider` reads back `cz` from a config box and Flutter's locale resolution doesn't find a match, the user falls back to system default rather than Czech. Worth a quick check.
- The CI fallback from `pod install` to `pod update` will resolve every pod to its latest, not just the one with the stale constraint. That's a wider surface change than necessary on every drift event. Limiting it to the failing pod (`pod update <name>`) would be a smaller hammer, though I appreciate that detecting which pod failed from `pod install` output is fiddly.

## Test plan I'd want to see green before cutting the build

- `just check_intl` and `flutter analyze` clean after the missing-translation fixes land.
- Manual smoke on each non-English locale: open the activity picker and confirm no English labels leak through.
- iOS smoke: cold start, foreground a local notification, confirm it presents (validates the `UNUserNotificationCenterDelegate` wiring).
- QR import smoke: scan a normal payload and a deliberately oversized one once the length cap is in.
- Hive migration smoke: install 1.1.1, log a meal and an activity, upgrade to this build, confirm intakes / custom meals / activities all survive and that any `cz` locale override still resolves.

Thank you for the patience that's gone into this release. The non-binary work in particular is something I think is going to mean a lot to the people it's for, and it's done with the kind of care that doesn't announce itself loudly. Happy to dig into any of the findings above in more detail if useful.

jordan-lee-code:
Merged suggested fixes into develop after local apk testing.

jordan-lee-code:
**Follow up review from Claude**

@simonoppowa, I've been back through the branch since the review fixes landed (`fix/pr361-review-followups`, merged at 4794389d), checked each follow-up commit against the original findings, and run the affected tests locally. Posting here so the audit trail sits with the original review rather than scattering across separate comments.

## Blocking findings — addressed

1. **~20 missing translations** (e34120d2). Spot-checked all 13 keys from the original review across cs / de / it / pl / tr / uk / zh — every one is now translated. The commit message notes cross-checking translator output against local Ollama models and keeping the natural-language form where the model returned something literal but unidiomatic (qwen3's `krajinné lyžování` for cross-country skiing got replaced with the natural `běh na lyžích`). `just check_intl` is clean.

2. **`UNUserNotificationCenterDelegate` silent cast** (7c440bac, then dfceb809). The optional cast is gone entirely — `UNUserNotificationCenter.current().delegate = self`, no cast at all. Loss of conformance now surfaces as a compile error instead of a runtime no-op. The second commit drops a redundant explicit conformance extension that the Swift compiler flagged once CI ran on the first attempt; the diagnostic itself is the reassurance the original `as?` cast was reaching for.

3. **QR-import gzip length cap** (7f7c0ef0). 64 KiB cap on all three payload types (`SharedActivityPayload`, `SharedMealPayload`, `SharedRecipePayload`). Guarded rethrow bypasses the plain-JSON fallback so a size violation reads as a real error rather than a malformed-input signal. New tests construct a 1 MiB all-zeros gzip blob (~1 KiB on the wire) and confirm refusal with the expected exception type.

4. **`RemoteSearchCacheDataSource` O(N²) writes** (4ac44c30). Per-call dedup index with separate by-code and by-name maps so a barcode-less entry whose name happens to match a barcode value can't collide. The index stays live within `cacheAll` / `cacheFromSearch` so a duplicate later in the same batch overwrites the earlier insertion rather than producing a second row. New test pre-populates 200 entries and asserts dedup-decision parity with the old linear-scan implementation.

5. **Recipe CSV decimal-comma handling** (eee95b1a). This is the one I'd flag as a softer fix than the others, and it's worth being explicit about. Rather than accepting bare `1,5,g,…`, the parser now surfaces a clear over-split error that tells the user to wrap the cell in double quotes (`\"1,5\"`). The fix also lifts the duplicated `_splitRow` / `_parseNumOrNull` helpers into a shared `CsvRowParser`, which is a real quality win — both flows now have a single place to evolve the format. Pre-processing decimal commas at the row level would have been ambiguous given commas are also separators, so the chosen middle ground is reasonable, but the sample CSV is now the place where the quoting requirement needs to be visible to the user. Worth confirming the in-app sample CSV download includes a quoted-decimal example so that European users discover the constraint before they hit the error.

## Smaller observations — addressed

- `SupportedLanguage` enum (501b2519) — all 8 codes wired in, with a comment at the top explaining the FDC-side fall-through and why it stays English-only for now. Honest framing.
- `Info.plist` (57c1c1a1) — `UIMainStoryboardFile` removed, only `UISceneStoryboardFile` remains.
- `SetWeeklyWeightGoalDialog` (41131c28) — sealed `WeeklyWeightGoalResult` with `Cancelled` / `Cleared` / `Set` variants. Reads more clearly at the call sites than the NaN sentinel did.
- IOM2005 PAL double-call (6c3bfb43) — PAL value computed once at the dispatch level and threaded into both helpers. The `averaged` non-binary path no longer pays for an identical lookup twice.
- `ComputeRecipeNutritionUseCase` (e1664543) — single-pass aggregation rather than 24 nutrients × `anyNonNull` + `total` traversals.
- `cz → cs` migration — `lib/core/data/data_source/config_data_source.dart:128-133` rewrites a stored `cz` to `cs` on read, so existing Czech users keep their language preference across the BCP-47 rename.

## Smaller observation deferred

- **CI `pod install` → `pod update` fallback.** Still updates every pod when one drifts. The original review flagged that detecting which pod failed from `pod install` output is fiddly, and that framing held — Jordan opted to leave it. I've filed #369 so it doesn't fall off the radar; tagged `enhancement` / `ios` / `backlog:low`, which matches the priority signalled in the original review.

## Additional fixes from local APK testing

Four commits Jordan added while smoke-testing on device that weren't in the original findings, but are worth surfacing for completeness:

- 7a652820 — settings refreshes cache count and size on every screen entry, so the user sees current numbers after a clear or a search-and-cache without having to back out and re-enter the screen.
- 4ebaed78, aa06a5ce, b2cb6de8 — three small commits tightening the QR-import scanner state machine to close a double-dialog window on the success-and-pop path.

## Verification I ran

- `flutter analyze` on the merged develop tip — no issues.
- `just check_intl` — clean.
- The six new and affected test files run 58 tests, all passing locally: `csv_row_parser_test`, `csv_meal_importer_test`, `csv_recipe_importer_test`, `remote_search_cache_data_source_test`, `shared_activity_payload_test`, `shared_meal_payload_compression_test`, `shared_recipe_payload_test`, and `set_weekly_weight_goal_dialog_test`.

## Net

Every blocking concern has a real fix with tests behind it. The smaller observations are mostly handled, the CSV decimal-comma case is a thoughtful middle ground rather than a strict accept, and the only item left on the floor is the Pods CI fallback — which is now tracked in #369 so it can be picked up on its own merits rather than rolled into a release.

The branch reads as ready for the develop → main merge.

jordan-lee-code:
**Final review from Claude**

@simonoppowa, this is the last pass before you cut the build. I've walked the head of the branch (`468d60c9`) against every concern raised in the original review, reading each affected file rather than trusting the commit messages or the previous follow-up's word, and I want to put what I actually saw on the record so you have it in one place when you come back to this.

## All five blocking findings have real fixes behind them

1. **Missing translations.** All sixteen keys flagged in the original review are present in every one of the eight locale ARBs (`en`, `cs`, `de`, `it`, `pl`, `tr`, `uk`, `zh`). I checked each one individually rather than trusting the count. The activity-import strings, the 2024 Compendium additions, and `duplicateMealDialogContent` are all there.

2. **`UNUserNotificationCenterDelegate`.** `ios/Runner/AppDelegate.swift:21` now does an unconditional `UNUserNotificationCenter.current().delegate = self` with no cast at all, and the comment immediately above it walks through the reasoning. A future loss of conformance will surface as a compile error rather than a silent runtime no-op, which is a stronger fix than what the original review asked for.

3. **QR-import gzip length cap.** All three payload classes carry a 64 KiB `_kMaxDecompressedBytes` cap (`SharedActivityPayload`, `SharedMealPayload`, `SharedRecipePayload`). Each `fromJsonString` uses a guarded rethrow so a size violation reads as a real error rather than getting swallowed by the plain-JSON fallback.

4. **`RemoteSearchCacheDataSource` O(N²) writes.** The linear scans are gone. Each `cache` / `cacheAll` / `cacheFromSearch` builds a per-call `_DedupIndex` with separate `byCode` and `byName` maps, so a barcode-less entry whose name happens to match a barcode value can't collide. The index is registered in-line in `cacheAll` so a duplicate later in the same batch overwrites the earlier insertion, which was the subtle case worth getting right.

5. **Recipe CSV decimal-comma.** This took the softer middle ground rather than trying to disambiguate `1,5,g` against `1,5` at parse time, and the chosen shape is the honest one. There's a shared `lib/core/utils/csv_row_parser.dart` extracted between the meal and recipe importers, with the quoting requirement (`"1,5"` for one point five) documented at the top of the file alongside the rationale.

## Smaller observations all genuinely addressed

`SupportedLanguage` now enumerates all eight locale codes with a comment explaining the FDC fall-through. `Info.plist` has lost the redundant `UIMainStoryboardFile`. `SetWeeklyWeightGoalDialog` returns a sealed `WeeklyWeightGoalResult` with `Cancelled` / `Cleared` / `Set` variants in place of the NaN sentinel. `getTDEEKcalIOM2005` computes `palValue` once at dispatch and threads it into both helpers. `ComputeRecipeNutritionUseCase` is a single-pass aggregation. The `cz` to `cs` migration in `config_data_source.dart:128-133` rewrites stored `cz` to `cs` on read so existing Czech users keep their language preference across the BCP-47 rename.

## A small thing the follow-up couldn't have known

The previous follow-up flagged the Pods CI fallback as deferred to #369, which was accurate at the moment it was posted. PR #370 then merged about thirteen minutes later, closes #369, and adds a small shared script at `.github/scripts/pod_install_with_targeted_fallback.sh` that parses the `pod install` failure output for the specific pod name and runs `pod update` against just that name. That's the smaller hammer the original review asked for, landing on this PR rather than getting carried into a follow-up.

## Test plan

The new and affected test files all exist (`csv_row_parser_test`, `csv_meal_importer_test`, `csv_recipe_importer_test`, `remote_search_cache_data_source_test`, `shared_activity_payload_test`, `shared_meal_payload_compression_test`, `shared_recipe_payload_test`, `set_weekly_weight_goal_dialog_test`). The current `linux-checks` and Android / iOS build jobs on this commit are the public signal on whether they all stay green; at the time of writing they were in progress.

## Where this leaves the branch

Every blocking concern has a real fix and tests behind it. Every smaller observation is genuinely addressed. The one item honestly deferred at the time of the follow-up has since been done too. The branch reads as ready for the `develop` → `main` merge whenever you're happy to cut the build.

lib/core/utils/calc/met_calc.dart
```diff
@@ -3,9 +3,9 @@ import 'package:opennutritracker/core/domain/entity/user_entity.dart';
 
 class METCalc {
   /// Calculates total kcal with formula by the
-  /// '2011 Compendium of Physical Activities'
-  /// https://pubmed.ncbi.nlm.nih.gov/21681120/
-  /// by Ainsworth et al.
+  /// '2024 Compendium of Physical Activities'
+  /// https://doi.org/10.1249/MSS.0000000000003624
```
simonoppowa: issue: seems to be wrong link. 
Should be https://pubmed.ncbi.nlm.nih.gov/38242596/
by Herrmann et al.

lib/core/utils/calc/tdee_calc.dart
```diff
@@ -24,29 +25,52 @@ class TDEECalc {
   /// Washington, DC: The National Academies Press.
   /// https://doi.org/10.17226/10490.
   /// https://nap.nationalacademies.org/catalog/10490/dietary-reference-intakes-for-energy-carbohydrate-fiber-fat-fatty-acids-cholesterol-protein-and-amino-acids
+  ///
+  /// For non-binary users without an explicit hormone profile, the result is
+  /// the mean of the male / female reference outputs. Users on HRT (or who
```
simonoppowa: question: Do you have any scientific reference for that? Taking the mean of male / female seems like a good solution

lib/features/add_meal/data/data_sources/off_data_source.dart
```diff
@@ -16,23 +17,23 @@ class OFFDataSource {
...
-          await httpClient.get(searchUrlString).timeout(_timeoutDuration);
-      log.fine('Fetching OFF results from: $searchUrlString');
-      if (response.statusCode == OFFConst.offHttpSuccessCode) {
-        final wordResponse = OFFWordResponseDTO.fromJson(
-          jsonDecode(response.body),
-        );
-        log.fine('Successful response from OFF');
-        return wordResponse;
-      } else {
+      return await withRetry(() async {
```
simonoppowa, line 20: praise: thanks for this change. I'm currently migrating the call to the search-a-licious api which hopefully works better

lib/features/profile/presentation/widgets/set_height_dialog.dart
```diff
@@ -67,10 +68,8 @@ class _SetHeightDialogState extends State<SetHeightDialog> {
         ),
         TextButton(
           onPressed: () {
-            Navigator.pop(
-              context,
-              clampHeightSelection(selectedHeight, minHeight),
-            );
+            final hardMax = widget.usesImperialUnits ? 15.0 : 457.0;
```
simonoppowa: suggestion(non-blocking): could be added to range_const

lib/features/settings/settings_screen.dart
```diff
@@ -112,6 +206,58 @@ class _SettingsScreenState extends State<SettingsScreen> {
...
+      BuildContext context, bool enabled, SettingsLoadedState state) async {
+    final notificationService = locator<NotificationService>();
+    await notificationService.initialize();
+    if (enabled) {
+      final granted = await notificationService.requestPermission();
+      if (!granted) {
+        if (context.mounted) {
+          ScaffoldMessenger.of(context).showSnackBar(
+            const SnackBar(
+                content: Text('Notification permission denied.')),
```
simonoppowa: issue: is this hardcoded string on purpose? Could be added as a translation key

lib/core/data/data_source/physical_activity_data_source.dart
```diff
@@ -1,11 +1,13 @@
 import 'package:opennutritracker/core/data/dbo/physical_activity_dbo.dart';
 
-/// Selection of  physical activities by the
-/// '2011 Compendium of Physical Activities'
-/// https://pubmed.ncbi.nlm.nih.gov/21681120/
-/// by Ainsworth et al.
+/// Selection of physical activities from the
+/// '2024 Compendium of Physical Activities'
+/// https://doi.org/10.1249/MSS.0000000000003624
```
simonoppowa: issue: seems to be wrong link.
Should be https://pubmed.ncbi.nlm.nih.gov/38242596/
by Herrmann et al.

simonoppowa (commented):
Incredible work 😁 I've tested the features and reviewed the changes. Added some minor comments. After merging is will upload a new build

lib/core/utils/calc/met_calc.dart
jordan-lee-code: You're right — the citation was stale. The 2024 Adult Compendium is Herrmann et al. (PMID 38242596). Fixed in d161ae22 alongside three other files in the same area that carried the same wrong attribution (`physical_activity_entity.dart` and `physical_activity_dbo.dart` had it too). Round-2 follow-ups for your review are in #372 — once that merges this release PR will pick them up. Thank you for the careful eye on this.

lib/core/data/data_source/physical_activity_data_source.dart
jordan-lee-code: Same fix as the `met_calc.dart` one above — d161ae22 covers both files plus the two others (`physical_activity_entity.dart`, `physical_activity_dbo.dart`) that carried the same stale citation, so the Herrmann reference reads consistently across the area now. Thank you again.

lib/features/add_meal/data/data_sources/off_data_source.dart
jordan-lee-code: Thank you, and really glad to hear search-a-licious is on the roadmap — that's a much bigger fix than retry-with-backoff, and it'll help users in regions where OFF's search has been flaky. Worth flagging for whenever your migration lands: the retry helper at this call site is a stopgap, and once you switch the underlying API the retry parameters will likely want re-tuning rather than removal — different latency profile, different failure modes. Excited to see it ship.

lib/features/profile/presentation/widgets/set_height_dialog.dart
jordan-lee-code: Done in d09bea5a — `_maxHeightCm` / `_maxHeightFt` and an `absoluteMaxHeight()` helper now sit next to the existing `_heightRangeCm` / `_heightRangeFt` constants. One small note: I put them in `profile_picker_bounds.dart` rather than the `Ranges` class in `ranges_const.dart`, because the picker bounds file is what `set_height_dialog.dart` already imports and it has the matching private-constants pattern. The `Ranges` class has `maxHeight = 300`, which is a tighter limit and a different concept from the picker's defensive 457 cm ceiling. Happy to move it across to `Ranges` if you'd prefer that let me know.

lib/features/settings/settings_screen.dart
jordan-lee-code: Not on purpose — caught and localized in 87b70676. While I was in there, two more hardcoded strings in the same notification flow surfaced (the daily-reminder title and body, used in both `_onNotificationToggled` and `_pickNotificationTime`), so the patch adds three new ARB keys — `notificationsPermissionDeniedSnack`, `notificationsDailyReminderTitle`, `notificationsDailyReminderBody` — translated across all eight locales. The localizations are captured before any `await` so the BuildContext lifetime concern stays contained. Thank you for the catch.

lib/core/utils/calc/tdee_calc.dart
jordan-lee-code: No peer-reviewed reference for the averaging itself — the IOM 2005 equations are sex-stratified, and there isn't a published non-binary baseline formula. The choice is a comfort and privacy one rather than a statistical claim, and your follow-up question made me realise that reasoning belongs in the file and in the UI rather than only in the PR thread. So this round-2 branch does both:

- 9ab73752 expands the docstring on `getTDEEKcalIOM2005` to explain the framing — the IOM equations only stratify by male and female reference samples, non-binary people aren't a homogeneous group sitting at any one point of that spectrum, and any single default we picked would be making a claim the source data doesn't support. Averaging gives a neutral midpoint bounded by the two reference outputs, and lets a non-binary user use the app without having to disclose whether their profile is estrogen- or testosterone-dominant.
- 87b70676 carries the same framing into the user-visible copy. The `caloriesProfileInfoBody` and the onboarding non-binary disclaimer are rewritten to say plainly that this is a starting point rather than a precise estimate, and to point at the kcal slider in Settings as the right place for personal fine-tuning. The radio labels are also relabelled so the formula routing is transparent: `Estrogen-typical (female formula)` and `Testosterone-typical (male formula)` make it clear that those opt-ins simply route through the existing male / female calculation rather than introducing a third equation.

Translated across all eight locales. Thank you for asking — the answer is more honest now than it was, and the people the work is for see the framing rather than having to take our word for it.

jordan-lee-code:
> Incredible work 😁 I've tested the features and reviewed the changes. Added some minor comments. After merging is will upload a new build

Thanks so much! ❤️ fixes made and ready for merge!

jordan-lee-code:
Sorry for the ping @simonoppowa , just to be sure, would you like me to merge to main or would you like to keep that in your line of responsibility?

simonoppowa (approved):
Amazing work 😁 Big thanks to @jordan-lee-code and all the other contributors. I will merge it now to main create a new release build

PR #371, opened 2026-05-07
Offline food catalog: download-prebuilt chunked model + supporting infrastructure

### LINKED ISSUE #353: Host curated OpenFoodFacts subset DB on the project website (deferred)
## Background

PR #319 proposed adding offline OFF search by having the user pick a manually-generated database file (~230 MB) on disk. The merged approach (PR #352, the per-scan local cache) covers the common case for free, but doesn't help cold-start scenarios — fresh installs, sudden offline use, or travelling somewhere with poor connectivity where the user hasn't yet had a chance to build a cache organically.

## Why "download direct from OFF" isn't viable

Investigated OFF's data export options as part of considering whether the app could just point users at OFF directly. **They do not offer per-country / per-region downloads.** Everything is whole-database:

| Format | Size | Per-country? |
|---|---|---|
| Parquet (Hugging Face) | **7.43 GB** food file | No — single file with `countries_tags` column |
| MongoDB dump | ~similar | No |
| JSONL | ~similar | No |
| CSV | 0.9 GB compressed / ~9 GB uncompressed | No |

The closest thing is filtering at *query time* via OFF's search API with `countries_tags=en:united-kingdom`. That works for runtime queries but isn't a "downloadable regional dataset" we could bundle or hand to users.

So this issue can't be solved by pointing the app at an OFF URL — the regional-subset filtering has to happen on infrastructure we control.

## Proposal

Once the project website is back up and being served via GitHub Pages (built by GitHub Actions):

- **Action runs weekly** (or whatever cadence makes sense): downloads the Hugging Face Parquet, filters by country / region, re-uploads small per-region `.json.gz` files (e.g. `uk-popular-2000.json.gz`, target ~5 MB each)
- **App settings** → "Download regional pack for [Country]" → fetches our small file from GitHub Pages → populates the existing cache box (the one PR #352 introduced) → the rest of the offline flow Just Works
- One-time download per region; user can refresh on demand
- Stored in app's external storage (doesn't count toward the APK size or Play Store entry)

## Why deferred

- The project website isn't currently being served — no place to host the file
- The local-cache approach (PR #352) already handles most users' offline needs
- Hosting + refresh automation is real ops effort; only worth doing if usage data shows the cache approach isn't enough
- An interim alternative considered — using OFF's search API to pre-populate the cache by region — was rejected because it shifts load onto OFF's already-strained infrastructure, which is the exact problem we're solving for. Better to do the filtering on our own infrastructure once.

## When to revisit

- The project's GitHub-Pages-based website is online again
- We see real user reports of "I just installed the app and can't search anything offline"
- Or: a contributor wants to take this on and run with it

## Implementation notes for the future PR

- Existing cache infrastructure to reuse: `lib/core/data/data_source/remote_search_cache_data_source.dart` (the box, plus the `cacheAll` / `getAll` / `getByBarcode` API)
- Settings already has the "Clear cached items" tile that shows count + size; the regional-pack flow should populate the same box so the existing UX stays consistent
- The GitHub Action should output stable filenames + a small `index.json` listing available regions + sizes + last-updated timestamps so the app can show a region picker

## Related

- PR #319 (closed in favour of the cache approach)
- PR #352 (the cache implementation)
- Issues #157, #229 (original OFF reliability complaints, closed)

jordan-lee-code:
Closes #353.

Adds an opt-in offline food catalog to OpenNutriTracker, plus the build pipeline and Cloudflare-side infrastructure that produces and serves it. Users install a prebuilt sqlite database of OpenFoodFacts products onto their device, and barcode and name searches resolve against the local index first with the live API as a fallback. The feature is opt-in and reversible: switching it off in settings reverts to the API-only path.

The catalog pipeline is running green end to end on this branch. The most recent run, including build, `tofu plan`, and `tofu apply` against the live infrastructure, is at [Actions run #25514623697](https://github.com/simonoppowa/OpenNutriTracker/actions/runs/25514623697); the run page carries the full plan output and the apply output for review.

Raised as a draft to allow review at your own pace before being marked ready.

## Overview

For the recommended default tier (`s1_n1_r5`, ~351k well-scanned products with full nutrition data, last five years, ~73 MiB compressed) the install completes in 10–20 seconds on Wi-Fi. The largest tier (`s0_n0_rany`, 3.2 million products, ~520 MiB compressed, ~3.7 GB unpacked) completes end-to-end in roughly two minutes. The catalog is read-only after install, so there is no ongoing background indexing.

## Domain

The `opennutritracker.org` domain has been purchased by Jordan-Lee-Code and is dedicated to this project. The catalog is served at `catalog.opennutritracker.org` through Cloudflare in front of the public-read R2 bucket; other subdomains are unused.

The domain can be transferred to the project owner's registrar account at any point. The procedure is:

1. EPP transfer code issued from Jordan-Lee-Code's registrar; the domain moves to the new owner's account.
2. The OpenTofu top-level API tokens and state-bucket credentials are re-bootstrapped under the new Cloudflare account, following `docs/opentofu.md` ("Bootstrapping from scratch").
3. The `catalog.opennutritracker.org` custom domain comes up against the new bucket on the next apply.

Transfer downtime is bounded by two protections. First, each chunk has a 7-day edge TTL via `cloudflare_ruleset.cache`, so existing users continue to fetch from Cloudflare's edge for up to a week even if the origin bucket is briefly unreachable. Second, the Flutter client refuses to begin a refresh unless the bucket is up and responding: the `CheckCatalogAvailabilityUseCase` runs at tap-time, and the per-chunk reachability probe runs in the download path, with both surfacing "caching currently unavailable, try again later" in settings rather than wedging the wizard. A transfer window of minutes to hours is therefore invisible to users with a catalog already installed and only blocks fresh downloads briefly.

## Cost

The pipeline runs at zero monetary cost on the following tiers:

- **Cloudflare free plan.** R2 bucket, custom domain, zone-level cache rule, and the two downstream API tokens are all within the free tier. R2 egress to Cloudflare's network is free regardless of plan.
- **GitHub Actions free for public repositories.** Standard GitHub-hosted runners on a public repo are free for unlimited minutes per [GitHub's billing reference](https://docs.github.com/en/billing/concepts/product-billing/github-actions). The catalog pipeline (`build_catalog.yml`) runs entirely on `ubuntu-latest`. The default Flutter workflow (`default_workflow.yml`, separate from the catalog pipeline) uses `ubuntu-latest` plus `macos-latest` for the iOS build job — `macos-latest` is a standard runner and is free on public repos; only the `-large` and `-xlarge` suffixed variants (e.g. `macos-latest-xlarge`) count as "larger runners", which are always charged. No `-large`/`-xlarge`, no GPU, and no self-hosted runners are used in either workflow. The catalog pipeline's single-job design also has no inter-job artefact handoff, so rolling artefact storage is zero too.

### Caching configuration

The free-tier cost model is contingent on the catalog responses being cached at Cloudflare's edge rather than served from R2 origin on every request. Two pieces of the design make that hold, both audited against [Cloudflare's documented default cache behaviour](https://developers.cloudflare.com/cache/concepts/default-cache-behavior/):

1. **Host-based cache rule overrides default extension rules.** Cloudflare's default cache logic caches a fixed list of static-asset extensions (`.gz`, `.zip`, `.png`, etc.) and explicitly does not cache HTML or JSON. The catalog serves chunk files named `variant.db.gz.part-NN` and per-variant `variant.manifest.json` manifests; neither is on the default-cached list (`.part-NN` is not a recognised extension, and `.json` is excluded). The `cloudflare_ruleset.cache` resource matches on `http.host eq "catalog.opennutritracker.org"` rather than on extension, so every response served from the custom domain is cached regardless of file extension or origin headers. Without this rule, Cloudflare would mark every response as `DYNAMIC` (compounded by R2 not setting `Cache-Control` headers on objects by default), the edge would never serve a hit, and every download would draw an R2 GetObject call.

2. **Chunk size stays under the free-plan cacheable cap.** Cloudflare's free, Pro, and Business plans cap cacheable response bodies at 512 MB. The largest catalog variant (`s0_n0_rany`) compresses to ~520 MiB — over the cap as a single file. The ≤256 MiB chunk size keeps every individual response comfortably under the free-tier limit while reading as one contiguous stream from the client's point of view. Cloudflare also caches HTTP 206 (Range) responses by default, which is what makes the client's parallel range-resume downloads efficient on cache hits.

The TTLs are split intentionally between the edge and the client:

- **Edge TTL: 7 days** (`var.edge_cache_seconds = 604800`). Long enough that a chunk that nobody requests for a couple of days still survives at the edge. Cloudflare's edge does not preferentially evict TTL-fresh content to make room for hotter data, so a 7-day TTL means cold chunks remain available for the full week between rebuilds.
- **Browser `Cache-Control: max-age`: 2 minutes** (`var.browser_cache_seconds = 120`). Deliberately short, so that the moment a rebuild's purge lands, HTTP clients converge on the fresh content within minutes rather than carrying stale bytes for hours. Client revalidations land on Cloudflare's edge — which still holds the chunks under the 7-day edge TTL — and either return 304 Not Modified (R2 sets ETags) for unchanged content or the fresh body for changed content; either way R2 origin is not touched, so the request volume costs nothing. This short value is also what makes the future in-app cache-refresh notification feature work cleanly: when that feature polls the manifest, it sees the post-purge state within minutes rather than hours.
- **Explicit edge purge after each apply.** A `Purge Cloudflare edge cache` step in `build_catalog.yml` runs immediately after `tofu apply` succeeds, calling `https://api.cloudflare.com/client/v4/zones/{zone_id}/purge_cache` with `{"hosts": ["catalog.opennutritracker.org"]}` to drop every cached response for the catalog host across Cloudflare's edge. The step is gated on the same triggers as the apply itself (push, schedule, dispatch on main) and inherits the implicit `success()` guard, so a failed apply does not purge a perfectly-good cache.

The combined effect is a sticky-but-fresh cache. Cold chunks survive a week without traffic, warm chunks pick up updates within seconds of each rebuild's apply via the explicit purge, and clients converge on the freshest edge state within two minutes even if no purge propagation reaches them. The seven-day edge TTL aligned to the weekly rebuild cadence keeps the cache hit ratio high in steady state, so R2 egress (and therefore monetary cost) stays at zero.

## Infrastructure as code

OpenTofu manages the Cloudflare side end to end: bucket, custom domain, cache rule, downstream API tokens, the seven sealed GitHub Actions secrets that publish those tokens into the build workflow, and **every individual chunk file**. Each chunk and every manifest is modelled as an `aws_s3_object` resource in the `catalog` module, keyed off `filesha256()`. Consequences:

- **Delta-only uploads.** A chunk whose content has not changed produces no diff and no PutObject. The Saturday cron uploads only the bytes that have actually changed week-over-week.
- **Plan as preview.** `tofu plan` previews which chunks are about to land before any bytes leave the runner.
- **No manual upload surface.** Adding, renaming, or removing a chunk happens through a code change.

State lives in a separate private R2 bucket and is encrypted at rest at the OpenTofu layer with PBKDF2-SHA512 + AES-256-GCM, `enforced = true`. The state file holds the raw values of the downstream API tokens, so the encryption layer prevents a leaked R2 credential from yielding a usable upload token. State locking uses OpenTofu's S3-native lockfile (`use_lockfile = true`); R2 has supported the required conditional-write semantics since October 2024, so no DynamoDB stand-in is required. A trailing `unlock_on_failure` job in the workflow handles a runner killed mid-apply, deleting the lockfile only when its `Created` timestamp is within the run's window so it cannot clobber a developer apply that happens to be in flight. Full reference, including the recovery procedure if the passphrase is lost, in [`docs/opentofu.md`](docs/opentofu.md).

## Build pipeline

The OpenFoodFacts CSV dump is approximately 1.2 GB compressed and several gigabytes uncompressed. The build pipeline reduces it to **a single immutable sqlite database per variant** — the largest at ~520 MiB compressed, the recommended default at ~73 MiB compressed. Stages, orchestrated by `tools/catalog_build/build_catalog.sh`:

1. Download the OFF dump.
2. Apply always-on filters (drop `obsolete`, `completeness < 0.3`, pet food, cosmetics, explicit non-food categories), then write each surviving row to every variant whose axis filters accept it.
3. Build a sqlite + FTS5 file per variant, with `unicode61 remove_diacritics 2` tokenisation.
4. Compress each `.db` with `pigz -6`, then split into ≤256 MiB chunks with a per-variant `manifest.json` carrying the chunk list, sha256 hashes (per chunk and combined), schema version, and total byte counts.
5. Hand `$WORK/dbs/` to OpenTofu, which uploads only chunks whose content has changed since the last apply.

Full reference, including the variant matrix and filter-axis semantics, in [`docs/catalog_build.md`](docs/catalog_build.md).

The Flutter client reverses the process: parallel chunked downloads with HTTP Range support, sha256 verification at part and combined levels, gunzip with backpressure, schema check, atomic rename. **Pause and resume are supported.** Backgrounding the app, dropping connectivity, or tapping "pause" in the wizard leave the partial install in a state where the next resume picks up at the last completed chunk and byte offset; per-chunk sha256 verification ensures a resume that picks up an incorrect byte stream fails at verification rather than silently corrupting the catalog.

## Wizard

The wizard exposes three filter axes for variant selection:

- **Well-scanned only** (`s` ∈ `{0, 1}`). With `s=1`, drops rows where `unique_scans_n < 2`, filtering out one-off submissions and abandoned edits. With `s=0`, retains all rows that survived the always-on filters.
- **Nutrition grade** (`n` ∈ `{0, 1}`). With `n=1`, requires `nutriscore_grade ∈ a..e`, guaranteeing parseable nutrition data on every row. With `n=0`, retains rows with no Nutri-Score.
- **Recency** (`r` ∈ `{3, 5, 10, any}`). Drops rows whose `last_modified_t` is older than the cutoff in years, or retains all rows for `any`.

Two by two by four gives sixteen combinations, named `s{0|1}_n{0|1}_r{3|5|10|any}`. The wizard estimates download size and on-disk size before commit, so the largest tier surfaces a "this will use 3.7 GB" warning before install rather than mid-install. The recommended default (`s1_n1_r5`) covers ~351k products at ~73 MiB compressed.

The wizard is six pages: welcome, quality (axis selection), estimate (computed download and on-disk sizes), download (active progress UI with per-chunk byte counts and a pause control), and a final ready / installed page. Bloc lifecycle, page-by-page state shape, and the fallback chain on download error are documented in [`docs/offline_catalog.md`](docs/offline_catalog.md).

## Reliability

- **Schema versioning in the catalog.** The sqlite carries `schema_version` (major) and `schema_version_minor` columns in a `catalog_meta` table. A client at major version N installs whatever the CDN serves at the same major, even if the minor has moved forward; a stale client is not locked out by a server-side schema bump.
- **Crash-safety auto-disable.** Two consecutive boot crashes turn the catalog off without deleting the data, and surface an explicit re-enable banner in settings, preventing the catalog from becoming a crash loop.
- **Tap-time availability probe.** The settings tile checks the CDN before opening the wizard. With no connection or a briefly unreachable bucket, the tile reads "caching currently unavailable, try again later" and the wizard refuses to open.
- **FTS5 query sanitiser.** Free-form input runs through a tokeniser that drops everything that is not a Unicode letter, digit, or whitespace, plus the four FTS5 boolean keywords. A natural-language query such as "Salt and Vinegar Crisps" cannot accidentally parse as binary operators. Three regression tests pin this behaviour.
- **Bundled sqlite on Android.** Android's bundled sqlite ships without FTS5 compiled in, so the app brings its own via `sqflite_common_ffi` and `sqlite3_flutter_libs`, adding ~3 MiB of APK weight in exchange for consistent FTS5 behaviour across Android, iOS, and desktop.

## Future enhancements

Out of scope for this PR; not blocked by it:

- **In-app cache refresh notifications.** Surface a banner when the manifest's `schema_version_minor` or combined sha256 differs from the installed copy. The `manifest.json` already carries the comparison data; the work is a settings-side check on app start plus a banner widget.
- **Project website on `opennutritracker.org`.** The apex domain is unused and could host a relaunched project website on GitHub Pages — landing page, screenshots, in-repo docs rendered as HTML, contributor guide, release notes. The catalog subdomain is independent and would not conflict.
- **Public sha hashes for sideload validation.** A static page (e.g. `releases.opennutritracker.org/<version>`) carrying the sha256 of every signed APK release, so users sideloading from F-Droid or a GitHub release can verify against a hash served from a project-controlled domain.

## Documentation

- [`docs/offline_catalog.md`](docs/offline_catalog.md) — client-side reference (wizard, bloc lifecycle, search and scanner fallback chain, availability gate, crash-safety auto-disable, forward compatibility).
- [`docs/catalog_build.md`](docs/catalog_build.md) — build pipeline reference (variant matrix, chunked layout, schema versioning, the CI workflow).
- [`docs/opentofu.md`](docs/opentofu.md) — infrastructure reference (resource list, secrets matrix, state locking, bootstrap procedure, recovery story, common gotchas).

`CLAUDE.md` is updated to point at all three.

## Testing

End-to-end verification on a Pixel 6 Pro on Android 16:

1. Recommended variant (`s1_n1_r5`): downloaded in ~14 s, installed cleanly, search returned hits.
2. Largest variant (`s0_n0_rany`): 520 MiB compressed downloaded in ~30 s, 3.7 GB unpacked in ~50 s with no OOM, 3.2 million rows installed.
3. Airplane mode + Wi-Fi off, with the largest catalog already on disk: punctuation-heavy query exercising the FTS5 sanitiser returned the matching product from the local index.
4. Tap-time availability probe: with both radios off, the settings tile shows "caching currently unavailable, try again later" and the wizard refuses to open; restoring connectivity recovers cleanly on the next tap.
5. Pause/resume mid-install: backgrounding the app mid-download, then returning, resumed at the next chunk boundary; final sha256 verification passed.

`flutter test` passes (356 tests). `flutter analyze` is clean. The catalog pipeline runs end-to-end on this branch under the single-job design (most recently in [run #25514623697](https://github.com/simonoppowa/OpenNutriTracker/actions/runs/25514623697)) with build, plan, and apply all green.

## Out of scope

- **Accessibility audit.** A per-feature `Semantics` audit, a style guide, and integration tests written against semantic labels rather than coordinates are deferred to a follow-up PR.
- **Two known UX bugs.** The progress UI does not converge to the final byte total when the phase flips from downloading to ready (the last emit is clipped); the wizard's auto-jump on a `ready` phase momentarily lands on the welcome page before bouncing forward. Both are minor and can be addressed in this PR or as a follow-up, per reviewer preference.

## Test plan

- [ ] Open the offline-catalog tile in settings on Android, walk through the wizard, install the recommended variant, confirm search works.
- [ ] Repeat with the largest variant, confirm install completes without the process being killed.
- [ ] Toggle airplane mode (with Wi-Fi off) and confirm the tile shows "caching currently unavailable, try again later" before any tap; confirm a tap re-runs the probe.
- [ ] With a catalog installed and both radios off, search for a product whose name contains a comma; confirm the result returns from the local catalog rather than failing through to the live API.
- [ ] Pause a download mid-flight (background the app, or toggle airplane mode briefly), then resume; confirm the wizard continues from the last completed chunk and the final sha256 verification passes.


jordan-lee-code:
@simonoppowa I've been working on this to allow the app to be fully offline with weekly db refreshes. It's all working and finished but I thought I'd wait to confirm if this the direction you want to go in before I raise it to develop. As stated in the PR review I am happy to hand over the domain to you anytime, I just registered it to play with 🙂

jordan-lee-code:
Not for self: implement file encryption using gpg to ensure only users can download files.

jordan-lee-code:
Search-a-licious api and request caching means this is no longer needed.

PR #410, opened 2026-05-14
feat(custom-meal): add a Simple form mode for direct macro entry

### LINKED ISSUE #232: Simplify custom meal page to allow direct input of macros
The current custom meal feature requires me to specify "meal size", "serving size" and "base quantity". It's not really clear how these three values feed into the final calculation however I don't think they're really necessary for most custom meals.

The main way that I use the custom meals feature is for when I have a meal where I can't separate out the various ingredients and I need to estimate the macros for the meal as a whole. It would then be simpler to just have a way to directly specify the macros rather than inputting the macros per 100g along with the weight of the meal.

This could live alongside the current input screen by having the current format as the "advanced input" mode.

jordan-lee-code:
Closes #232.

Adds a Simple/Advanced toggle to the custom-meal edit screen, replacing the per-100 base-quantity scaffolding with a direct "totals for one serving" entry mode that the issue specifically asked for. Advanced mode is unchanged. The chosen mode is remembered in `ConfigDBO`. Includes pulled-along kJ and per-meal-share work from the same area.

Synced with current `origin/develop` (Tier A + Tier B all merged). Slovak translations added for every new ARB key on this branch. `flutter analyze` clean; `just check_intl` clean.

## Test plan

- [ ] Build a debug APK on this branch, install, walk the affected flow on device
- [ ] Confirm existing features still behave as before (no regressions in the surrounding screens)


jordan-lee-code:
Heads-up before review: I pushed a follow-up restoration commit (16a68b6f) because the earlier rebase onto develop after #402 landed had used `--ours` too aggressively and silently dropped the screen-side half of the Simple/Advanced form along with its Hive plumbing and locale strings. The bloc, state, locator, and ADB probe survived; the screen UI, `@HiveField(18) customMealFormMode`, the repository/data-source `get/setCustomMealFormMode` methods, and every `customMealForm*` locale string did not. The branch as previously pushed did not compile — the bloc called repository methods that weren't there.

The restoration commit puts those pieces back from the pre-rebase tip without disturbing develop's `mealPattern*` additions from #402. The three earlier feature commits read partial in isolation; the message on 16a68b6f names what each was missing so the trail is in git for anyone reading later. `flutter analyze` reports zero issues, all 505 unit tests pass, and the diff stat against develop now matches what the branch was always meant to ship.

Apologies for the noise on this one.

PR #412, opened 2026-05-14
feat(onboarding): capture target weight on the current-weight screen

jordan-lee-code:
Builds on #406 (target weight) and #400 (weight history). The target weight field is now offered during onboarding, on the same screen where the user enters their current weight, so a new user doesn't have to discover the Profile setting after first-run to set one.

## What changed

- **Onboarding page 2** grows an optional target-weight section below the existing current-weight section. Same kg/lbs unit toggle, same input validation reused as `validateOptionalTargetWeight` (empty is valid; non-empty must parse as a weight). Leaving the field blank is a valid path — the user proceeds without a target, matching the Profile-side behaviour.
- **UserDataMaskEntity** carries the new `targetWeight` field through onboarding and writes it onto `UserEntity.targetWeightKg` in `toUserEntity()`.
- **Scroll wrapper.** The page body now sits inside a `SingleChildScrollView` so the extra row doesn't overflow on shorter screens or inside the widget tests, which pump straight into a bare Scaffold body.
- **ADB walker** (`tools/adb/walk-onboarding.sh`) swipes the page up before tapping the target field so the verifier can reach it on devices where it ends up below the fold. The walker fills the field with an example value (63) so the happy path is exercised end-to-end on every branch run.
- **Localisation.** `onboardingTargetWeightSubtitle` ("Is there a weight you're working towards? You can leave this blank or change it later in Profile.") and `onboardingTargetWeightHintOptional` ("Optional") added across all eight locale ARBs plus the matching `messages_*.dart` entries and `l10n.dart` getters. The existing `profileTargetWeightLabel` is reused as the section heading so the language stays consistent with Profile.

`flutter analyze` clean; `flutter test` all green (existing onboarding widget tests updated for the new 5-arg `setButtonContent` signature). Walked end-to-end on a Pixel 6 Pro via `tools/adb/walk-onboarding.sh`.

## Test plan

- [x] Build a debug APK, install, walk onboarding
- [x] Page 2: confirm the target-weight section appears below the current-weight section, with the same kg/lbs unit toggle
- [x] Leave target blank → NEXT remains enabled, complete onboarding, confirm Profile → Target weight reads "Not set"
- [x] Set a target weight → complete onboarding, confirm Profile → Target weight shows the value
- [x] Toggle to imperial mid-flow → both weight and target switch unit, both parse back to kg correctly
- [x] Navigate back from page 3 to page 2 → previously entered target value is restored

### DIFF
--- a/lib/features/onboarding/domain/entity/user_data_mask_entity.dart
+++ b/lib/features/onboarding/domain/entity/user_data_mask_entity.dart
@@ -16,6 +16,12 @@ class UserDataMaskEntity {
   UserActivitySelectionEntity? activity;
   UserGoalSelectionEntity? goal;
 
+  /// Optional weight the user is working towards. Captured on the same
+  /// onboarding screen as the current weight so new users don't have to
+  /// dig into Profile after first-run to set it. Stored in kilograms
+  /// regardless of the display unit the user picked.
+  double? targetWeight;
+
   bool acceptDataCollection = false;
 
   bool usesImperialUnits = false;
@@ -28,6 +34,7 @@ class UserDataMaskEntity {
     this.weight,
     this.activity,
     this.goal,
+    this.targetWeight,
     this.acceptDataCollection = false,
     this.usesImperialUnits = false,
   });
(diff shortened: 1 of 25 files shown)


jordan-lee-code:
Consolidated into #400 — the onboarding capture, the chart's dashed reference line, and the weight-history screen all sit on the same target-weight thread, so it reads more cleanly as one PR. The commit (1e6d6f2d) was cherry-picked onto feature/weight-tracking-log-38 and pushed; #400 now covers both.

Issue #107, opened 2024-07-23
iOS: descenders are cropped in dropdowns when text zoom is enabled


fuddl:
1. Go to Settings → Display & Brightnes → Text Size
2. Set the slider to anything bigger then 5 (I used 7 below)

Settings → Units | Add food
-----------------|--------------
![IMG_6574](https://github.com/user-attachments/assets/cd8d1a91-816e-41da-9e50-b2c8fdfa9be3) | ![IMG_6573](https://github.com/user-attachments/assets/945636eb-615c-409d-ae3e-f7efe4076ead)

jordan-lee-code:
Hi @fuddl — confirming this is still outstanding on `develop`.

Code location: `lib/features/` has zero usages of `MediaQuery.textScaleFactor` or `TextScaler` — none of the dropdowns are aware of accessibility text-scaling. Likely fix: audit `DropdownButton` / `DropdownButtonFormField` usages and apply text-scale-aware vertical padding (or wrap them in `MediaQuery` with a clamped `TextScaler`).

Labelled `bug`, `ios`, `accessibility`.

jordan-lee-code:
This is merged into `develop` now and is due to ship in 1.6.0. Thank you again @fuddl for the clear report and the side-by-side screenshots, they made this an easy one to pin down. If you get a chance to try it at a large Text Size once 1.6.0 is out, I would love to hear that the descenders look right on your end.

PR #433, opened 2026-05-17
ci(ios): manual workflow that builds unsigned IPA for AltServer sideload

jordan-lee-code:
## Why this branches off `main` rather than `develop`

`workflow_dispatch` triggers only work when the workflow file exists on the repository's default branch. If this lands on `develop` first it sits unusable until the next `develop → main` release merge. Treating it like the citations-onboarding hotfix in #423 — CI-only infrastructure, no runtime impact, defensible as a direct-to-main change.

## What it adds

Two new files. No edits to existing workflows, no edits to app code.

- `.github/workflows/ios_sideload.yml` — manual-trigger workflow that runs on a macOS runner, builds Flutter for iOS without code signing, wraps the produced `Runner.app` into an unsigned `.ipa`, and uploads it as a workflow artefact. Takes two inputs: a branch / tag / SHA to build against (defaults to `develop`), and a bundle identifier override (defaults to `com.opennutritracker.ont.opennutritracker.sideload`).
- `docs/ios-sideload.md` — setup notes for the Linux side of the flow: libimobiledevice install, AltServer-Linux configuration, free-Apple-ID quirks, the seven-day re-sign cadence, and troubleshooting for the common failure modes.

## Why the bundle ID gets patched

The production identifier `com.opennutritracker.ont.opennutritracker` is registered to the project's paid Apple Developer team. A free personal Apple ID — which is what AltServer-Linux uses on the developer's side — can't claim a bundle ID that's already on a paid team. The workflow patches the identifier to a `.sideload` suffix before running the iOS build so the signed sideload app installs without conflict.

Side-effect by design: the sideloaded build is a distinct iOS application from the App Store build. Separate Hive boxes, separate preferences, separate notifications. For UX / accessibility testing that's a feature, not a bug — the sideload install never collides with a real install on the same device.

## Why this exists

I'm hitting accessibility bugs like #107 (iOS dropdown descenders cropped at large text scales) that need a real device to verify, and I'm working from Linux without a Mac. The choices were: borrow a Mac for every test cycle, lean on @simonoppowa for TestFlight builds whenever I touch iOS code, or build the IPA in CI and sign locally with AltServer. The third one keeps the iOS dev loop honest without burning anyone else's time.

The workflow is `workflow_dispatch`-only so it doesn't run on every PR — macOS-runner minutes are noticeably more expensive than Linux, and the existing `ios-build` job in `default_workflow.yml` already covers compile-validation for fork / PR scenarios. This one exists specifically to hand a developer an installable artefact for on-device iOS testing, which is a different need.

## Test plan

- [x] YAML lint (`yamllint .github/workflows/ios_sideload.yml`) — no errors
- [ ] **First dispatch**: trigger the workflow against this branch, confirm the macOS runner completes, IPA artefact downloads, and the wrapped `.ipa` opens correctly via AltServer-Linux on a tethered iPhone. I'll do this immediately after merge and report back here with the artefact size and a screenshot of the installed sideload app.

## What this doesn't change

- No app code touched.
- No edits to `default_workflow.yml`. The existing `ios-build` job is unchanged.
- No new dependencies in `pubspec.yaml`. No iOS-native changes. `Podfile.lock` is regenerated by the existing `pod_install_with_targeted_fallback.sh` script the same way it does in the regular iOS job.


jordan-lee-code:
Closing until I've confirmed the workflow runs end-to-end against the branch. There was a bug in the bundle-id patch when the workflow fires without inputs (push trigger), which I'm iterating on now. Will reopen (or raise a fresh PR) once the IPA builds cleanly and an AltServer install lands on a real iPhone.

PR #435, opened 2026-05-18
feat(search): persist user-edited nutrient values across searches

h4rib0:
## Problem

When a product is found in the search but has no nutritional data, the user can fill in values manually in the Edit screen. The logged intake saves correctly with those values. However, **the search cache keeps the original empty version** — so the next time the same product is searched, the fields are blank again and the user has to re-enter everything.

## Solution

Two targeted changes:

### 1. Write enriched nutrients back to the search cache

`EditMealBloc` receives a `RemoteSearchCacheDataSource` dependency and a new `updateCachedMealNutrients()` method. After the user saves an OFF or FDC product in `EditMealScreen`, this method is called, writing the enriched `MealDBO` back to the cache. The next search returns the version with the user's values already filled in.

### 2. Protect cached nutrients during re-searches

`RemoteSearchCacheDataSource.cacheFromSearch()` now checks before overwriting an existing entry: if the cached version already has kcal data (user-entered) but the incoming remote result does not, the cached entry is kept. This prevents a background re-search from silently erasing manually entered values.

## Files changed

| File | Change |
|---|---|
| `remote_search_cache_data_source.dart` | `cacheFromSearch()` preserves user-enriched entries; `_hasNutrients()` helper |
| `edit_meal_bloc.dart` | `RemoteSearchCacheDataSource` dependency; `updateCachedMealNutrients()` |
| `edit_meal_screen.dart` | Call `updateCachedMealNutrients()` for OFF/FDC sources after save |
| `locator.dart` | Wire `RemoteSearchCacheDataSource` into `EditMealBloc` |

## Testing

- `flutter analyze` — no issues
- `flutter test` — 574 / 574 passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/supabase-fdc-setup/user_custom_meals.sql
+++ b/docs/supabase-fdc-setup/user_custom_meals.sql
@@ -1,0 +1,21 @@
+-- Table for user-created custom meals.
+-- Run this in the Supabase SQL Editor once to enable cloud sync.
+--
+-- Anonymous Auth must be enabled in Supabase:
+--   Authentication → Providers → Anonymous Sign-In → Enable
+
+CREATE TABLE IF NOT EXISTS user_custom_meals (
+    id          UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+    user_id     UUID NOT NULL REFERENCES auth.users(id) ON DELETE CASCADE,
+    lookup_key  TEXT NOT NULL,
+    meal_data   JSONB NOT NULL,
+    updated_at  TIMESTAMPTZ DEFAULT now(),
+    UNIQUE (user_id, lookup_key)
+);
+
+ALTER TABLE user_custom_meals ENABLE ROW LEVEL SECURITY;
+
+CREATE POLICY "Users manage their own custom meals"
+    ON user_custom_meals FOR ALL
+    USING (auth.uid() = user_id)
+    WITH CHECK (auth.uid() = user_id);
--- a/lib/core/data/data_source/custom_meal_data_source.dart
+++ b/lib/core/data/data_source/custom_meal_data_source.dart
@@ -1,10 +1,14 @@
+import 'dart:async';
+
 import 'package:hive_ce_flutter/hive_flutter.dart';
+import 'package:opennutritracker/core/data/data_source/custom_meal_supabase_data_source.dart';
 import 'package:opennutritracker/core/data/dbo/meal_dbo.dart';
 
 class CustomMealDataSource {
   final Box<MealDBO> _customMealBox;
+  final CustomMealSupabaseDataSource? _supabase;
 
-  CustomMealDataSource(this._customMealBox);
+  CustomMealDataSource(this._customMealBox, [this._supabase]);
 
   Future<void> saveCustomMeal(MealDBO mealDBO) async {
     final existing = _customMealBox.values.cast<MealDBO?>().firstWhere(
@@ -17,6 +21,7 @@ class CustomMealDataSource {
     } else {
       await _customMealBox.add(mealDBO);
     }
+    unawaited(_supabase?.upsert(mealDBO));
   }
 
   List<MealDBO> getAllCustomMeals() => _customMealBox.values.toList();
@@ -27,6 +32,20 @@ class CustomMealDataSource {
         .toList();
     for (final meal in toDelete) {
       await meal.delete();
+    }
+    unawaited(_supabase?.delete(mealKey));
+  }
+
+  /// Fetches custom meals from Supabase and saves any that are missing locally.
+  /// Called once at startup to restore data after a reinstall.
+  Future<void> syncFromSupabase() async {
+    final remote = await _supabase?.fetchAll() ?? [];
+    if (remote.isEmpty) return;
+    final localKeys = _customMealBox.values.map((m) => m.code ?? m.name).toSet();
+    for (final meal in remote) {
+      if (!localKeys.contains(meal.code ?? meal.name)) {
+        await _customMealBox.add(meal);
+      }
     }
   }
 }
--- a/lib/core/data/data_source/custom_meal_supabase_data_source.dart
+++ b/lib/core/data/data_source/custom_meal_supabase_data_source.dart
@@ -1,0 +1,82 @@
+import 'package:logging/logging.dart';
+import 'package:opennutritracker/core/data/dbo/meal_dbo.dart';
+import 'package:supabase_flutter/supabase_flutter.dart';
+
+/// Syncs user-created custom meals to Supabase so they survive a local cache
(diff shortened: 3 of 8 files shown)


h4rib0:
For reference: I also have a Layer 2 on top of this in my fork that adds optional cloud backup via Supabase anonymous auth.

When the user opts in (a toggle in Settings), manually entered nutrient values are written to a user_nutrient_overrides table in Supabase, scoped per device via anonymous auth — no account required. On app startup the overrides are fetched and merged back into the local cache, so the data survives a cache clear (Settings → Clear cache) as well.

I kept it out of this PR intentionally since it requires self-hosted Supabase infrastructure and wouldn't make sense as an upstream feature without a shared backend. Happy to share the implementation details if it's useful as a reference.

jordan-lee-code (commented):
Thank you for picking this up, @h4rib0. The core idea behind the first commit is genuinely good, and I want to start there before getting into the structural concerns, because it would be a shame for the useful piece to get lost in the rest.

## What I think is right

The cache-preservation logic in `remote_search_cache_data_source.dart` is the kind of small, well-placed fix that quietly improves the app. The bug it addresses is real: someone scans a barcode, finds an OFF entry with no nutrient data, fills in the values by hand, and the next time they search for the same item the cache hands back the original empty version. Adding `_hasNutrients()` and using it to gate the overwrite in `cacheFromSearch()` reads cleanly, the comment explaining the intent is helpful, and the `EditMealBloc.updateCachedMealNutrients()` callsite from `edit_meal_screen.dart` is in the right place. That part I'd happily see land.

## The reformatting sweep

What makes that piece hard to review right now is that the same commit also reformats large portions of `locator.dart` and `edit_meal_screen.dart`. `locator.dart` shrinks from 506 lines to 316, and `edit_meal_screen.dart` from 1,135 to 1,033, almost entirely by collapsing multi-line constructor calls onto single lines. None of that is mentioned in the PR description.

The project's existing style at BLoC registration sites uses the multi-line form deliberately: when a constructor takes seven positional `locator()` calls, having them stacked vertically makes the count and the dependency boundary legible at a glance. The single-line form, while shorter, is harder to scan and inconsistent with the rest of the file.

Could you split the reformatting out into a separate commit and drop it from this PR? Whether the project wants to move to a single-line registration style is a worthwhile conversation to have, but it shouldn't ride along with a feature change, because right now the real five-file diff is buried in around 250 lines of cosmetic churn and the actual logic is hard to find.

## The second commit

The bigger thing I want to raise, and it's the part I'd most like @simonoppowa's eyes on, is the second commit, `df8a2416 feat(custom-meals): sync user-created meals to Supabase for cross-install recovery`. That commit isn't mentioned anywhere in the PR title, body, or test plan, and it's doing something architecturally significant: adding Supabase Anonymous Auth, creating a new `user_custom_meals` table, and silently syncing every user-created custom meal to the cloud on save, with a restore-on-startup path that runs unconditionally.

I want to be careful here because I can tell the intent is kind. Losing custom meals on a reinstall is a real frustration, and the impulse to fix it for people is a good one. But there are a few things that I don't think can land in this shape:

**The privacy posture.** OpenNutriTracker's identity, both in `CLAUDE.md` and in the in-app disclaimer, leans on "all user data stored locally in an AES-encrypted Hive database." The export flow even deliberately excludes user profile data. This commit ships user-generated content to a third-party backend the next time someone updates, with no toggle, no opt-in, and no UI surface where someone could see that it was happening or ask for it to stop. Custom meal names can be quite personal, things like "Mum's chicken soup" or "Sunday at Sarah's", and shipping that to Supabase by default is a meaningful shift in what the app does on someone's behalf.

**The advertised feature doesn't actually work as written.** The commit message says "a reinstall no longer loses user data," but Supabase Anonymous Auth identities are tied to the device's local secure storage. A reinstall gives the user a fresh anonymous `user_id`, and the restore-on-startup query then fetches `user_custom_meals WHERE user_id = (new uuid)`, finds nothing, and restores nothing. There's no account linking, no recovery code, no email or any other anchor that would let the new install reattach to the old identity. So the very scenario the feature is trying to solve, reinstalling and getting your meals back, is the scenario it doesn't handle.

**Some smaller things, while we're here.** Sync is fire-and-forget with `unawaited(...)` swallowing every failure, so a user whose sync is silently failing has no way of knowing. Serialising the entire `MealDBO` to JSONB locks the wire schema to the on-disk Hive schema, which is going to bite us in the future when a `@HiveField` changes. `meal.code ?? meal.name` as a dedup key means renaming a meal orphans the previous row in Supabase forever. And `docs/supabase-fdc-setup/user_custom_meals.sql` lives under the FDC food-database setup folder, which is for an unrelated feature.

@simonoppowa, the broader question this raises is one I think really sits with you: does OpenNutriTracker want a cloud-sync surface for user-generated content at all? If the answer is no, then this commit comes out and we move on. If the answer is yes, then it's worth a proper design conversation about consent, recovery flow, what data is and isn't in scope, and how it's exposed in the UI, because that's the kind of feature that touches the heart of what the app is and how it presents itself to people. Either way I don't think it lands quietly as a second commit on a cache-update PR.

## A path forward

If you're up for it, @h4rib0, I'd suggest reshaping this as:

1. Keep only the cache-preservation feature here. The four-file substantive change (`remote_search_cache_data_source.dart`, `edit_meal_bloc.dart`, the small surgical edit to `edit_meal_screen.dart`, and the one-line dependency wire in `locator.dart`) is what the PR title promises, and it's worth landing.
2. Drop the reformatting from `locator.dart` and `edit_meal_screen.dart`. If you'd like to propose a style change separately, that's a conversation worth having on its own.
3. Pull the Supabase sync commit out into its own thread, ideally an issue first, so Simon can weigh in on whether it's a direction the project wants to go before more work happens on the implementation.

Thank you again for the contribution. The cache fix really is a nice catch, and I'd like to see it land in a form where it can be reviewed on its own terms.

h4rib0:
Thanks for the thorough review, @jordan-lee-code.
On the formatting changes:
Fair point — the auto-formatting of locator.dart and edit_meal_screen.dart shouldn't have been part of this PR. I'll clean that up in the new branch.
On the second commit (df8a241):
That commit ended up here by mistake — it was meant to stay on a separate branch. The Supabase sync feature is unfinished and was never intended to be part of this PR. I was already unsure whether cloud sync would be something the project wants at all, and I hadn't thought through what it would mean for regular users — the point about anonymous Supabase IDs being device-bound and not surviving a reinstall is something I wasn't aware of, so thank you for catching that.
I'll open a new branch with only the cache-preservation fix and the formatting corrected. The cloud sync feature will remain in my fork for personal use — I won't open a PR for it unless it's explicitly wanted by the project.

PR #436, opened 2026-05-18
docs: add FDC filter/translate scripts and Supabase setup guide

h4rib0:
## Summary

- Adds `scripts/filter_fdc.py` — filters the full USDA FoodData Central CSV download to Foundation & SR Legacy items only (~8 200 rows), making it fit within Supabase's free-tier 500 MB limit
- Adds `scripts/translate_fdc_de.py` — translates FDC food names to German via Google Translate (no API key required), batch-resumable
- Adds `docs/supabase-fdc-setup/README.md` — step-by-step guide covering download, filtering, translation, schema creation, and import; includes the SQL `CREATE TABLE` statements for `fdc_food`, `fdc_portions`, and `fdc_nutrients`

This documents the workflow needed to self-host FDC data in Supabase for apps that use the FDC search backend. No app code is changed.

## Test plan

- [ ] Run `filter_fdc.py` against a local FDC download, verify output row counts match documented ranges
- [ ] Run `translate_fdc_de.py` on a small subset, verify German names are written to output CSV
- [ ] Follow README step-by-step in a fresh Supabase project, verify tables are created and data imports without errors

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/docs/supabase-fdc-setup/README.md
+++ b/docs/supabase-fdc-setup/README.md
@@ -14,6 +14,28 @@ via Google Translate.
 
 ---
 
+## Quickstart – use pre-built CSV files
+
+If you want to skip Steps 1–4, pre-built CSV files (Foundation & SR Legacy,
+snapshot 2026-04-30, with German translations) are available at:
+
+[h4rib0/OpenNutriTracker – data/supabase-fdc/](https://github.com/h4rib0/OpenNutriTracker/tree/docs/supabase-fdc-setup-scripts/data/supabase-fdc)
+
+| File | Rows |
+|---|---|
+| `fdc_food_20260430_de.csv` | ~8 200 |
+| `fdc_portions_20260430.csv` | ~14 600 |
+| `fdc_nutrients_20260430.csv` | ~665 000 |
+
+Download the files and jump straight to **Step 5**.
+Use them as-is for the `fdc_food` and `fdc_portions` import in Step 6
+(set `base` to the folder containing the downloaded files and use the
+filenames above instead of `fdc_food.csv` / `fdc_portions.csv`).
+For `fdc_nutrients`, import `fdc_nutrients_20260430.csv` via the
+Supabase Table Editor CSV import.
+
+---
+
 ## Step 1 – Download the FDC data
 
 Go to the official USDA FoodData Central download page:
--- a/docs/supabase-fdc-setup/README.md
+++ b/docs/supabase-fdc-setup/README.md
@@ -1,0 +1,270 @@
+# Supabase FDC Database Setup
+
+This guide explains how to populate the `fdc_food`, `fdc_portions`, and
+`fdc_nutrients` tables in your Supabase project using freely available
+USDA FoodData Central (FDC) data — including optional German translations
+via Google Translate.
+
+## Prerequisites
+
+- A Supabase project (Free tier is sufficient)
+- Access to the Supabase SQL editor
+- ~500 MB of free disk space for the downloaded data
+- Python 3 with `deep-translator` (`pip install deep-translator`)
+
+---
+
+## Step 1 – Download the FDC data
+
+Go to the official USDA FoodData Central download page:
+
+```
+https://fdc.nal.usda.gov/download-datasets
+```
+
+Download **"Full Download of All Data Types – CSV"** (latest release, ~460 MB zipped).
+
+Unzip the archive. You will find many CSV files; you only need three:
+
+| File | Description |
+|---|---|
+| `food.csv` | Food items with names |
+| `food_portion.csv` | Serving size / portion data |
+| `food_nutrient.csv` | Nutrient values per food |
+
+---
+
+## Step 2 – Filter to Foundation & SR Legacy only (recommended for Free tier)
+
+The full dataset includes hundreds of thousands of branded products that
+would exceed Supabase's 500 MB free limit. Filter `food.csv` to keep
+only **Foundation** and **SR Legacy** rows (~8 000–9 000 items total).
+
+Run the script from the directory containing the unzipped CSV files:
+
+```bash
+cd /path/to/unzipped-fdc-data
+python3 /path/to/repo/scripts/filter_fdc.py
+```
+
+Output files are written to `./filtered/`:
+
+| File | Rows |
+|---|---|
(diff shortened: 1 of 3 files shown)


jordan-lee-code (commented):
Thank you for putting this together, @h4rib0. The filter script is genuinely useful, and walking the whole workflow from FDC download through to a populated Supabase project is the kind of thing that's been missing for self-hosters. I can tell a lot of care went into it.

Before this lands I'd like to loop in @simonoppowa, because the questions it raises are bigger than just code review and they touch decisions that are really his to make as the maintainer of the data side.

## The translation question

The German column comes from `deep-translator`'s Google Translate scraper, run unattended over ~8,200 USDA food descriptions. For most of those it will be fine, but for a nutrition app the failure mode of a wrong translation is meaningfully worse than in other domains: someone searching for a food in German and landing on a subtly different item gets the wrong calorie and macro numbers, which is exactly the thing OpenNutriTracker is being trusted to get right. There's no human-reviewed pass, no known-good dictionary for the high-traffic items, and no fallback when a translation is suspect.

The broader question this raises is how we want to think about translations into Supabase generally. The app already supports Czech, Italian, Polish, Turkish, Ukrainian, and Chinese in the UI, but the FDC backend currently only stores English and German. If we accept auto-translation for German, we're implicitly setting a precedent for the other six locales, and the same quality concerns apply to all of them. Conversely, if we'd want manual or human-reviewed translations for the other locales, it's worth asking whether German should be held to the same bar.

Simon, would you be more comfortable with auto-translated German shipping as part of an upstream-recommended setup, or would you rather Supabase be populated only with translations that have had a human in the loop? Either is a defensible choice, but it feels like a call you should make rather than one we sleepwalk into.

## The Quickstart link

The Quickstart section currently points at `h4rib0/OpenNutriTracker – data/supabase-fdc/` as the canonical source of pre-built CSVs. That's a really kind offer, @h4rib0, and I don't want to wave it away. But for upstream documentation to depend on a personal fork is fragile: the link breaks the day that branch is renamed or deleted, there's no way for a future user to verify the data hasn't drifted, and it puts you on the hook for maintaining a snapshot indefinitely. If we did want to ship pre-built CSVs from upstream, that's something we'd need to host somewhere the project controls.

## The Supabase setup as written

Step 5 disables RLS on all three FDC tables and Step 6 imports using the anon key. The anon key ships in the app (envied obfuscates the value but doesn't make it private), so a self-hosted setup following this guide ends up with three tables where anyone holding an APK can write. The right shape is to use the service-role key for the import, then re-enable RLS with a read-only policy for anon afterwards. Worth fixing before this is recommended widely.

## Smaller things

- `translate_fdc_de.py`'s `translate_batch` function sleeps `time.sleep(1.0)` after every single name and loops serially despite the name. That makes the README's "approximately 20–30 minutes" estimate not quite right (it's closer to 2.5 hours of sleep alone, before network time). `GoogleTranslator` has a real batch method that would help.
- `scripts/` doesn't exist on develop today. If the scripts stay, they probably belong under `docs/supabase-fdc-setup/scripts/` so the whole workflow lives together.
- This file overlaps with the `docs/supabase-fdc-setup/README.md` that's also in #437. Whichever direction we go here, that one should drop the docs portion to avoid the two PRs fighting each other.
- The README implies German is the supported translated locale, which matches the current `SupportedLanguage` enum. A sentence noting that the other locales currently fall back to English would help future contributors not assume they can extend this workflow to Czech or Polish without app-side changes too.

Again, @h4rib0, none of this is meant to discount the work. The mechanical pieces are well written and the README is genuinely thorough. The questions are about whether upstream OpenNutriTracker is the right venue for an unreviewed auto-translation pipeline, and I think that one's really for @simonoppowa to decide.

h4rib0:
Thanks for the detailed review, @jordan-lee-code.
On the translation quality:
I understand the concern about unreviewed machine translations in a nutrition app where accuracy matters. That said, manually reviewing ~8,200 food descriptions is not realistically feasible — certainly not for a single person, and definitely not scalable across all supported languages. Whether this sets an unwanted precedent for other languages is a fair question that @simonoppowa should decide.
On German search quality — alternative approach:
One reason I went down this path is that I was unsatisfied with the current German search results for basic foods like "Tomate" (tomato). As an alternative, I've integrated the Swiss Food Database ([naehrwertdaten.ch](https://naehrwertdaten.ch/)) in my fork (feat/swiss-food-api branch). It provides officially maintained nutrient data in German and English (I left out French since the app doesn't support it). This could be a cleaner alternative to auto-translating FDC data, at least for German-speaking users.
On the Quickstart link to my fork:
I intentionally kept the pre-built CSV files in my fork because I don't want to take on maintenance responsibility in the upstream project. I also assumed @simonoppowa might already have maintained datasets that haven't been published yet — if so, those would be the better official source.
On the RLS / security concern:
Fair point. The README is missing the step to re-enable RLS with a read-only policy for anonymous users after the import. I'd propose updating the README accordingly — use the service role key for the import, then re-enable RLS with anonymous read-only access.

simonoppowa:
Hi — thank you for the contribution! This is a great resource.

Just a heads-up: the Supabase backend has since been migrated to a multi-source schema and moved to its own repository. The old  /  /  tables are no longer used.

The new schema uses two relations:  (materialized view) and , and the backend is now maintained at https://github.com/simonoppowa/OpenNutriTracker-Backend with its own README covering the schema, import pipeline, and translation tooling.

This PR is based on the previous schema, so I'll close it as stale — but the idea and the scripts are valuable, and the contributor might want to revisit with the new setup. Thanks again for the effort!

simonoppowa:
Hi — thank you for the contribution! This is a great resource.

Just a heads-up: the Supabase backend has since been migrated to a multi-source schema and moved to its own repository. The old `fdc_food` / `fdc_nutrients` / `fdc_portions` tables are no longer used.

The new schema uses two relations: `food_summary` (materialized view) and `food_translation`, and the backend is now maintained at https://github.com/simonoppowa/OpenNutriTracker-Backend with its own README covering the schema, import pipeline, and translation tooling.

This PR is based on the previous schema, so I'll close it as stale — but the idea and the scripts are valuable, and the contributor might want to revisit with the new setup. Thanks again for the effort!

PR #437, opened 2026-05-18
feat: keep + button always visible in meal lists

h4rib0:
## Summary

- The + button (PlaceholderCard) in each meal section (breakfast, lunch, dinner, snack) is now fixed at index 0 (leftmost position) in the horizontal scroll list
- Previously, new items were inserted before the button, pushing it progressively to the right until it scrolled out of view
- New items now appear to the right of the button, keeping it always accessible without scrolling

## Test plan

- [x] Add several items to a meal section and verify the + button stays visible on the left
- [x] Verify tapping the + button still opens the add-meal flow
- [x] Verify drag & drop reordering still works between meal sections
- [x] Verify the leftmost item spacing (16px leading padding) is applied correctly to the + button


jordan-lee-code (commented):
Hi @h4rib0, thank you for picking this up. The + button change itself is exactly the right move, and the new index-0 layout in `intake_vertical_list.dart` reads cleanly.

There are a few things I'd like to sort out before this can land, and I want to walk through them rather than just point at them.

## Bundle scope

This branch has grown a little beyond what the title describes. Alongside the + button fix, it also contains:

- A separate citations/sources UX change (entry points added to onboarding intro, onboarding overview, and settings, with the link removed from the disclaimer dialog), plus a new `onboardingIntroSourcesLinkLabel` string across all eight locales.
- A new 271-line `docs/supabase-fdc-setup/README.md`.
- Two pubspec version bumps.

Each of those is genuinely useful, but for review and rollback I'd much rather see them as three separate PRs:

1. The + button fix on its own (this one is almost ready).
2. The "sources entry points" feature (the citations commit, plus the matching disclaimer dialog change).
3. The Supabase FDC docs.

The version bumps can come out entirely. We only bump `pubspec.yaml` on the `develop → main` release merge, never on feature branches into `develop`.

## `intl_en.arb` has accidental duplicates

The citations commit (`441fc15d`) ended up re-inserting roughly 184 lines of existing strings at the top of `lib/l10n/intl_en.arb`. JSON still parses because the later entry silently wins, but the file now has around 170 duplicate keys (`appTitle`, `onboardingIntroDescription`, `settingsSourcesLabel`, and many more each appear twice). That will likely trip `just check_intl`, and it's worth cleaning up regardless of how the branch ends up being split.

## A small cleanup in `intake_vertical_list.dart`

While you are in there: the `childWhenDragging` row now has `const SizedBox(width: 0)` as its first child. Since the + button is always leftmost and the dragged-from card no longer needs a leading spacer, that `SizedBox` can just be removed.

No rush at all. Ping me when you've had a chance to split this up, and the + button change on its own should be quick to land once it's standalone.

jordan-lee-code:
One follow-up on the citations piece, @h4rib0: I'm not sure I understand the motivation for removing the Sources button from the disclaimer dialog. The other entry points you're adding (settings, onboarding intro, onboarding overview) layer on top of what's already there, but the disclaimer dialog removal is a net loss of a path users currently have to the Sources screen.

Could you say a bit more about what's behind that change? If there's a specific reason the disclaimer dialog felt like the wrong place for it, I'd like to understand the thinking. Otherwise I'd lean towards keeping it as it stands on develop, so people who tap into the disclaimer still have the sources one tap away.

(For what it's worth, the existing info-icon entry point on the gender selection page should stay regardless. That one is well placed.)

h4rib0:
Thanks for the detailed review!

**On the Sources button removal:** This was unintentional. I have a separate private branch (`influx_db_activity_sync`) that I haven't submitted as a PR — it contains personal infrastructure integrations that aren't suitable for public use:
- **Polar OAuth sync** running on a separate server I host myself, to track activity data
- **OpenScale weight sync** pulling body weight from my OpenScale setup
- **A different nutrition model** adapted to my personal tracking setup
- **InfluxDB/Grafana sync** — all data from OpenNutriTracker (intake, activity, weight) flows into my personal Grafana health dashboard where I also visualize Polar and OpenScale data

The Sources button removal was a side effect of that branch's changes bleeding into this PR — not a deliberate UX decision. I've restored it.

**On splitting the PR:** I understand the reasoning, but splitting at this point would be quite time-consuming. I've addressed all other feedback (SizedBox cleanup, intl_en.arb duplicates, version bump reverted). Going forward I'll keep PRs smaller and more focused.

jordan-lee-code:
Thank you for your patience with this one, h4rib0. The branch had drifted a fair way behind develop, so I merged develop in to bring it back into line rather than ask you to untangle it yourself. Your commits are all still here, so the + button work stays yours.

The reason most of the diff has vanished is that the citations/sources work this PR originally bundled has since landed on develop on its own, so those files simply fall back to what already shipped. What remains as this branch's own contribution is exactly the part that was always the point: the + button pinned to the left of each meal list, plus the Supabase FDC setup docs.

It is mergeable now, and `flutter analyze` and `just check_intl` both pass on the merged tree. I tested the + button on a device and it stays put as items fill the row; the cross-section drag-and-drop still keys off the dragged item rather than its index, so the reordering is unaffected.

PR #439, opened 2026-05-19
Integrate cicd store deployment

simonoppowa:
Bildschirmfoto vom 2026-05-19 11-20-22

This pull request introduces significant improvements to the CI/CD pipeline for both Android and iOS platforms, focusing on automated build, signing, and deployment processes. It adds Fastlane configuration for both platforms, custom GitHub Actions for environment setup, and makes important adjustments to code signing and provisioning. Additionally, it bumps the app version and makes minor Android permission changes.

**CI/CD Pipeline Enhancements**

* Added Fastlane configuration files for iOS (`Fastfile`, `Appfile`, `Matchfile`) to automate building, signing, and deploying the app to TestFlight and App Store, with supporting documentation (`README.md`). [[1]](diffhunk://#diff-96a113f9048a44bf631e40227acc27a302b3daa91b60217ad3a7ff91fa6973deR1-R93) [[2]](diffhunk://#diff-4be9cff0008e0c5aaec38ec7efa93324983d02b6a53f40aa42f9c81a0163f3e7R1-R3) [[3]](diffhunk://#diff-c85a9e151eb8f38291c1f7e8e0f37499149413d1be702823b8893d3051700205R1-R10) [[4]](diffhunk://#diff-24d19c7c42c2d16405b8daea904115ce70e7746fe409ae8cb6b9a1da7c42bef5R1-R56)
* Added Fastlane configuration for Android (`Fastfile`, `Appfile`) to automate building and uploading app bundles to Google Play internal testing. [[1]](diffhunk://#diff-d0c99c3583617d28a3669e90c98158993dfbd7a57c6e010b1a70e2a8acaed5a6R1-R24) [[2]](diffhunk://#diff-9a04fd97ecbecb094c8083c0edba4070bfd3b834479000ef278ff1c134a1009bR1-R2) [[3]](diffhunk://#diff-61751fb38ef24d7c31ebf37a4397c15f2b3ba2f1c34b6bdade8c25a6b46ae7e5R1-R4)

**Custom GitHub Actions**

* Introduced `.github/actions/setup-flutter-cache/action.yml` to set up the Flutter environment and cache pub packages in CI.
* Added `.github/actions/write-env-file/action.yml` to generate a `.env` file from provided secrets, ensuring all required values are present.

**iOS Code Signing and Build Settings**

* Updated `Runner.xcodeproj/project.pbxproj` to use manual signing with "Apple Distribution" and specified the App Store provisioning profile for release builds. [[1]](diffhunk://#diff-413fd2395fd4377132c8c1135c9bc9ab215c6f97322d27717a487223bd03e2c7L368-R369) [[2]](diffhunk://#diff-413fd2395fd4377132c8c1135c9bc9ab215c6f97322d27717a487223bd03e2c7L382-R382) [[3]](diffhunk://#diff-413fd2395fd4377132c8c1135c9bc9ab215c6f97322d27717a487223bd03e2c7L538-R539) [[4]](diffhunk://#diff-413fd2395fd4377132c8c1135c9bc9ab215c6f97322d27717a487223bd03e2c7L552-R552)
* Added `ios/Gemfile` to manage Ruby dependencies for Fastlane and CocoaPods, and `.bundle/config` to specify the bundle path. [[1]](diffhunk://#diff-23327eb2c1bc1f40fe5f31045e7949b7012638a7732d6a8eef56371b453e1080R1-R6) [[2]](diffhunk://#diff-ca525a15f92638e5f03f685f0a051d71ce2ee9d21da9a43f99d4026b0eb51862R1-R2)
* Enabled GPU validation mode in the Xcode scheme for improved debugging.

**Android Permissions and Versioning**

* Removed the `READ_MEDIA_IMAGES` permission in `AndroidManifest.xml`
* Bumped the app version to `1.3.2+54` in `pubspec.yaml`.


android/app/src/main/AndroidManifest.xml
```diff
@@ -5,11 +5,6 @@
     <!-- #312: Notification reminders -->
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
-    <!-- #64/#144: Attach a photo to a custom meal. Android 13+ requires the
-         scoped READ_MEDIA_IMAGES permission for image_picker's gallery flow;
-         camera capture is delegated to the system Camera app, so no
-         CAMERA permission is needed here. -->
-    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
```
jordan-lee-code, line 12: Heads up — `image_picker: ^1.1.2` is still a dependency in `pubspec.yaml`, and it's still called from `lib/features/edit_meal/presentation/edit_meal_screen.dart`, `lib/features/recipes/presentation/screens/recipe_builder_screen.dart`, and `lib/core/presentation/widgets/user_image_picker_tile.dart`. On Android 13+ the gallery flow needs this scoped permission; without it, choosing a photo for a custom meal will throw at runtime. The original PR (#399) added the permission with a comment explaining exactly that scenario. If this removal isn't part of removing the photo feature wholesale, it likely needs to be reverted in this PR — otherwise this change ships a regression to anyone on Android 13 or newer.

Retracted: the original removal was correct. `image_picker_android` 0.8.13+17 uses Android's Photo Picker (`PickVisualMedia`) for the gallery and only ever asks for `CAMERA` at runtime, so no scoped media permission is needed. See follow-up on #442.

.github/workflows/default_workflow.yml
```diff
@@ -485,6 +406,346 @@ jobs:
...
+        uses: actions/checkout@v6
+
+      - name: Setup Ruby + Bundler
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.3'
+          bundler-cache: true
+
+      - name: Download iOS IPA artifact
+        uses: actions/download-artifact@v4
```
jordan-lee-code: Worth flagging a version mismatch: the package jobs use `actions/upload-artifact@v7` (lines 512, 656, 668), but every download in this workflow is on `@v4`. upload-artifact's storage format changed across major versions, and v7 artefacts aren't guaranteed to be readable by v4 downloads — in practice this tends to fail at download time with a 'no matching artifact' or format-incompatibility error. Easiest fix is to bump each `download-artifact` reference to `@v8` (current major) so the major versions line up; alternatively pin everything to `@v4`.

Addressed in #442 ([commit](https://github.com/simonoppowa/OpenNutriTracker/commit/e7e6b5d1730e9a22bc39dcc3f406234065be2891)).

ios/fastlane/Fastfile
```diff
@@ -0,0 +1,93 @@
...
+    sync_signing
+  end
+
+  desc 'Build a signed App Store IPA using Flutter'
+  lane :build do
+    setup_ci if ENV['CI']
+    sync_signing
+    enforce_app_store_signing
+    sh('cd .. && fvm flutter pub get')
+    sh('cd .. && fvm flutter build ipa --release --export-method app-store')
```
jordan-lee-code: The build command doesn't pass `--build-number`, so `CFBundleVersion` comes straight from `pubspec.yaml` (currently `+55`). TestFlight rejects an upload whose `CFBundleVersion` matches a previously uploaded one for the same `CFBundleShortVersionString`, so the second time this pipeline runs against the same pubspec version, the upload step fails. The earlier scaffolding solved this by deriving the build number from `github.run_number + 1000` in the workflow and passing it through, which keeps every upload strictly increasing without needing a pubspec bump per run. Could be done by reading an env var here (e.g. `--build-number=#{ENV.fetch('CI_BUILD_NUMBER', '1')}`) and setting `CI_BUILD_NUMBER` in the `ios-package` job.

Addressed in #442 ([commit](https://github.com/simonoppowa/OpenNutriTracker/commit/ed20666e28680340f6fed45704a5eccf56aa5db9)).

.github/actions/setup-flutter-cache/action.yml
```diff
@@ -0,0 +1,24 @@
+name: Setup Flutter and cache pub packages
+description: Installs Flutter and restores/saves the pub cache.
```
jordan-lee-code: Small mismatch between the description and what the action does — the description says it restores and saves the pub cache, but the steps only call `subosito/flutter-action@v2` with `cache: true`, which caches the Flutter SDK itself rather than `~/.pub-cache`. The previous workflow had a separate `actions/cache@v5` step keyed on `pubspec.lock` for the pub cache; without it, every job pays the `flutter pub get` cost from scratch (around a minute on each run). The `cache-key-prefix` input is also declared but never read. Two small fixes would tighten this up: add the `~/.pub-cache` step here, and either wire `cache-key-prefix` through to its key, or remove the input.

Addressed in #442 ([commit](https://github.com/simonoppowa/OpenNutriTracker/commit/b8142f924903612157e5c5587d2cefebdef6562c)).

.github/workflows/default_workflow.yml
```diff
@@ -485,6 +406,346 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-pods-
+
+      - name: Install FVM
+        run: |
+          dart pub global activate fvm
+          echo "$HOME/.pub-cache/bin" >> "$GITHUB_PATH"
+
+      - name: Install Flutter version for FVM
+        run: fvm install 3.41.7
```
jordan-lee-code, line 476: Worth noting that `setup-flutter-cache` a few steps up installs Flutter via subosito's action, and then this step downloads Flutter 3.41.7 a second time under FVM. The build lane uses `fvm flutter`, so the subosito copy is never read on this runner — it's a download and SDK extraction that takes several minutes on macOS for no behavioural reason. Picking one approach for the whole job (FVM throughout, or subosito throughout) would shave noticeable time off each package run.

Addressed in #442 ([commit](https://github.com/simonoppowa/OpenNutriTracker/commit/852848d41b6bcc00688bab81e8503b6660c94bc5)).

.github/workflows/default_workflow.yml
```diff
@@ -485,6 +406,346 @@ jobs:
...
+          path: build/ios/ipa/*.ipa
+          if-no-files-found: error
+
+  ios-deploy:
+    if: |
+      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
+      github.event_name == 'workflow_dispatch'
+    needs:
+      - ios-package
+    runs-on: macos-26
```
jordan-lee-code: Small one worth flagging: `ios-deploy` only downloads the IPA artefact from the previous stage and uploads it to TestFlight via fastlane — both pure network operations. Nothing in this job actually needs a macOS environment, so `runs-on: ubuntu-latest` would handle it equally well and would free `macos-26` capacity for the package job that genuinely needs it. Easy one-line change whenever you're tidying.

Addressed in #442 ([commit](https://github.com/simonoppowa/OpenNutriTracker/commit/e8322867ef2f66a694a32d604314a3c780675676)).

.github/workflows/default_workflow.yml
```diff
@@ -485,6 +406,346 @@ jobs:
...
+      - name: Download Android APK artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: android-apk
+          path: release-assets/android
+
+      - name: Prepare release tag and asset names
+        run: |
+          APP_VERSION=$(awk '/^version:/ {print $2}' pubspec.yaml)
+          TAG_NAME="v${APP_VERSION}-build.${GITHUB_RUN_NUMBER}"
```
jordan-lee-code: The tag this produces is `v1.3.2+55-build.42` (and similar). F-Droid's build server watches `vX.Y.Z` tags specifically when deciding what to build, so the `+55-build.42` suffix would prevent it from spotting new releases — F-Droid releases would silently stop with this format. The project's convention historically has been plain `vX.Y.Z`. If a per-run identifier is useful for debugging, it could go into the release name or body rather than the tag itself. Something like:

```bash
APP_VERSION=$(awk '/^version:/ {print $2}' pubspec.yaml)
VERSION_NO_BUILD="${APP_VERSION%+*}"
TAG_NAME="v${VERSION_NO_BUILD}"
```

Addressed in #442 ([commit](https://github.com/simonoppowa/OpenNutriTracker/commit/3076f18c916ce26fac600323f7f6e332cbcc92b6)).

pubspec.yaml
```diff
@@ -15,7 +15,7 @@ publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 # In iOS, build-name is used as CFBundleShortVersionString while build-number is used as CFBundleVersion.
 # Read more about iOS versioning at
 # https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
-version: 1.3.1+49
+version: 1.3.2+55
```
jordan-lee-code: Patch bump might be understating this release — develop has shipped a substantial run of user-facing features since 1.3.1, including the fasting timer (#432), Material You theming with the accent picker (#426), the water tracker (#418), weight history with the trend chart (#400), custom meal photos (#399), custom meal barcodes (#407), an explicit target weight with optional calorie taper (#406), the daily and weekly micronutrient panel with DRI bars (#404 / #427), kJ as an alternative energy display (#403), and the Slovak locale (#392), among others. Per the project's bump-level guide in CLAUDE.md, that lands firmly in minor-bump territory — 1.4.0+50 would be the natural number for the release this PR ships.

Addressed in #442 ([commit](https://github.com/simonoppowa/OpenNutriTracker/commit/24d781db222dcb1095cb3efe7f32d1510e291d12)).

ios/Gemfile
```diff
@@ -0,0 +1,6 @@
+source 'https://rubygems.org'
+
+gem 'fastlane', '~> 2.234'
+gem 'cocoapods', '~> 1.15.2'
+gem 'rexml', '>= 3.3.6'
+gem 'ffi', '>= 1.15.0'
```
jordan-lee-code: Tiny thing — this file is missing its trailing newline. Most editors and tools (rubocop, `cat -A`, POSIX text-file conventions) prefer a final newline; some flag its absence. One stray byte to add.

Addressed in #442 ([commit](https://github.com/simonoppowa/OpenNutriTracker/commit/0b1a31fa6cc1e4a98018e371b46e4c7a531adc95)).

ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -48,6 +48,7 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
+      enableGPUValidationMode = "1"
```
jordan-lee-code, line 51: Curious about this one — `enableGPUValidationMode = "1"` turns on Metal's GPU validation layer for debug runs, which adds a small per-frame runtime cost during development. If it was switched on while debugging something specific, it might be worth taking back out before this lands, since the scheme is shared and every contributor inherits the setting. If it's intentional (for catching graphics issues early during development), no concern — just flagging in case it slipped in by accident.

On reflection: not addressing this in #442. Your PR body for #439 explicitly notes you enabled GPU validation "for improved debugging", which is a deliberate dev-time choice rather than something that slipped in. The scheme attribute only affects debug runs in Xcode anyway, so release builds aren't paying the per-frame cost. Leaving the call entirely with you.

jordan-lee-code (commented):
Thank you for this pipeline — separating the package and deploy phases, using fastlane match with SSH auth, and going through `setup_ci` and `update_code_signing_settings` are all the right choices, and the github-release job that attaches every artefact is a thoughtful touch for self-hosters. Comments below are mostly edges that show up the first time the pipeline runs end-to-end; the architecture itself is sound.

I've also raised #442 stacked on this branch with the subset of these notes that are factual regressions or pipeline correctness issues. The fixes are split into one-commit-per-issue so it's easy to cherry-pick the ones you agree with and drop the rest — each addressed inline comment below links to the specific commit. The judgment-call notes (runner choice, single-vs-double Flutter install, GPU validation mode, version-bump semver level, BUNDLE_PATH redundancy in CI) stay here as conversation rather than fixes, since those are yours to make.

PR #441, opened 2026-05-19
ci: shard iOS + Android integration tests by file, gated on a clean build

jordan-lee-code:
## Summary

The two integration jobs ran every test file serially on a single runner, so wall-clock time grew linearly with the suite. There's an obvious second integration test landing in the share-QR work and more on the way, so this is the right moment to put the matrix infrastructure in place.

- A new `discover-integration-tests` job lists `integration_test/*.dart` at the start of the run and outputs the list as JSON.
- `ios-integration-tests` and `android-integration-tests` both consume that list as a matrix dimension. One shard per file, running on its own runner.
- Both matrices `needs: [<platform>-build, discover-integration-tests]` — there's no point burning macOS minutes on a matrix of integration shards if the app doesn't compile.
- `fail-fast: false` so a flake in one shard doesn't cancel the rest. Each test reports independently which is easier to triage.

## What landed in the integration_test folder

The single `app_smoke_test.dart` bundled three independent assertions into one `testWidgets` block. I split it into two files along the natural question boundaries a smoke test asks of the boot:

- `app_boot_test.dart` — boot-health. Did `main()` finish and a `MaterialApp` end up on screen? Did anything fire `FlutterError.onError` along the way? One simulator boot, two assertions sharing it. Catches the loud failures (Hive can't open, secure storage can't derive an AES key, GetIt missing dep) and the quiet ones (Hive type-id collision, dropped `await` in plugin init, half-broken notification re-register).
- `first_screen_test.dart` — first-screen smoke. What does a fresh-install user actually see on page one? Routing picked `OnboardingScreen`, `OnboardingBloc` reached loaded state so `IntroductionScreen` mounted, intl delegates loaded so a known ARB string rendered. Three layers of the same observation, one boot.

## Why two shards rather than five

I tried this at file granularity first — five focused single-assertion files. The first run on a busy macOS pool surfaced the real-world ceiling. Only two of the five iOS shards got runners; the other three sat queueing on:

> All GitHub-hosted runners with label [macos-latest] are busy.

That's GitHub's global Mac fleet, not our per-account 5-concurrent cap. On evenings like that, a five-way fan-out reverts to roughly two waves of effective parallelism, with the queue depth gating wall-clock time instead of our work. Two shards stays friendlier to that pool — both runners are much more likely to be granted at once — and the assertions group cleanly along their natural seams anyway. The matrix infrastructure is unchanged; the file count is just the tactic that fits the current pool reality.

If pool availability turns out to be better than that one observation suggested, the same workflow grows back toward more shards by adding files — no workflow edit needed.

## Why this is safe

Sharding via `strategy.matrix` is a first-class GitHub Actions feature explicitly designed for this. For public OSS on GitHub-hosted runners (which this repo is) there's no minute cost. The free-tier per-account concurrency limit on macOS is 5 jobs; we're nowhere near that with two shards.

The AVD cache key is unchanged, so only the first Android shard on a cold cache pays the warm-up cost; later shards in the same run reuse the snapshot via `actions/cache` restore-keys.

## What's not in scope

This PR doesn't try to "build once, test many" by sharing artifacts across jobs. Flutter's `flutter test integration_test/` rebuilds the app inside each shard. Going further would mean swapping to `flutter drive --use-application-binary` or xcodebuild's `test-without-building`, both of which add fragility. Worth doing later if shard count grows enough that the per-shard build cost dominates wall-clock time.

## Test plan

- [ ] Confirm `discover-integration-tests` lists `app_boot_test.dart` and `first_screen_test.dart`.
- [ ] Confirm both matrices expand to two shards each, gated on the platform build.
- [ ] Confirm the run finishes in roughly `build + max(shard_duration)` rather than the previous linear sum.
- [ ] Re-evaluate file count after a week of seeing how often the macOS pool actually has capacity for our shards at the times we typically push.


jordan-lee-code:
Waiting for #439 and then will integrate these changes.

PR #442, opened 2026-05-20
ci: address review follow-ups on the CI/CD pipeline PR

jordan-lee-code:
Stacked on top of your CI/CD pipeline PR (#439) with the subset of our review notes that are factual regressions or pipeline correctness issues. Splitting them into atomic commits so you can cherry-pick the ones you agree with and drop the rest without unpicking a single squashed diff.

What's here, in commit order:

- `fix(android): restore READ_MEDIA_IMAGES permission` — `image_picker` is still in use across the custom-meal photo flow and the avatar tile, so removing the scoped permission would crash the gallery picker on Android 13+.
- `ci(actions): drop unused cache-key-prefix from setup-flutter-cache` — the input was declared but never consumed, and the action description claimed to cache `~/.pub-cache` while it only invokes subosito's SDK cache. No behaviour change.
- `chore(ios): add trailing newline to Gemfile` — one stray byte.
- `ci(ios): pass CI build number through fastlane` — wires `CI_BUILD_NUMBER` from `github.run_number` into the `ios-package` job and threads it through to `flutter build ipa --build-number`, so repeat TestFlight uploads don't collide on `CFBundleVersion` when the pubspec hasn't bumped. Fastfile defaults to `'1'` when the env is unset, so local fastlane invocations are unaffected.
- `ci(workflow): align download-artifact to @v7 with the uploads` — every `download-artifact` reference moves to `@v7` to match `upload-artifact@v7`. The on-disk format changed between v4 and v7, so the mismatch breaks the deploy/release jobs at download time.
- `ci(release): keep GitHub release tag as plain vX.Y.Z` — F-Droid's build server only spots tags of the `vX.Y.Z` shape, so the `+N-build.M` suffix would silently halt F-Droid releases. The build identifier moves into the release name where it's still visible in the release listing.

Deliberately not included, left as conversation in your PR:

- Runner choice (`macos-26`) — your call.
- Single-vs-double Flutter install (subosito + FVM) — design tradeoff.
- GPU validation mode in the shared scheme — could be intentional.
- `BUNDLE_PATH` redundancy in CI — informational only.
- Version-bump semver level — release coordination decision.

The corresponding inline comments on #439 have been updated to point at the specific commit here so you can map them one-to-one as you triage.


android/app/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,7 @@
     <!-- #312: Notification reminders -->
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
+    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
```
simonoppowa: Is this permission really required for the apps functionality? I couldn't find any documentation that is is needed for image_picker.

Image

> When it is allowed: Only apps where the core functionality genuinely requires persistent access to all user photos (such as cloud backup services or photo gallery apps) are approved.

jordan-lee-code: Thank you for catching this, you're right. `image_picker_android` 0.8.13+17 uses Android's Photo Picker (`PickVisualMedia`) for the gallery flow and only ever asks for `CAMERA` at runtime, so no scoped media permission is needed and your removal was the correct call. Commit dropped and force-pushed.

jordan-lee-code:
@simonoppowa feel free to merge this branch into your branch if you're happy with my suggestions, then we can get your branch into develop asap.

simonoppowa (approved):

PR #448, opened 2026-05-23
add develop flavor for gradle

seeplusplus:
This may be a skill issue on my part, so feel free to close if it is,

I started using this app and wanted to work on some bugs and features. I tested with my personal device and nuked the install of the app, because Android doesn't warn you that running a build from Android Studio is going to erase the Play Store version.  This is my first time working with a Flutter app or doing any mobile dev, but I _think_ the solution to this is to use an app flavor for development builds (I tested, and this works for me). This way I can push builds to my device with features/bug fixes that I want without harming the installation I use daily.


jordan-lee-code:
Thank you for opening this, Caleb. The "Android Studio silently overwrote my real install" footgun is real and properly horrible, and you have absolutely identified the right solution shape. A `.develop` applicationId suffix is exactly how this is meant to be solved.

I had a careful look at what the diff would mean for everyone else's day-to-day, and there are a few places that would break the moment Gradle starts requiring a flavor to be specified. None of them are hard to fix, but they would need to land in the same PR so the next person to pull `develop` doesn't hit a wall:

- `.github/workflows/default_workflow.yml:389` — the `android-build` CI job runs `flutter build apk --debug` and would fail with "no matching flavor." The android-integration-tests job around line 486 (`flutter test integration_test/`) would also fail since it builds an APK to install on the emulator.
- `tools/adb/run-branch-tests.sh:336` — the local branch-test harness runs `flutter build apk --debug`. Same issue.
- `.claude/CLAUDE.md` lines 23 and 27 — the documented local build commands need the `--flavor` flag added too.

Easiest fix is to add `--flavor full` to each of those (or `--flavor develop` for the local sideload case where that's the intent).

One thing worth thinking about on scope: the project's cross-platform rule is that any new platform-specific path gets a matching implementation on the other platform. iOS currently only has a single `Runner` scheme, so this PR creates an asymmetry where Android developers can sideload-without-nuking but iOS developers still can't. Adding parallel Xcode scheme + Info.plist configs for iOS would round it out, though that's noticeably more work than the gradle bit and could reasonably be a follow-up.

A small extra that often goes with this pattern: a flavor-specific app label and launcher icon overlay (e.g. "ONT Dev" with a different background colour) so the develop install is visually distinct on the home screen. Otherwise both icons look identical and it's easy to tap the wrong one. Optional, but if you wanted to extend, that's the natural next step.

Happy to push the CI / docs / script updates onto your branch if you'd like a hand, just say the word.

jordan-lee-code:
Pushed three more commits onto your branch so the picture's complete:

**`01bc141f` — pin CI and the local branch-test harness to `--flavor full`.** Covers the two Android steps and the two iOS steps in `.github/workflows/default_workflow.yml`, plus `tools/adb/run-branch-tests.sh`. The rule we landed on: automated paths pin to `--flavor full` so their output is unchanged (production applicationId, no `.develop` suffix), and humans typing into a terminal pick the flavor explicitly per the docs.

**`2d82cba3` — flavor-specific Android label.** The `develop` install now shows up as "ONT Develop" on the home screen, the `full` install keeps "OpenNutriTracker." Done by moving `android:label` to `@string/app_name` and providing the override in `android/app/src/develop/res/values/strings.xml`. Tapping the wrong icon at 7am is now one less worry.

**`edc45466` — iOS parallel scheme.** Same shape as Android: a new `develop` Xcode scheme targets three new `Debug-develop` / `Release-develop` / `Profile-develop` configurations that override `PRODUCT_BUNDLE_IDENTIFIER` with the `.develop` suffix and `BUNDLE_DISPLAY_NAME` with "ONT Develop." The original `Runner.xcscheme` is renamed to `full.xcscheme` so `flutter build ios --flavor full` finds it. I edited the pbxproj from Linux without Xcode to verify visually — the brace and parenthesis balance, the configuration counts, and the structure all check out, and the iOS-build CI job is what'll confirm it lands cleanly.

The launcher-icon overlay is still a nice follow-up if you ever want to circle back, but the label + bundle id distinction alone is enough to keep the two installs visually and structurally separate. Thanks again for opening this one — it scratches a real itch.

jordan-lee-code:
Quick update on the merge: develop moved a fair bit since you opened this, so I pulled it in (Simon's `ci: add ios deployment` work landed there with a new App Bundle + IPA + fastlane release pipeline) and the conflict on `Runner.xcscheme` came out the other side clean — git's rename detection moved Simon's `enableGPUValidationMode = "1"` over to `full.xcscheme` automatically. I added the same line to `develop.xcscheme` so both flavors get the same debug helper.

The merge also surfaced that Simon's new pipeline assumed an unflavored Gradle/Xcode setup, so I threaded `--flavor full` through the rest of it in commit `bd320d4f`:

- `flutter build appbundle --release`, `flutter build apk --release`, and the fastlane iOS `build` lane all now use `--flavor full` so production artefacts ship with the unsuffixed `com.opennutritracker.ont.opennutritracker` bundle ID / applicationId.
- The full flavor's Gradle output is renamed in place back to the historic `app-release.apk` and `app-release.aab` filenames. That keeps Simon's upload paths, the fastlane defaults, the website's verify-your-download instructions, and the `update-release-fingerprint` workflow working unchanged — none of them have to learn about flavors. The develop flavor keeps Gradle's default suffixed naming so a sideloaded debug build is easy to spot in `build/app/outputs/`.

Full audit done across CI workflows, both Fastfiles, Gradle, ADB scripts, and the website. `flutter analyze` is clean.

PR #449, opened 2026-05-23
fix: use correct possessive form of "what's"

seeplusplus:
- Fix issue where "Whats your current height/weight" should be "What's"


jordan-lee-code:
Thank you, Caleb, this is a lovely small fix to pick up. Confirmed the scope is right: only the English ARB carried the typo, all eight other locales translate the questions naturally so nothing else needs touching. All three files (ARB source + both generated files) are updated in lockstep with the escaping convention the existing `What's your gender?` / `What's your current weight goal?` strings already use, so `just check_intl` stays happy and the manual-maintenance rule for `lib/generated/` is respected. Reads cleanly to me.

PR #450, opened 2026-05-23
fix(settings): make Show Activity Tracking toggle hide blocks

k4black:
Hey! Could you please take a look!  
Currently there is a bug: Show Activity Tracking buttons does absolutely nothing.

No other changes, only use existing setting `showActivityTracking` to conditionally hide blocks in the main screen and add button.

| toggle On | toggle Off | toggle Off |
|---|---|---|
| Simulator Screenshot - iPhone 17 - 2026-05-22 at 23 19 12 | Simulator Screenshot - iPhone 17 - 2026-05-22 at 23 19 04 | Simulator Screenshot - iPhone 17 - 2026-05-23 at 09 35 42 |

### DIFF
--- a/lib/core/presentation/main_screen.dart
+++ b/lib/core/presentation/main_screen.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:opennutritracker/core/domain/usecase/get_config_usecase.dart';
 import 'package:opennutritracker/core/presentation/widgets/add_item_bottom_sheet.dart';
 import 'package:opennutritracker/core/utils/locator.dart';
 import 'package:opennutritracker/core/utils/navigation_options.dart';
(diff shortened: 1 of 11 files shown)


jordan-lee-code:
Thank you so much for picking this one up, Konstantin. The Show Activity Tracking toggle had quietly broken at some point and it is lovely to see it doing the thing the setting promises again. Your fix on Home + the Add bottom sheet is exactly right, and the screenshots in the description make the behaviour easy to confirm at a glance.

I pushed three small commits on top of your branch so that the rest of the app respects the same setting:

- `fix(settings): also hide activity in diary when tracking is off` threads `showActivityTracking` through `DiaryBloc` → `DiaryState` → `DayInfoWidget` the same way `showMealMacros` already flows. The diary's per-day Activity section was the one remaining surface that still showed activities after toggling off, which felt inconsistent if someone is using the app for food only.
- `fix(settings): refresh DiaryBloc when activity tracking toggles` fires `LoadDiaryYearEvent` from the settings switch handler. `DiaryBloc` is registered as a lazy singleton in the locator, so without that nudge the diary kept its stale flag until the next cold start; with it the change lands live.
- `test(activity-tracking): lock in AddItemBottomSheet visibility behaviour` adds three quick widget tests in the same shape as `intake_vertical_list_test`, so a future refactor that drops the conditional or its default trips the test before review.

I ran a full ADB pass on a real device covering both toggle states across Home, the FAB sheet, and the Diary, plus the live round-trip toggling on → off → on without restarting the app. Everything behaves as expected and the full test suite (`just test`) is green at 594 tests. `flutter analyze` is clean too.

PR #464, opened 2026-05-24
feat(add-meal): show emoji icons for FDC base-food items

k4black:
Hey! Could you please have a look on the small change adding icons for base products without images.

**Problem:** Search of the raw food is pretty hard, as eye has nothing to catch on to fast filter options (in the milk search find actual milk)

**Added:** colored emoji icons to the leading slot of FDC search results that don't have a thumbnail. OFF (Products tab) is unchanged.
 "Milk, fluid, nonfat..." → 🥛, "Apples, raw, fuji..." → 🍎,


| OLD | NEW |
|---|---|
|<img width="200" alt="Simulator Screenshot - iPhone 17 - 2026-05-24 at 12 27 20" src="https://github.com/user-attachments/assets/a4fab331-54ef-4a29-a3d5-9328149adff5" /> | <img width="200" alt="Simulator Screenshot - iPhone 17 - 2026-05-24 at 12 24 41" src="https://github.com/user-attachments/assets/f14c11c4-17ce-4cb4-bf04-fa3ef7737a69" /> |


## Design                                                                                                                                                                          
                                                           
  - Emoji-keyed map -- (with Dart compiler refuses to compile duplicate const keys).               
  - A `_variantsOf(noun)` helper auto-generates lookup keys at module init: the canonical noun, its `+s` plural, and (for multi-word entries) the no-space form and its `+s` plural.
  So "olive oil" silently matches "olive oils" / "oliveoil" / "oliveoils" without anyone listing four entries.                                                                       
  - Irregular plurals (-ies, -oes, fungus→fungi, peach→peaches) are listed explicitly.
  - ~80 canonical head nouns covering the most frequent USDA SR Legacy items.

### DIFF
--- a/lib/features/add_meal/presentation/widgets/meal_item_card.dart
+++ b/lib/features/add_meal/presentation/widgets/meal_item_card.dart
@@ -7,6 +7,7 @@ import 'package:opennutritracker/core/utils/locator.dart';
 import 'package:opennutritracker/core/utils/navigation_options.dart';
 import 'package:opennutritracker/features/add_meal/domain/entity/meal_entity.dart';
 import 'package:opennutritracker/features/add_meal/presentation/add_meal_type.dart';
+import 'package:opennutritracker/features/add_meal/util/food_emoji_resolver.dart';
 import 'package:opennutritracker/features/meal_detail/meal_detail_screen.dart';
 import 'package:opennutritracker/generated/l10n.dart';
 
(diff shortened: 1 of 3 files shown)


jordan-lee-code:
Thank you for this, k4black. It is a genuinely thoughtful usability win, and the way you scoped it to FDC base foods so OFF's branded products don't get a misleading head-noun emoji shows real care. The tests are lovely too, especially the one guarding against Dart's silent adjacent-string-literal concatenation.

Since you enabled maintainer edits, I pushed two small commits on top rather than leave you a list to action. Here is what changed and why:

**Emoji that wouldn't render everywhere.** A few were very recent Unicode additions that show as missing-glyph boxes on the older end of the versions we still support (the app targets iOS 15.5, with no raised Android `minSdk`), which is worse than the outline icon the feature replaces. So: brown mushroom (Unicode 15.1) became plain `🍄`; pea pod (15.0) folded its legume nouns into the green-circle entry that already held "pea"; and ginger root (15.0) became the herb `🌿` for the spice group. That last one is only a broad-availability stand-in, so please swap it for something you prefer if it doesn't sit right.

**Screen readers.** The emoji is decorative since the card title already names the food, so I wrapped it in `ExcludeSemantics`. Otherwise TalkBack and VoiceOver announce "glass of milk" just before reading the actual title.

**A couple of tidies.** The resolver's doc comment now matches the FDC-only call site, and I hand-wrapped the one map line that ran past our 120-column width.

Your original commit is untouched underneath, and the tests and analyzer still pass. Thank you again for picking this up.

PR #247, opened 2025-06-17
[Test]:. Implementing Unit Tests for MealDetailBloc

dudaa28:
# Unit Tests for MealDetailBloc

- **Tested Method**: `UpdateKcalEvent`  
- **Tested Class**: `MealDetailBloc`  
- **Test Location**: [`OpenNutriTracker/blob/PTOSS-03/test/unit_test/meal_detail_bloc_test.dart`](https://github.com/OpenNutriTracker/blob/PTOSS-03/test/unit_test/meal_detail_bloc_test.dart)

Unit Tests for MealDetailBloc with MC/DC Coverage and Nutritional Calculation Validations

## Description
This PR adds a comprehensive suite of unit tests for the `MealDetailBloc`, ensuring validation of input fields, unit conversions, and nutritional calculations related to a meal.

The tests follow the Modified Condition/Decision Coverage (MC/DC) methodology to cover logical decisions affecting the app flow, focusing on:

- Validation of required fields (`selectedUnit` and `selectedTotalQuantity`)
- Correct conversion between measurement units (`serving`, `oz`, `fl.oz`, `g`)
- Accurate calculation of nutritional values based on the converted quantity
- Handling of parsing errors and invalid inputs
- Scenarios with null or zero nutrient values
- Correct conversion when dealing with decimal commas and large quantities

## Changes Included
- Tests validating empty or invalid inputs
- Tests for quantity conversion based on the selected unit
- Tests for accurate nutritional value calculations (energy, carbs, fat, protein)
- Tests ensuring expected behavior with null, zero, and Brazilian decimal number formats
- Additional tests for edge cases (very large quantities)
- Test coverage ensuring MC/DC combinations are adequately tested

## How to Test
- Run `flutter test` in the project root to execute all unit tests.
- Confirm all tests pass without errors.
- Check coverage reports to ensure key conditions and decisions are covered.

## Technical Context
- Nutritional values are normalized per unit quantity (e.g., kcal per 1g)
- Unit conversions use the `UnitCalc` utility class
- Tests focus on calculation logic and input validation simulating user entries

## Checklist
- [x] Code is commented and organized
- [x] Tests created for main use cases and edge cases
- [x] Tests pass in local environment
- [x] Documentation updated as needed

## Related Issues
- Not applicable

PR #465, opened 2026-05-24
fix: restore the Nix dev flake dropped in the #247 merge

### LINKED ISSUE #462: flake.nix dropped?
Hi, it seems the flake.nix was dropped in 50f62eb. Is there any specific reason? I'm on nixos too and would love to use it to build ONT debug for testing.

jordan-lee-code:
Closes #462.

STSchiff opened #462 to ask where the `flake.nix` went, and it turned out to be a quiet casualty of a merge rather than a deliberate removal. The `flake.nix` and `flake.lock` were both present on the develop side of `50f62eb` (the PR #247 merge), and the conflict resolution lost them. The `.envrc` with its `use flake` line survived, so anyone with direnv has been dropping into a broken shell ever since. Thank you for catching this, and sorry it bit you in the first place.

I did not want to just restore the files as they were, because develop has moved on a long way since August 2024. The old flake pinned Android build-tools `30.0.3` and platforms up to `34`, which will not build the current app. So this restores them brought up to date:

- Android `platformVersions` `34/35/36` and `buildToolsVersions` `34/35/36.0.0`, matching `compileSdk`/`targetSdk 36` in `android/app/build.gradle` and AGP 8.11.1 in `android/settings.gradle`. JDK stays at 17.
- `nixpkgs` pinned to `nixos-unstable` so the bundled Flutter is recent enough for the Dart `>=3.11.0` constraint in `pubspec.yaml`.
- A `shellHook` that points Gradle at the SDK own `aapt2`, since the Maven-downloaded binary will not exec under the NixOS loader.

One honest caveat for review: Flutter comes from nixpkgs here, not the FVM-pinned 3.41.7 in `.fvmrc`, so the exact version will drift with the channel. The flake comments explain how to layer FVM on top when an exact match matters. I also do not have a Nix environment to build this in, so I would be really grateful if you (@chrisheib) could confirm `nix develop` resolves and a debug APK builds on your NixOS box before this merges. If `composeAndroidPackages` complains about a missing build-tools `36.0.0` or wants an NDK, that is the first thing to adjust.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,59 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1724395761,
+        "narHash": "sha256-zRkDV/nbrnp3Y8oCADf5ETl1sDrdmAW6/bBVJ8EbIdQ=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "ae815cee91b417be55d43781eb4b73ae1ecc396c",
+        "type": "github"
+      },
+      "original": {
+        "id": "nixpkgs",
+        "type": "indirect"
+      }
+    },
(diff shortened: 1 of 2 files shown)


chrisheib:
Thanks for this! I was working on this today too (got the debug build running, thanks again for your work, some great features in there!). 

I took a slightly different route, going with [nix-direnv](https://github.com/nix-community/nix-direnv) instead of pure flakes. 

It always pulls the latest packages from nix-unstable, giving a bit less control over the versions while being quite a lot more easy by avoiding lock conflicts.

Pushed them here: https://github.com/chrisheib/OpenNutriTracker/commit/287931a8f4fb504c01f39f060595ad0be6ad79eb 

If you want I can open a PR, or try to learn flake direnv if you prefer that.

jordan-lee-code:
Thank you for digging into this and getting a debug build running on your own box, that hands-on testing is exactly what I could not do myself. Since I do not use flakes and you are the one who actually runs nix here, it makes far more sense to go with your nix-direnv version in #469 than with my flake. Closing this in favour of that one, and really grateful you picked it up.

PR #466, opened 2026-05-24
fix(scanner): own + dispose mobile_scanner controllers, observe app lifecycle

k4black:
Hey! Small bugfix for the barcode scanning.

Five screens use `mobile_scanner`: `scanner_screen.dart`, three `import_*_scanner_screen.dart` flows, and `_EditMealBarcodeScanPage` in `edit_meal_screen.dart`. All pass an externally-owned `MobileScannerController` — but:

- **`MobileScanner` doesn't dispose external controllers**, only `stop()`s them. Four screens created the controller inline in `build()` and never disposed → leaked
  `AVCaptureSession` until GC.
- **`scanner_screen.dart` re-created the controller on every rebuild**, so the AppBar torch / camera-flip buttons could end up bound to an orphan controller via                   
  `ValueListenableBuilder` (silent no-op).
- **No app-lifecycle handling** → backgrounding the app leaves the camera session running.

## Fix
Same pattern in all five screens:

1. Hoist `MobileScannerController` into a `late final` `State` field, init in `initState`.
2. Dispose in `dispose()`.
3. Mix in `WidgetsBindingObserver`; in `didChangeAppLifecycleState`:
     - `paused` / `hidden` / `detached` → `stop()`
     - `resumed` → `start()`
     - `inactive` → no-op
     
     
---

Did smoke test on the iphone

### DIFF
--- a/lib/features/edit_meal/presentation/edit_meal_screen.dart
+++ b/lib/features/edit_meal/presentation/edit_meal_screen.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:math' as math;
 
 import 'dart:io';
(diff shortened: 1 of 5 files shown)


jordan-lee-code:
Ran this on a physical Pixel 6 Pro, on a `develop` debug build off your branch, so the real camera and flash were in play. The torch fix only shows up on hardware, and it holds up nicely. Thank you for the careful write-up, it made the screens easy to reason about.

Driving the main Scan Product screen (`scanner_screen.dart`) end to end, with `CameraService` / `flash_lm3644` logcat as ground truth:

- **Torch now drives the real controller.** Tapping it logged `flash_lm3644_base.cc: flash torch on, 125.15mA each led` and the preview lit up; the AppBar icon flipped to the active-flash glyph; tapping again gave `flash off`. The orphan-controller no-op is gone.
- **Flip camera** switched cleanly: `Disconnected client for camera 0` then `Created a device session for camera 1`.
- **Lifecycle observer works both ways.** Backgrounding closed the session (`CameraService: disconnect: Disconnected client for camera 0`, `onCameraClosed`); resuming reopened it to a fresh `First frame done`.
- **Disposal / re-entry is clean.** Backing out released the camera and re-entering reacquired it with no `CAMERA_IN_USE` and no exception, which is the leak fix landing.

I exercised the main scanner screen directly rather than walking the three import scanners and the edit-meal scan page as well. They apply the identical controller-as-field plus dispose plus lifecycle pattern, so I'm confident in them, and your iPhone smoke test covers the other platform.

One tiny non-blocking note: `_EditMealBarcodeScanPage` keeps the `_controller` name and a plain `final`, while the other four moved to `_cameraController` / `late final`. Behaviour is identical, purely cosmetic if you ever want them uniform.

I also liked the `if (!_controller.value.hasCameraPermission) return;` guard at the top of `didChangeAppLifecycleState` — it sidesteps `start()`/`stop()` firing before permission is granted, which is exactly the window those would otherwise throw in.

Verified working as described.

PR #467, opened 2026-05-24
Preselect quantity in MealDetailBottomSheet

chrisheib:
When tapping into the quantity field, the 1 thats usually there usually needs to get removed manually. 

This adds an listener that highlights the whole field, so the number can be replaced straight away. 

This seems like a somewhat convoluted implementation, I'm sure there are more elegant ways. 

Image


jordan-lee-code:
Thank you for this, it is a lovely little quality-of-life fix and I noticed the "1" needing manual deletion myself. No need to second-guess the implementation either, the post-frame callback with the `mounted` and `hasFocus` guards is exactly the standard, robust way to do this. Pairing the focus listener with `onTap` is the right call too, since it covers both first focus and re-tapping an already-focused field. Merging shortly.

PR #469, opened 2026-05-24
Restore nix dev environment with nix-direnv

chrisheib:
Alternative version of the devshell. 

See #465 , #462

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,2 @@
-use flake
+# https://github.com/nix-community/nix-direnv
+use nix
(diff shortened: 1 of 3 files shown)


jordan-lee-code:
This is lovely, thank you. I especially like that Flutter still comes from FVM here, so we keep the `.fvmrc`-pinned 3.41.7 rather than drifting with the channel, and you have already confirmed a debug build works on NixOS, which is the part I could not verify myself. This is the approach we are going with over my flake PR #465, which I am closing in its favour.

PR #459, opened 2026-05-23
Release v1.4.0+51 — Recipes ingredient scanner, fasting timer, water tracker, weight history, Material You, sk locale, and CI/CD pipeline

jordan-lee-code:
> @simonoppowa — this is the next release rolling up everything that's landed on `develop` since `1.3.1+50` (last release PR #381). Once you've had a look, please **merge into `main` and cut a build for the App Store and Play Store** so users get the new water tracker, fasting timer, weight history, Material You theming, recipes ingredient scanner, Quick add, and the rest of the work below. Happy to walk through any section if useful.

## Review feedback addressed (#470)

Thank you for the careful review. Your comments are worked through in #470, which merges into `develop` and so flows up into this release:

- The diary nutrient panel now shows a *limit* qualifier on the ceiling nutrients (sodium, saturated fat, sugar) so they no longer read as targets.
- IOM sources are cited inline: the 2019 sodium/potassium update for the sodium CDRR and potassium AI, and the 2011 calcium/vitamin D report. Those adult calcium and vitamin D figures are the current RDAs, which superseded the 1997 AI values, so they stay as-is with the citation at the line.
- The low-kcal floor comment now cites the Harvard Health reference.
- Notification channel names and descriptions are localized across all nine locales, including the reboot-restore path.
- The "2024 Compendium" attribution is hidden for custom activities.

## Summary

Release **v1.4.0+51** (from `1.3.1+50`). Substantial feature release plus the full CI/CD release pipeline you have been building on `integrate-cicd-deployment` — TestFlight and Play Store internal deployment now run from a green workflow on every release merge. ~83 non-merge commits across ~404 files (+40K / -6.2K LoC overall; +30.2K / -5.8K in `lib/`).

### Water tracker (#418, closes #32)
- Home-screen water chip with quick-add increments and an editable daily goal.
- Profile reflects the same goal and history.
- Undo for the last entry on the home chip.
- New Hive type `WaterIntakeDBO` in its own `WaterIntakeBox`.

### Fasting timer (#432)
- Optional intermittent-fasting timer with a content-warning gate the first time you open it.
- Home chip surfaces the time remaining; tap to start, pause, or reset.
- Completion notification fires when the chosen window elapses (via `flutter_local_notifications`).
- New Hive type `FastingSessionDBO` in its own `FastingBox`.

### Weight history + target weight (#400, #406)
- Capture weight during onboarding and on demand from the profile screen.
- Trend chart on profile with a dashed line at the user's target.
[...]

PR #470, opened 2026-05-25
Address review feedback on the v1.4.0 release PR

jordan-lee-code:
This works through the review comments Simon left on the v1.4.0 release PR (#459). It targets `develop` so the fixes flow up into that release once merged.

- **Diary nutrient panel:** ceiling rows (sodium, saturated fat, sugar) now carry a small localized "limit" qualifier, so a reference you are meant to stay under no longer reads as a target. It rides on the existing `excessMatters` flag, so the three ceilings stay consistent.
- **DRI citations:** the 2019 IOM sodium/potassium update is now cited inline for the sodium CDRR and potassium AI, and the 2011 calcium/vitamin D report for calcium and vitamin D. I kept the 2011 RDA figures rather than the 1997 AI values, with a note at the line explaining why (see below), so the citation sits where the next reader will look.
- **Low-kcal floor:** the comment now cites the Harvard Health page Simon linked, since the earlier NIH / Mayo Clinic attribution could not be verified.
- **Notification channels:** the channel name and description for the daily reminder and the fasting timer are now localized across all nine languages, including the reboot-restore path in `main()`. They surface in the system settings, so they read in the user's language now.
- **Activity detail:** the "2024 Compendium of Physical Activities" attribution is hidden for custom activities, which are user-entered.

A note on vitamin D and calcium: Simon read the source as listing AI values. Those are the 1997 IOM figures; the 2011 update established an RDA for both (vitamin D 15/20 µg, calcium 1000/1200 mg), which is what NIH ODS publishes today. I kept the 2011 RDA so we are not showing older, lower numbers, and documented the 1997 to 2011 change inline with the source link.

`flutter analyze` is clean and `just test` passes (597 tests).

### DIFF
--- a/lib/core/utils/bounds/ranges_const.dart
+++ b/lib/core/utils/bounds/ranges_const.dart
@@ -12,12 +12,14 @@ class Ranges {
   /// Soft floor below which a computed daily kcal goal is flagged as low
   /// enough that the app surfaces a non-blocking warning encouraging the
   /// user to consult a professional. Aligned with the commonly-cited
-  /// NIH / Mayo Clinic minimums for sustained healthy intake (≈1500 kcal
-  /// for adults with a testosterone-typical hormonal profile, ≈1200 kcal
-  /// otherwise). Non-binary users on the averaged midpoint, or who have
-  /// not yet chosen a calories profile, use the lower (female) floor
-  /// because the averaged TDEE is itself a midpoint and the lower bound
-  /// sits more naturally beside it.
+  /// minimums for sustained healthy intake (≈1500 kcal for adults with a
+  /// testosterone-typical hormonal profile, ≈1200 kcal otherwise). These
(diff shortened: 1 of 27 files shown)


jordan-lee-code:
**Verified on a device (debug `develop` build).**

- Custom-activity detail no longer shows the "2024 Compendium of Physical Activities" attribution, while a built-in activity (bicycling) still does, so the `!isCustom` guard is scoped to exactly the right case.
- The nutrient panel shows the italic *limit* qualifier on sodium, saturated fat, and sugar, and not on the target nutrients (fibre, calcium, iron).
- Switched the app to German and enabled the daily reminder: it schedules cleanly with the localised channel strings. Android only materialises the channel when the first notification posts, so the rendered name in the system settings isn't observable until then, but the German path runs without issue.

`flutter analyze` is clean and `just test` passes (597 tests).

PR #483, opened 2026-05-28
Release v1.5.0 (build 58): switch food search to Search-a-licious

jordan-lee-code:
This brings `develop` to `main` for the v1.5.0 release. main moves from 1.4.1+57 to 1.5.0+58, a single build-number increment and a minor bump for the new features below.

## Food search
- OFF text search now runs on Search-a-licious, with full-fidelity hydration so results keep complete product detail, and products sold in your country are gently boosted so local shelf items surface first.
- FDC base-food results without a photo now show a colored emoji in the leading slot (🥛 for milk, 🍎 for apples) so the list is far easier to scan. Thank you to Konstantin Chernyshev (@k4black).

## Logging
- The + button in each meal section stays pinned to the left of the list now, so it's always reachable no matter how many items you've logged. Thank you to @h4rib0.
- Tapping into the quantity field preselects the value, so you can type straight over it, and the meal-detail bottom sheet has a cleaner top-only border. Both thank you to @chrisheib.

## Transparency
- The sources behind the medical calculations are easy to find on first run now, with entry points in onboarding and Settings.

## Under the hood
- Restored the Nix dev shell for contributors (@chrisheib), added an `.env.example` template (@k4black), and made the GitHub release name use the pubspec build number consistently so the release name, tag, and published artifacts all agree.

---

@simonoppowa this is the v1.5.0 release candidate. When you have a moment, could you cut the App Store (TestFlight) and Play Store builds from it once it's merged? Thank you.


chrisheib:
In local dev the Food-search no longer completes (only dummies 'da' and 'ab' are found). Are there new env settings required?

jordan-lee-code:
> In local dev the Food-search no longer completes (only dummies 'da' and 'ab' are found). Are there new env settings required?

Probably an unintended consequence of #483. I'll take a look

jordan-lee-code:
Had a proper look, and the short version is that the search didn't actually regress, sorry the change in behaviour was confusing.

The "Food" tab searches a Supabase FoodData Central database that the maintainer (Simon) hosts on behalf of the published app, so it needs real `SUPABASE_PROJECT_URL` and `SUPABASE_PROJECT_ANON_KEY` values in your local `.env`. If those are still the placeholders from `.env.example` (`PROJECT_URL` / `ANON_KEY`), the Supabase request can't resolve and the search quietly falls back to local results only, which is why you're seeing just your own cached/test entries.

Here's the part that explains the "it worked before" feeling: a fix in this release stopped the Products (Open Food Facts) cache from bleeding into the Food tab. Previously, when the Food tab's Supabase call failed, it would borrow your cached OFF products and display those, so the tab looked populated even without Supabase configured. Now that the two sources are properly separated, an unconfigured Food tab honestly shows nothing rather than the wrong results. So nothing about FDC search broke; the change just stopped masking a missing local config.

Since the shared Supabase project isn't something we can hand out credentials for, the way to run the Food tab locally is to stand up your own small Supabase FDC project and point your `.env` at it. We have a guide written for exactly this: [Self-hosting the Supabase FDC database](https://github.com/simonoppowa/OpenNutriTracker/blob/main/docs/supabase-fdc-self-hosting.md) (it links onward to the table setup steps). Once your `.env` has your project's URL and `anon` key, run `just build` to regenerate `env.g.dart`. The Products tab keeps working without any of this, since it uses a public endpoint.

chrisheib:
I see, thank you for looking into this.

PR #487, opened 2026-05-28
Navigate to main screen after quick add

chrisheib:
I'm not sure if this is true for all users, but I'd prefer to navigate back to the main page after adding a meal via 'quick add'. I normally don't add multiple quick entries, but rather calculate a portion in advance so I only have to enter a single number. 

@k4black What do you think, would this be a sensible default?


chrisheib:
@jordan-lee-code : Do you prefer to branch develop off the main branch again after version merges, or let it be on it's own in parallel without pulling main in?

jordan-lee-code:
As commits to main directly are very rare outside of develop, develop is usually in line, if main is changed directly develop is caught up right away.

chrisheib:
I reset the pr to be in line with dev.

Issue #495, opened 2026-05-30
Custom food seems bugged


legoraft:
I tried adding a custom base food item (tomatoes in this case), but when adding the food in the normal tracker the serving size seems bugged. I think I've set the serving sizes correctly, although the UI isn't entirely clear on what the difference is between meal and serving size. If this is a user-issue I'd love to hear!

<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/a404e0f9-7b0a-4741-887e-00559a51e0af" />

_Serving size doesn't appear correctly_

<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/9a5c6a02-fe54-4156-b787-f52134ce719f" />

_Custom food settings_

jordan-lee-code:
Thank you for this, Legoraft, and for the clear screenshots — they made it easy to track down. It was a real bug: the serving size was saving the quantity in place of its unit, so the serving option showed up blank instead of "Serving (50 g)". Fixed in #496, with an extra fix so imperial serving sizes stop drifting when you reopen the meal.

You were also right that the meal-vs-serving wording was unclear. Those Advanced fields are now **Pack size**, **One serving** and **Nutrition is per**, and the unit selector has moved to the top so the values below follow it.

This is merged to `develop` and will ship with the next release. Thanks again for taking the time to report it so carefully.

legoraft:
Awesome! I'm looking forward to the update getting in app stores

Issue #472, opened 2026-05-25
[Feature Request] Add activity by entering consumed calories and spent time


ferkulat:
I have an app counting my steps when I walk and estimating the consumed calories. If I assume its values are right, then non of the presented walking activities match the consumed calories when entering the spent time. Its either way to low or too high.

I would like to be able to create/add a custom activity, where I enter the spent time and the consumed calories myself.

Is that possible?

Thank you in advance

jordan-lee-code:
Merged to develop in #490. In 1.6.0 there will be a Quick Add button when adding an activity that'll allow you to just enter the activity name, calories burned and optionally how long you did it for. Thanks for raising the feature request!

Issue #471, opened 2026-05-25
Add support for multiple profiles


nalladev:
Allow users to create and manage multiple profiles within the app. 

Suggested functionality:
- Add multiple profiles
- Switch between profiles easily
- Edit or delete profiles
- Store separate settings/data per profile

Example use case:
Parents could manage separate profiles for multiple kids.

jordan-lee-code:
Hi @nalladev, thank you for suggesting this. Multiple profiles are now implemented in #489.

Each profile keeps its own diary, weight history, activity and goals, fully separate from the others, so the parent-managing-kids case you described works as you'd hope. From the Profile tab you can add, rename, give a picture to, switch between and delete profiles, and anyone upgrading from a single-user install keeps all their existing data with no migration.

A few things are deliberately shared across profiles rather than duplicated, so switching to a child's profile doesn't flip your theme or lose your recipe library: the app settings (theme, language, units, notifications) and the custom-meal, recipe and activity-template libraries. The personal nutrition goals and all the daily logs stay private to each profile. There's also a "Copy to profile" action for sending a logged meal to another profile.

It's lined up for the next release, likely 1.6.0. Thanks again for the thoughtful request.

Issue #165, opened 2025-03-09
Camera View rotates with phone rotation


lajawi:
When using the camera to scan a product, if the phone isn't right side up, the camera view in the app itself is rotated too. The UI correctly rotates with the phones rotation, but the camera view shouldn't.

|Image|Description|
|---|---|
|![Image](https://github.com/user-attachments/assets/c67ec854-cc2c-46af-b8cd-7f2d7bf53048)|Phone is right side up, UI *and* camera view are correct (arrow points up as it did in the real world).|
|![Image](https://github.com/user-attachments/assets/6c336c58-553a-4d7f-a4df-d356ec166025)|Phone is rotated to the left (selfie camera to the left, speakers to the right), as you can see the arrow doesn't point up any more. The camera view has rotated too.|

## Information

Device: iPhone SE 3 (iPhone 14,6)
Software: iOS 18.3.1
App version: 0.9.0, build 38 (via TestFlight)

JEdits20:
same issue on android

jordan-lee-code:
Hi @lajawi — confirming this is still outstanding on `develop`.

Code location: `lib/features/scanner/scanner_screen.dart:118-137` uses the `mobile_scanner` package directly with no orientation lock or counter-rotate logic. Likely fix: lock the scanner screen to portrait via `SystemChrome.setPreferredOrientations([DeviceOrientation.portraitUp])` on entry and restore on exit; or counter-rotate the preview transform with `MediaQuery.orientationOf(context)`.

Labelled `bug`.

jordan-lee-code:
Thank you @lajawi for the clear report and the side-by-side photos, and @JEdits20 for confirming it on Android. They made the rotation behaviour easy to pin down.

This is fixed in #492: the scanner now stays upright on phones, with a toggle if you'd rather let it rotate, which is handy on tablets. It's expected in the 1.6.0 release.

PR #521, opened 2026-07-13
Release v2.0.0 – UI overhaul, multi-source food database, profiles & revamped Trends

### LINKED ISSUE #316: Feature: Add BLS (Bundeslebensmittelschlüssel) as a local food data source

simonoppowa:
## OpenNutriTracker 2.0

This release brings a full visual redesign, a new multi-source food database backend, multiple profiles, and a much deeper Trends experience.

### 🎨 UI overhaul
- New "calm friendly-flat" design language across every screen, with an AA contrast pass
- Restructured navigation: settings move into the new **You** tab, grouped into labelled categories — no more scattered gear icons
- Diary swipe between days with jump-to-today, full-width meal & activity rows, friendlier empty states
- Recipes get their own screen with easy access from You

### 🍎 Food database & search
- Food search migrated to the new Supabase multi-source backend (FDC, Open Food Facts, and more)
- Food database selection in Settings
- FDC-style serving labels with localized household units, and sensible default units for backend foods
- Resilient Open Food Facts text search and more reliable search input with clearer result-state feedback
- Disclosure hint for machine-translated food names
- Add-meal search consolidated into one combined results list, defaulting to recently added foods
- Recent foods on the Add sheet ranked by frequency, plus quick-quantity preset chips — logging in two taps

### 👤 Multiple profiles (#471)
- Create and switch between multiple user profiles

### 📊 Trends
- Time ranges including a new **All** range spanning full history, goal-vs-actual, macro averages, and week-over-week deltas with tappable chart points
- Daily water trend card, current/longest streaks, and a weight-change outlook
- Log weight straight from the Trends card, with date picking to backfill history
- Full weight-history chart

### ⚖️ Units
- Independent preferences for food, height, and body-weight units
- Stones for body weight, feet + inches for height
- Onboarding now asks for food units explicitly; new "Other options" onboarding page

### ⚡ Quick add
- Quick add for activities with an optional name field (#472)
- Log weight directly from the main screen (thanks @chrisheib!)

### ♿ Accessibility
- Screen-reader labels on the calorie ring and Trends charts
- Layouts hold up at large text scales: wrapping titles, non-overflowing chips, macro tiles, and dialogs

### 🐛 Fixes
- Custom foods store the selected unit on the serving, not the quantity (#495)
- FDC nutrients keyed on id instead of number for correct energy values
- Scanner camera preview stays upright in landscape (#165)
- Recipe tags survive duplication; weight-chart windows align across Trends
- Search bar no longer overflows in landscape; dropdown descenders no longer clip at large text sizes

### 🔧 Maintenance
- CI falls back to Temurin when the Zulu JDK download fails
- Dependency bumps (jwt for Android/iOS) @


lib/core/utils/custom_text_input_formatter.dart
```diff
@@ -1,8 +1,15 @@
...
-  static List<TextInputFormatter> doubleOnly() => <TextInputFormatter>[
-        FilteringTextInputFormatter.allow(RegExp(r'[0-9]+[,.]{0,1}[0-9]*')),
+  /// Allows a decimal number with at most [maxDecimals] places (comma or
+  /// period), normalising the separator to a period. Defaults to two decimals
+  /// so every food-quantity field accepts the same precision; pass a different
+  /// limit only where a field genuinely needs more.
+  static List<TextInputFormatter> doubleOnly({int maxDecimals = 2}) =>
+      <TextInputFormatter>[
+        FilteringTextInputFormatter.allow(
+          RegExp('^\\d+([.,]\\d{0,$maxDecimals})?\$'),
```
jordan-lee-code: **High.** This anchored pattern (`^...$`) with `FilteringTextInputFormatter.allow` blanks the whole field rather than rejecting the bad keystroke, because `allow` keeps only matched substrings and an anchored regex matches all-or-nothing. `"12.345"` and `"1,234"` both filter to `""`. Since `doubleOnly()` is on every quantity/macro/kcal field, going one digit too precise wipes what you'd typed. A `TextInputFormatter.withFunction` that returns `oldValue` on reject would cap decimals without the wipe.

lib/core/domain/usecase/delete_all_user_data_usecase.dart
```diff
@@ -2,39 +2,42 @@ import 'package:logging/logging.dart';
...
     // Closing Sentry first stops any in-flight queue from referencing
     // boxes mid-clear. The user can re-enable crash reporting later from
     // Settings; nothing here re-opens the SDK on its own.
     await Sentry.close();
 
+    // Only the active profile's own data is cleared. The shared content
+    // libraries (custom meals, recipes, activity templates) and the shared
+    // app settings belong to every profile, so they're deliberately left
+    // alone — wiping them here would take them from the other profiles too.
     await Future.wait([
```
jordan-lee-code, line 32: **High.** This now clears only the active profile's eight boxes, but the tile and dialog still say "delete all my data", so it reads as a device wipe when it's a per-profile reset. We're keeping the per-profile scope deliberately, so this is really a copy change: retitle the tile and confirm dialog to say it resets the current profile. It pairs with the config resurrection note on `config_data_source.dart`.

lib/core/data/data_source/config_data_source.dart
```diff
@@ -2,129 +2,178 @@ import 'package:hive_ce_flutter/hive_flutter.dart';
...
+
+  ConfigDataSource(this._db);
+
+  Box<ConfigDBO> get _appBox => _db.appConfigBox;
+  Box<ConfigDBO> get _profileBox => _db.configBox;
+
+  /// Builds a detached config from the shared app box with the personal
+  /// fields overlaid from the active profile's box. Detached (via JSON) so
+  /// callers can't accidentally mutate a stored instance.
+  ConfigDBO _readMerged() {
```
jordan-lee-code, line 41: **Medium.** `_writeBoth` also stores the personal goal fields into the shared `appConfigBox`, and `deleteAll` clears only `configBox`. After a reset, `_readMerged` finds `profile == null`, skips the overlay, and hands back the deleted profile's kcal adjustment and water goal until the next relaunch. Resetting those fields in `appConfigBox` during the reset (or falling back to `ConfigDBO.empty()` when the profile box has no key) closes it.

lib/core/domain/usecase/send_intake_to_profiles_usecase.dart
```diff
@@ -0,0 +1,155 @@
...
+        }
+        await addTrackedDay.addDayCaloriesTracked(day, copy.totalKcal);
+        await addTrackedDay.addDayMacrosTracked(
+          day,
+          carbsTracked: copy.totalCarbsGram,
+          fatTracked: copy.totalFatsGram,
+          proteinTracked: copy.totalProteinsGram,
+        );
+      }
+    } finally {
```
jordan-lee-code, line 134: **Medium.** `openScopedBox` returns the already-open instance if that box is open, so if the active profile were ever passed as a target, this `finally` would close its live boxes out from under the app (the getters don't re-check `isOpen`). The sheet filters the active profile out today, so it isn't reachable, but a one-line `if (target.id == activeProfileId) return;` here would keep the usecase safe regardless of caller.

lib/core/utils/hive_db_provider.dart
```diff
@@ -65,59 +163,144 @@ class HiveDBProvider extends ChangeNotifier {
...
+  Future<void> openProfileBoxes(String profileId, String boxSuffix) async {
+    _activeProfileId = profileId;
+    _activeBoxSuffix = boxSuffix;
+    await _openActiveProfileBoxes();
+  }
+
+  /// Closes the current profile's box-set and opens [profileId]'s. The
+  /// `_switching` guard makes any racing read fail loudly rather than
+  /// hitting a half-open box.
+  Future<void> switchProfile(String profileId, String boxSuffix) async {
```
jordan-lee-code, line 203: **Medium.** `_switching` guards racing reads but not two overlapping `switchProfile` calls: the early-return compares the old `_activeProfileId` before it's updated, so a fast double-tap can interleave two close/open cycles on the same fields. A small in-flight guard that awaits or ignores the second call would settle it.

lib/features/trends/presentation/trends_page.dart
```diff
@@ -0,0 +1,622 @@
...
+            Text(
+              _projectionLabel(context, projection),
+              style: text.bodySmall?.copyWith(color: palette.textMuted),
+            ),
+          const SizedBox(height: Dimens.spacing12),
+          WeightTrendChart(
+            entries: entries,
+            bodyWeightUnit: bodyWeightUnit,
+            targetWeightKg: targetWeightKg,
+            windowDays: rangeDays < 30 ? 30 : rangeDays,
```
jordan-lee-code: **Medium.** The weight card forces `windowDays: rangeDays < 30 ? 30 : rangeDays`, and `weightProjection` just above runs over the full history regardless of range, so on the 7d chip the chart shows 30 days and the rate headline is an all-time slope. The release notes mention the weight windows aligning, so this looks like the last spot to bring into line with the calorie/water windowing.

lib/features/trends/presentation/trends_page.dart
```diff
@@ -0,0 +1,622 @@
...
+    );
+  }
+}
+
+class _MacrosTrendCard extends StatelessWidget {
+  final List<TrackedDayEntity> days;
+  final AppPalette palette;
+  const _MacrosTrendCard({required this.days, required this.palette});
+
+  double _avg(Iterable<double?> values) {
```
jordan-lee-code, line 457: **Medium.** `_avg` divides by the tracked-day rows that exist (days logged), so a 90-day range with 20 logged days reports the 20-day mean under a "daily average over the window" heading, while the calorie line above is zero-filled across all 90. `whereType<double>()` also means each macro can use a different denominator. Worth deciding whether the average is per-logged-day or per-window-day and matching the two cards.

lib/features/trends/presentation/trends_page.dart
```diff
@@ -0,0 +1,622 @@
...
+    final onTrackDays = <DateTime>{
+      for (final d in days)
+        if (onTrack(d)) DateTime(d.day.year, d.day.month, d.day.day),
+    };
+    final stats = streakStats(onTrackDays, windowStart, today);
+
+    // Week-over-week on-track delta (this week vs the prior week).
+    final thisWeek =
+        days.where((d) => !d.day.isBefore(weekStart) && onTrack(d)).length;
+    final delta = thisWeek - priorWeek.where(onTrack).length;
```
jordan-lee-code: **Medium.** For a brand-new user the prior week has no rows, so `delta = thisWeek - 0` shows a big green "+7" as if they improved against a real baseline. Suppressing the chip when the prior week is empty reads more honestly.

lib/features/add_meal/data/data_sources/sp_food_data_source.dart
```diff
@@ -0,0 +1,154 @@
...
+    if (enabled.length == SPConst.foodSourceDisplayNames.length) return null;
+    return enabled;
+  }
+
+  Future<List<SpFoodDTO>> _searchEnglish(
+    SupabaseClient client,
+    String searchString,
+    List<String>? enabledSources,
+  ) async {
+    var query = client.from(SPConst.foodSummaryTable).select().textSearch(
```
jordan-lee-code, line 82: **Medium.** Both the English and localized paths do `textSearch(...).limit(20)` with no `.order()`, so PostgREST returns an arbitrary 20 in physical order and a common query like "milk" can push the obvious match off the slice. Ordering by `ts_rank` before the limit would put the best matches in the returned set.

lib/features/profile/presentation/widgets/profile_switcher_header.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'package:opennutritracker/generated/l10n.dart';
+
+/// Tappable header at the top of the Profile tab showing who is active
+/// and opening the switcher sheet.
+class ProfileSwitcherHeader extends StatelessWidget {
+  const ProfileSwitcherHeader({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    final activeProfile = locator<GetProfilesUsecase>().getActiveProfile();
```
jordan-lee-code, line 17: **Medium.** The header is a `const` widget reading `getActiveProfile()` in `build`, and the editor saves a rename without firing a reload event, so the You-tab header keeps the old name and avatar until a relaunch or switch. Emitting a reload after save (and dropping the `const`) refreshes it.

lib/features/diary/diary_page.dart
```diff
@@ -108,16 +110,31 @@ class _DiaryPageState extends State<DiaryPage> with WidgetsBindingObserver {
...
-            } else if (state is CalendarDayLoaded) {
-              return DayInfoWidget(
+        if (!DateUtils.isSameDay(_selectedDate, _currentDate))
+          Align(
+            alignment: Alignment.centerRight,
+            child: Padding(
+              padding: const EdgeInsets.only(right: Dimens.spacing8),
+              child: IconButton(
+                icon: const Icon(Icons.today_rounded),
+                onPressed: () => _onDateSelected(_currentDate, trackedDaysMap),
```
jordan-lee-code, line 123: **Medium.** `_currentDate` (defined as a raw `DateTime.now()` in the state) drives this jump-to-today and the swipe clamp below, but `CalendarDayBloc`/`DiaryBloc`/`HomeBloc` resolve "today" through the day-start offset. With a 4am offset, between midnight and 4am Home treats yesterday as today while the diary jumps/clamps to the new date, so the two screens disagree. It's also captured once, so past midnight the jump button points at yesterday. Deriving it from the same day-boundary calc and refreshing on resume fixes both.

lib/core/utils/bounds/validator.dart
```diff
@@ -58,6 +58,30 @@ class ValueValidator {
     return isImperial ? UnitCalc.lbsToKg(weight) : weight;
   }
 
+  /// Combines a feet + inches entry into a bounds-checked centimetre value.
+  /// Returns null when the parts are missing or the combined height falls
+  /// outside the same cm range the metric path uses.
+  static double? parseFeetInchesHeightInCm(int? feet, double? inches) {
```
jordan-lee-code, line 64: **Low / defensive (a few smaller things across the PR, gathered here for lack of a single home, none blocking):** these new parsers let `NaN` past the bounds (a `< min`/`> max` pair both fail on `NaN`), and feet+inches accepts `in >= 12`. Elsewhere: a second profile's onboarding overwrites the device-wide anonymous-data consent flag; `reloadTabBlocs()` omits `TrendsBloc` (masked today by the shell rebuild); missing `mounted` guards after awaits in `profile_editor_screen.dart` and the weight/height date dialogs; the dashboard gauge can produce `NaN` when the goal is 0 (pre-existing); and the intake/activity row ink ripple is painted over by `AppCard`'s opaque fill. Two nice catches worth calling out too: the meal-detail controller leak fix and the `usesImperialFoodUnits` split.

jordan-lee-code (commented):
Simon, this is an enormous and genuinely lovely release, and the multi-profile core is careful work. The empty-suffix trick that lets an existing user's data come across untouched is lovely, and I checked the scary parts hard: no Hive `typeId` collisions, the generated adapters match their DBOs, new fields are appended nullable, and the two-box config migration preserves theme/units/locale/goals on upgrade. The migration layer is sound.

A handful of things I'd want to catch before this reaches `main`, left inline and grouped by severity. None are merge gates, just feedback. Happy to pair on any of them.

PR #524, opened 2026-07-14
Codex/apple health readonly

jinjiacheng586-max:


jordan-lee-code:
Thank you for opening this, and apologies for the slow reply.

I do not think this is what you meant to send. As it stands it touches 302 files and removes about 38,000 lines, including the CI workflows, the Android manifest, the launcher icons and the Gemfiles, and it targets `main`, which only ever receives the release merge from `develop`. That pattern usually means a rebase went sideways rather than that any of it was intended.

Closing it on that basis rather than reviewing it. If there is a change in here you did want to make, please do open a fresh branch from `develop` with just that change on it and we will look at it properly. A short description of what you were trying to fix would help us give you a useful answer first time.

Issue #250, opened 2025-06-19
[Feature Request] Ask AI to fill out custom dish information.


devilAPI:
For example, I wanted to eat a nectarine sometime and I had to add it as custom dish. Because I didn't have any nutrition information on my nectarine ofc, I asked ChatGPT about my nectarine, and let GPT estimate the nutrition information.

It would be great to automate this process. That there would be a button with the AI icon, that automatically opens  Maybe with a speech recognition option, so if you cook stuff, you could tell AI what you cooked and it could estimate the nutrition information, and automatically fill it out.

ibart:
I disagree with this idea. This is an open source "cheap project". Chatgpt API costs a lot.

alettsy:
What about using the same food APIs we are already using in the app to generate auto-fill suggestions? 

There could be a slide up panel with similarly named items you can select to fill or options that appear under the "meal name" field so you can select one.

m5x5:
I am in favor of this idea. It is an open source project that is supposed to be affordable, but this feature does not have to hinder that. As long as it is an additional feature that you are free to opt in too, and it does not lock you in to a specific AI provider.

AI will become more affordable and eventually we will have good AI running on peoples mobile devices.
If we had this feature it would very likely save people some time and reduce the friction of entering daily meals.

Features I would be happy to see:
1. **Picture analysis**, when you create a custom dish, and let AI guess macro nutrients, dish name (and maybe ingredients).
2. **Speech recognition**, where one can dictate information for all dishes in a day, or information on a single meal and it automatically enters it.

Maverobot:
I like the idea. For the concern of APIs being expensive, the user can choose to bring their own key or not. If no key, then the feature is disabled.

jordan-lee-code:
Hi @devilAPI — confirming this is still outstanding on `develop`.

Code location: `pubspec.yaml` has no AI-related dependencies (no `openai`, `anthropic`, etc.). Adding AI auto-fill would require: (1) a third-party API client + opt-in privacy disclosure, (2) a dedicated UI on the edit-meal screen, (3) a fallback for offline / no-API-key users.

Labelled `enhancement`. Privacy and cost considerations make this a careful design before implementation.

jordan-lee-code:
Hi @devilAPI, and apologies for sitting on this so long. I spent a day on it yesterday and want to share what I found, because the answer is a no and you deserve more than a quiet close for an issue you opened back in June.

[...]

PR #587, opened 2026-07-25
feat(android): add JSON share intent for AI-assisted meal import

### LINKED ISSUE #123: [Feature Request] Let Users define exact goals

Marc4k:
## Summary
Adds Android share intent support so any app — including AI assistants
using the bundled food-tracker skill — can share meal JSON directly to
OpenNutriTracker. The skill accepts natural-language descriptions and
food photos; it generates the app's existing JSON export format and the
user taps Share → OpenNutriTracker, confirms the import dialog, and the
meals land in their diary.

iOS share intent is out of scope for this PR (no Share Extension target
exists yet). The Flutter service returns null on iOS so there is no
regression — tracked as a follow-up.

## Ai Skill
[food-tracker.skill.zip](https://github.com/user-attachments/files/30372804/food-tracker.skill.zip)

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] CI / tooling
- [x] Localization
- [ ] Other

## Changes
- **Android** — `AndroidManifest.xml` registers `ACTION_SEND` intent-filters
  for `text/plain` and `application/json` so the app appears in the system
  share sheet
- **Android** — `MainActivity.kt` captures the shared text via a
  `MethodChannel` and holds it until Flutter reads it (handles both cold-start
  and singleTop `onNewIntent`)
- **Flutter** — `ShareIntentService` bridges the channel; returns `null` on
  iOS (no-op fallback)
- **Flutter** — `MainScreen` checks for a pending share on cold start and on
  every foreground resume via `AppLifecycleListener`, shows a confirmation
  dialog, and imports approved meals using the existing
  `ImportMealsJsonUsecase`
- **l10n** — two new strings (`shareJsonImportContent`,
  `shareJsonImportErrorLabel`) added to all 9 supported locales (EN + DE, CS,
  IT, PL, SK, TR, UK, ZH)
- **fix** — pre-existing `curly_braces_in_flow_control_structures` lint in
  `meal_relevance_ranker.dart` that blocked CI

## Screenshots / recordings

https://github.com/user-attachments/assets/3f58e1dc-e0ec-4abe-992e-cd1d87f6081f

## Test plan
- [x] Manual check on Android
- [ ] Manual check on iOS (if applicable) — N/A, iOS not yet supported

### Steps
1. Build and install the app (`fvm flutter run --flavor develop`)
2. Use the food-tracker AI skill to generate a meal — it outputs JSON in
   ONT's export format
3. In the AI app, tap Share → OpenNutriTracker
4. Confirm the import dialog — meals appear in today's diary
5. Share invalid/non-JSON text to ONT — error snackbar is shown, no crash

## Checklist
- [x] New interactive widgets include `Semantics(identifier: '...')` where
  needed (`share-json-import-confirm` on the dialog confirm button)
- [x] Localization updated in ARBs **and** `lib/generated/` when user-facing
  strings change
- [ ] Codegen ran if DBOs/DTOs/env changed (`just build`) — no codegen changes

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -45,6 +45,16 @@
                 <action android:name="android.intent.action.MAIN"/>
                 <category android:name="android.intent.category.LAUNCHER"/>
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.SEND"/>
+                <category android:name="android.intent.category.DEFAULT"/>
+                <data android:mimeType="text/plain"/>
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.SEND"/>
+                <category android:name="android.intent.category.DEFAULT"/>
+                <data android:mimeType="application/json"/>
+            </intent-filter>
         </activity>
         <!-- #312: Notification scheduling receivers -->
         <receiver android:exported="false" android:name="com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver"/>
--- a/android/app/src/main/kotlin/com/opennutritracker/ont/opennutritracker/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/opennutritracker/ont/opennutritracker/MainActivity.kt
@@ -1,6 +1,45 @@
 package com.opennutritracker.ont.opennutritracker
 
+import android.content.Intent
 import io.flutter.embedding.android.FlutterActivity
+import io.flutter.embedding.engine.FlutterEngine
+import io.flutter.plugin.common.MethodChannel
 
-class MainActivity: FlutterActivity() {
+class MainActivity : FlutterActivity() {
+    private val channelName = "com.opennutritracker/share_intent"
+
+    // Holds text shared to the app until Flutter consumes it.
+    private var pendingSharedText: String? = null
+
+    override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
+        super.configureFlutterEngine(flutterEngine)
+
+        // Capture any text present in the launch intent (cold start).
+        pendingSharedText = extractSharedText(intent)
+
+        MethodChannel(flutterEngine.dartExecutor.binaryMessenger, channelName)
+            .setMethodCallHandler { call, result ->
+                when (call.method) {
+                    "getSharedText" -> {
+                        result.success(pendingSharedText)
+                        pendingSharedText = null
+                    }
+                    else -> result.notImplemented()
+                }
+            }
+    }
+
+    // Called when the app is already running and receives a new share intent
+    // (singleTop behaviour — Android re-uses the existing activity).
+    override fun onNewIntent(intent: Intent) {
+        super.onNewIntent(intent)
+        pendingSharedText = extractSharedText(intent)
+    }
+
+    private fun extractSharedText(intent: Intent?): String? {
+        if (intent?.action != Intent.ACTION_SEND) return null
+        val type = intent.type ?: return null
+        if (!type.startsWith("text/") && type != "application/json") return null
+        return intent.getStringExtra(Intent.EXTRA_TEXT)
+    }
 }
--- a/lib/core/presentation/main_screen.dart
+++ b/lib/core/presentation/main_screen.dart
@@ -3,12 +3,18 @@ import 'package:opennutritracker/core/domain/usecase/get_config_usecase.dart';
 import 'package:opennutritracker/core/presentation/widgets/add_item_bottom_sheet.dart';
 import 'package:opennutritracker/core/presentation/widgets/demo_mode_banner.dart';
 import 'package:opennutritracker/core/styles/app_palette.dart';
+import 'package:opennutritracker/core/utils/json_meal_importer.dart';
 import 'package:opennutritracker/core/utils/locator.dart';
+import 'package:opennutritracker/core/utils/share_intent_service.dart';
 import 'package:opennutritracker/features/diary/diary_page.dart';
+import 'package:opennutritracker/features/diary/presentation/bloc/calendar_day_bloc.dart';
+import 'package:opennutritracker/features/diary/presentation/bloc/diary_bloc.dart';
 import 'package:opennutritracker/core/presentation/widgets/home_appbar.dart';
 import 'package:opennutritracker/features/home/home_page.dart';
+import 'package:opennutritracker/features/home/presentation/bloc/home_bloc.dart';
 import 'package:opennutritracker/core/presentation/widgets/main_appbar.dart';
 import 'package:opennutritracker/features/profile/profile_page.dart';
+import 'package:opennutritracker/features/settings/domain/usecase/import_meals_json_usecase.dart';
 import 'package:opennutritracker/features/trends/presentation/trends_page.dart';
 import 'package:opennutritracker/generated/l10n.dart';
 
@@ -25,11 +31,21 @@ class _MainScreenState extends State<MainScreen> {
 
   late List<Widget> _bodyPages;
   late List<PreferredSizeWidget> _appbarPages;
+  late final AppLifecycleListener _lifecycleListener;
 
   @override
   void initState() {
     super.initState();
     _loadDemoDataFlag();
+    _lifecycleListener = AppLifecycleListener(onResume: _handleSharedMeal);
+    // Check for a share intent that launched the app cold.
+    WidgetsBinding.instance.addPostFrameCallback((_) => _handleSharedMeal());
+  }
+
+  @override
+  void dispose() {
+    _lifecycleListener.dispose();
+    super.dispose();
   }
 
   // Checked once per screen instance rather than kept live — leaving demo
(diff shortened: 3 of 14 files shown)


Marc4k:
@simonoppowa, any updates on this one? I use it daily and would love to switch back to the main app instead of running my fork. Happy to also implement iOS support if that's a blocker for merging.

simonoppowa (commented):
Thanks for this, @Marc4k — and apologies for the wait. **The delay was not about your PR.** It sat behind the fork-workflow approval gate, so CI had never run on it at all; that was mine to notice and I didn't. I've approved it now and **everything passes**: analyzer, unit tests, intl check, and both platform builds. (One iOS integration run hung at 41 minutes and I re-ran it — flaky simulator, unrelated to your changes; the re-run finished in 16m54s.)

Reading the diff, this is more careful than most: `launchMode="singleTop"` really is already set so `onNewIntent` fires as your comment says, consume-and-clear on the native side genuinely does prevent a double import, `mounted` is checked after every await, the lifecycle listener is disposed, and the new string is translated into all nine locales with no parity drift. That is good work.

---

## The open question is ours, not yours

`JsonMealImporter` assigns `MealSourceEntity.custom`, so macros an assistant estimated land in the diary indistinguishable from a meal the user typed in themselves — and flow from there into TDEE, goal progress and trends.

That is the same question #250 was closed over. The closing comment there explicitly pre-empted the bring-your-own-model route: it doesn't change the underlying claim, which is *a model estimated this* rather than *this came from a database*. Your design sidesteps the cost and privacy objections in that thread completely — there's no key, no network call, no dependency — but not the provenance one.

**That is a project decision we owe you an answer on before this PR can be judged fairly**, and it is being worked through on #599. Nothing in your code causes it and no change you could make here would settle it. I did not want to leave you guessing about what the actual blocker is.

---

## Code review

Ordered by what I'd want addressed, if and when the above is resolved.

**1. `text/plain` puts the app in every text share sheet.** `AndroidManifest.xml` — with a `text/plain` filter, any text share on the device (a browser URL, a selected paragraph) now offers OpenNutriTracker, and choosing it launches to a *"Could not read shared JSON"* snackbar. That footprint reaches every user, including those who never use the skill. `application/json` alone is much narrower; the trade-off is that plenty of apps only share as `text/plain`. Worth a deliberate decision either way.

**2. `onNewIntent` can silently discard a pending share.** `MainActivity.kt`:

```kotlin
override fun onNewIntent(intent: Intent) { pendingSharedText = extractSharedText(intent) }
```

A non-SEND intent arriving before Flutter reads the value — a notification tap, for instance — overwrites it with `null` and the share is lost with no error. Suggest only assigning when the extract is non-null. It also never calls `setIntent(intent)`; harmless today because you use the parameter, but it leaves `getIntent()` stale for anything added later.

**3. The dialog's count can exceed what actually imports.** `main_screen.dart` shows `parsed.intakes.length`, then imports and reports `result.imported`. If the usecase drops entries, the user approved 5 and silently received 3. The JSON is also parsed twice — once to count, once to import.

**4. No tests for the new plumbing.** The importer itself is well covered (`json_meal_importer_test.dart`, `import_meals_json_usecase_test.dart`), but `ShareIntentService` and `_handleSharedMeal` have none. A widget test with a faked `MethodChannel` would cover confirm / cancel / malformed cheaply, and this path writes to the diary.

**5. Unrelated change.** The `meal_relevance_ranker.dart` hunk is a line wrapped in braces, presumably from `dart format`. Harmless, but it is noise in the diff.

---

## One to think about rather than fix

The dialog says "These meals will be added to your diary" with a count, and never shows the food names or the numbers. That is **consistent with the existing QR importer**, so it is not something you introduced — but that path imports another user's database-sourced export, while this imports model estimates, and the same dialog carries a different amount of trust in the two cases.

For contrast, the multi-item review screen being built on #599 states in its own doc comment that its confirmation exists "so nothing reaches the diary that the user has not looked at". A count is not looking at it.

---

## Not reviewed

I have not opened the linked `food-tracker.skill.zip`. It is a meaningful part of what this PR enables, so what I have reviewed is the transport rather than what travels over it.

## Heads-up on overlap

There is a multi-item logging feature in flight on `feature/ai-assisted-meal-logging` (#599) that also touches `meal_relevance_ranker.dart` and the same intake-writing path. Neither branch knows about the other yet. Worth coordinating before either merges so we don't collide.

simonoppowa:
@Marc4k — this went unanswered since your 31 July message, and that's on me. Here's where it actually stands, and a way forward.

## Why it's been held

On 14 August the project settled a rule for AI in this app (#633, and recorded in `docs/ai-model-candidates.md`): **no model may emit a nutrition value.** The AI meal assistance that shipped in 2.2.0 is built on it — a model turns "two eggs and a slice of toast" into food *names and portions*, and every calorie comes from the food databases. The response schema drops any `calories` field a model returns, whichever provider produced it (`lib/features/add_meal/domain/meal_items_api.dart:69`).

This PR ships the other design: the bundled skill has an assistant *generate* kcal, protein, carbs and fat, and the share intent delivers them into the diary. That's the thing the rule exists to keep out, and it's why I couldn't merge it as-is.

## Where that reasoning is weaker than I'd like

I should be straight that the app already has a door this size. `JsonMealImporter` (Settings → Import, PR #395) accepts pasted JSON and *requires* `kcal`, `protein`, `carbs` and `fat`. Nothing stops someone pasting model output there today. So the objection to #587 isn't really the transport — a share intent and a paste box move the same bytes. It's that this PR would make the app **ship and endorse** a workflow where a model produces the numbers, with a skill in the repo to do it, one tap from the diary.

## A version I think I could merge

Drop the bundled skill and reframe the PR as what its code mostly is: **an Android share intent for the app's existing JSON format.** Same contract as the paste importer, one tap instead of copy-paste, no claim about where the JSON came from. That's a genuine convenience for anyone with a JSON source — a script, a spreadsheet export, an assistant they've set up themselves — and it doesn't put the project's name on model-emitted calories.

If you're up for that:

- it's `CONFLICTING` against `develop` now, so a rebase first
- the import confirmation dialog should stay, so the user sees the numbers before they land
- iOS as the follow-up you already offered

If instead the point of the PR *is* the skill, I understand — but then it's the same decision as #633, and I'd rather say so plainly than leave you waiting another month.

ibrahim-iqbal (commented):
Read the whole flow end-to-end. The plumbing is clean and the guarantees around double-import are correct; a couple of things worth surfacing before it lands.

**Manifest + MainActivity glue**

- `SEND` intent-filter split into two distinct `data` elements (one for `text/plain`, one for `application/json`) is the right shape — Android's intent-filter matcher requires each `mimeType` in its own `data` node for the OR-of-types semantics to work.
- `configureFlutterEngine` reads `intent` for the cold-start case, `onNewIntent` handles the warm-start case, and `pendingSharedText` is cleared on `getSharedText` — so a subsequent call after a share returns null. That's the double-import guard.
- `extractSharedText` filters on `text/*` or `application/json`. Any other mime types the intent-filter might match (`application/*+json`, `text/x-json`) fall through as no-ops. Fine — the intent-filter itself won't match those anyway.

**Two real things worth fixing before merge:**

1. **`AppLifecycleListener` is never disposed** — this is a leak. In `initState`:

   ```dart
   _lifecycleListener = AppLifecycleListener(onResume: _handleSharedMeal);
   ```

   ...but the state has no `dispose()` override that calls `_lifecycleListener.dispose()`. `AppLifecycleListener` registers itself with `WidgetsBinding`, so without `dispose()` it holds a strong reference to `_handleSharedMeal` (which captures `this`) for as long as `WidgetsBinding` lives. On `MainScreen` specifically the leak is bounded (it's the app's root route), but it will trip lint on strict analyzers and the pattern is worth being consistent about since it applies to every `AppLifecycleListener` in the codebase.

   ```dart
   @override
   void dispose() {
     _lifecycleListener.dispose();
     super.dispose();
   }
   ```

2. **`_handleSharedMeal` fires on every resume, including resumes that aren't a share** — Kotlin's `pendingSharedText` will be `null` in that case and the whole thing returns early, but it still round-trips the MethodChannel every time the app comes back to the foreground (screen unlock, notification tap, task-switcher return, etc.). Cheap in absolute terms but wasteful in aggregate on a top-level widget's lifecycle callback. `onNewIntent` on the Kotlin side is the natural signal — you could add a second `MethodChannel` invocation from Kotlin *into* Flutter (`invokeMethod('onSharedText', text)`) to push the notification only when there's actually something, and drop the resume-poll entirely. Non-blocking, more of a "next-iteration" note.

**Smaller notes:**

- `JsonMealImporter.parse` returning `.intakes.isEmpty` as the "no meals found" signal works but conflates parse-failed with parsed-but-empty. If the ARB `shareJsonImportErrorLabel` needs to distinguish the two later (e.g. "Not a valid meal JSON" vs "No meals found in that share"), returning a richer result type would let it.
- Three bloc dispatches at the end (`HomeBloc`, `DiaryBloc`, `CalendarDayBloc`) — matches how the CSV importer already refreshes state, so consistent.
- The nine ARB entries look like real translations, not English placeholders. Locale parity is preserved.

Feature-wise this is a nice win — the "AI-assistant → OpenNutriTracker" hop becomes a native share instead of copy-paste-import.

PR #590, opened 2026-07-26
feat(add-meal): suggest a meal type from the current time (#580)

### LINKED ISSUE #580: Suggest a meal type from the current time
### Problem / motivation

Adding food from the global `+` action currently asks users to select Breakfast, Lunch, Dinner, or Snack before they can continue. In most cases the current time provides a useful default, making this an avoidable decision in the common path.

This is a focused extraction from #458.

### Proposed solution

Preselect a suggested meal type based on the device's local time while keeping every meal type immediately selectable.

Acceptance criteria:

- [ ] Define the time ranges for Breakfast, Lunch, Dinner, and Snack in one testable helper
- [ ] Use the device's local time to determine the initial suggestion
- [ ] Visually preselect the suggested meal type when the add-meal flow opens from the global `+` action
- [ ] Let the user choose a different meal type before continuing
- [ ] Do not change flows launched from a specific meal section; those must retain that explicit meal type
- [ ] Handle boundary times deterministically
- [ ] Add unit/widget tests covering each range, boundaries, and explicit meal-section behavior
- [ ] Add or preserve stable `Semantics(identifier: ...)` values for interactive controls

The exact time ranges should be proposed in the PR and kept easy to adjust during review.

### Alternatives considered

Remembering only the last selected meal type. That may be stale across the day and is less useful than a time-aware default.

### Feature area

Meal detail / logging

### Target platforms

Both

### Mockups or references

Parent proposal: #458

### Additional context

This issue covers only the default selection. It does not change search, ranking, serving sizes, or quick-add persistence.

### Checklist

ibrahim-iqbal:
Fixes #580.

## Summary
Preselect a meal type in the global add-item bottom sheet based on the device's local time. Every tile stays immediately tappable; the suggestion renders as a small `Suggested` chip on the matching tile so nothing changes about the tap flow.

## Approach
- New `MealTypeSuggester.suggestFromTime(DateTime)` in `lib/core/utils/` centralizes the time-of-day ranges so they can be tweaked in one place:
  - Breakfast: 04:00 – 10:59
  - Lunch: 11:00 – 14:59
  - Snack: 15:00 – 16:59
  - Dinner: 17:00 – 21:59
  - Snack: 22:00 – 03:59 (late night / pre-dawn)
- `AddItemBottomSheet` accepts a nullable `suggestedType`. The global `+` action in `main_screen.dart` passes `MealTypeSuggester.suggestFromTime(DateTime.now())`. Flows launched from a specific meal section keep passing nothing and see no highlight, per the acceptance criteria.
- The `Suggested` chip uses `colorScheme.secondaryContainer` / `onSecondaryContainer` so it inherits both light and dark themes cleanly.

## Acceptance criteria mapping
- [x] Define time ranges in one testable helper — `MealTypeSuggester`
- [x] Use device local time to determine the initial suggestion — `MealTypeSuggester.suggestFromTime(DateTime.now())` in `main_screen.dart`
- [x] Visually preselect the suggested meal type when opened from global `+` — `Suggested` chip on the matching tile
- [x] Let the user choose a different meal type before continuing — every tile stays tappable (unchanged `onTap` handlers)
- [x] Do not change flows launched from a specific meal section — those callers keep passing nothing, so `suggestedType` stays null and no chip renders
- [x] Handle boundary times deterministically — inclusive-start / exclusive-end ranges; boundary hours (04, 11, 15, 17, 22, 00, 03:59) covered in tests
- [x] Add unit / widget tests covering each range, boundaries, and explicit meal-section behavior — 6 unit tests + 2 widget tests
- [x] Add or preserve stable `Semantics(identifier: ...)` values — chip has `add-item-suggested-chip`; the existing tile identifiers are untouched

## Tests
- `test/core/utils/meal_type_suggester_test.dart` — 6 unit tests covering every range and each boundary hour
- `test/core/presentation/widgets/add_item_bottom_sheet_test.dart` — 2 new widget tests: chip renders on the matching tile, no chip renders when `suggestedType` is null

All 11 tests in the two files pass; `flutter analyze` clean on the touched files.

## Localization
Adds `suggestedLabel: "Suggested"` to `lib/l10n/intl_en.arb`. Other locales fall back to English until they're translated.


jordan-lee-code:
Thank you for this. The shape is right: it suggests without deciding, every tile stays equally tappable, and flows launched from a specific meal section see nothing at all. Pointing the sheet at a likely answer while leaving the choice entirely open is exactly the register we want.

I have pushed two commits rather than sending you back and forth, and repointed the base from `main` to `develop`, since `main` only ever receives the release merge.

- The chip needed all nine locales, not just English. It now reads Vorschlag, Navrženo, Suggerito, Sugerowane, Navrhnuté, Önerilen, Запропоновано and 建议, and the key moved into its alphabetical slot between `sugarLabel` and `suppliedLabel`. Worth knowing for next time: `lib/generated/` is gitignored gen-l10n output, so the ARBs are the whole job and there is nothing to hand-edit underneath.
- The chip announced itself twice. Its `Semantics` wrapper set a label and the `Text` inside said the same word again, so a screen reader landing on it heard "Suggested, Suggested". `excludeSemantics: true` fixes it, and there is a test that fails with exactly that doubled string if it ever comes back out.

Your `add-item-suggested-chip` identifier is untouched, which the on-device driver script navigates by, so thank you for adding it in the first place.

Analyzer clean and the full suite green at 991.

jordan-lee-code (approved):

PR #615, opened 2026-08-03
fix(profile): stop stone weights from showing a full stone of pounds

munzzyy:
## Summary

With the body weight unit set to stones, some weights render with 14 lb in the pounds slot, which is a whole stone. 63.49 kg shows as "9 st 14 lbs" when it is 10 st.

`UnitCalc.kgToStLb` already guards against this, but at two decimals: it rounds the remainder to 2dp and rolls over when that hits 14.00. Every screen then prints the value at one decimal, and a remainder like 13.97 clears the 2dp guard and rounds to 14 on the way to the label. The existing test in `body_weight_format_test.dart` names this ("formatProfileWeight may round 13.99 up to 14.0 for display") and leaves it as a separate concern, so I picked it up.

Repro on `develop`, calling the real `formatBodyWeight`:

```
31.74 kg -> "4 st 14 lbs"
44.44 kg -> "6 st 14 lbs"
63.49 kg -> "9 st 14 lbs"
76.19 kg -> "11 st 14 lbs"
88.89 kg -> "13 st 14 lbs"
```

Same call with this branch:

```
31.74 kg -> "5 st 0 lbs"
44.44 kg -> "7 st 0 lbs"
63.49 kg -> "10 st 0 lbs"
76.19 kg -> "12 st 0 lbs"
88.89 kg -> "14 st 0 lbs"
```

Sweeping 30 to 250 kg in 10 g steps, 62 values hit it before, 0 after.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] CI / tooling
- [ ] Localization
- [ ] Other

## Related issues

None open that I could find, and no PR touching the st/lb path.

## Changes

- `UnitCalc.kgToStLb` takes an optional `poundsDecimals`, defaulting to the existing 2. The remainder is rounded to that precision and the roll-over guard is checked against the rounded value, so the invariant holds at whatever precision the caller is going to print.
- The three places that display the split pass `poundsDecimals: 1`: `formatBodyWeight` (profile page, home weight chip, weight history), the st/lb seed in `BodyWeightInput`, and the target-weight delta on the profile page. All three print one decimal today.
- `stLbToKg` and `ValueValidator.parseStLbWeightInKg` are untouched. They go through the default, so stored values and the kg round-trip are unchanged.

I kept this as a parameter on the existing function rather than a second helper because the roll-over rule and the rounding it depends on belong together. Splitting them is what let the two drift apart in the first place.

## Screenshots / recordings

Text-only change, no layout difference. The before/after strings above are the user-visible part.

## Test plan
- [ ] Manual check on Android

I do not have an Android or iOS device set up here, so this is unit-tested only. The three call sites are pure formatting off the same helper, and the test sweep covers the range the app accepts, but flagging that I did not put it on a handset.

### Steps
1. `flutter test test/unit_test/body_weight_format_test.dart test/unit_test/unit_calc_imperial_round_trip_test.dart` - 46 passing.
2. `flutter test` - 966 passing, up from 962 on `develop`.
3. `flutter analyze` - "No issues found!".
4. Mutation check: reverted the `lib/` half and re-ran, which fails "a remainder that rounds to 14 lb is shown as the next stone" and "no weight in the supported range prints 14 lb".

New tests:

- `body_weight_format_test.dart`: 63.49 kg formats as "10 st 0 lbs", plus a sweep of 30 to 250 kg asserting the pounds component never reads as 14.
- `unit_calc_imperial_round_trip_test.dart`: `kgToStLb(63.49)` is still `(9, 13.97)` at the default, and `(10, 0.0)` with `poundsDecimals: 1`. Also a plain case so the new precision is not just testing the boundary.

I also dropped the now-stale note in the existing "raw pounds value never reaches 14.0" test that said the display rounding was out of scope.

## Checklist
- [x] New interactive widgets include `Semantics(identifier: '...')` where needed
- [x] Localization updated in ARBs **and** `lib/generated/` when user-facing strings change
- [x] Codegen ran if DBOs/DTOs/env changed (`just build`)

No new strings, no widgets, no DBO or DTO changes, so the localization and codegen boxes are nothing-to-do rather than done. On formatting: I matched the surrounding style by hand instead of running `just format`, since `dart format` on 3.44.6 rewrites 260 files here and the workflow comment says the format pass is deliberately parked for its own PR.

### DIFF
--- a/lib/core/utils/calc/unit_calc.dart
+++ b/lib/core/utils/calc/unit_calc.dart
@@ -73,10 +73,20 @@ class UnitCalc {
   /// the UK-style "11 st 5.2 lb" display and input. Pounds keep two decimals,
   /// mirroring [lbsToKg], and a value that rounds up to a full 14 lb rolls
   /// into the next stone so we never show "10 st 14.0 lb".
-  static (int stones, double pounds) kgToStLb(double kg) {
+  ///
+  /// [poundsDecimals] is the precision the pounds figure is rounded to, and
+  /// the roll-over is checked against that rounded value. It matters because
+  /// every screen prints one decimal while the default here keeps two: a
+  /// 13.99 lb remainder clears the guard, then renders as "14", which is a
+  /// whole stone shown as a remainder ("9 st 14 lb" for a weight that is
+  /// 10 st). Callers that display the result pass 1; the two-decimal default
+  /// is what [stLbToKg] round-trips against.
+  static (int stones, double pounds) kgToStLb(double kg,
+      {int poundsDecimals = 2}) {
     final totalLbs = kg * 2.20462;
     var stones = totalLbs ~/ 14;
-    var pounds = double.parse((totalLbs - stones * 14).toStringAsFixed(2));
+    var pounds =
+        double.parse((totalLbs - stones * 14).toStringAsFixed(poundsDecimals));
     if (pounds >= 14.0) {
       stones += 1;
       pounds = 0.0;
--- a/lib/features/profile/presentation/utils/profile_display_format.dart
+++ b/lib/features/profile/presentation/utils/profile_display_format.dart
@@ -45,7 +45,9 @@ String formatBodyWeight(
     case BodyWeightUnit.lb:
       return '${formatProfileWeight(UnitCalc.kgToLbs(weightKg))} $lbLabel';
     case BodyWeightUnit.st:
-      final (stones, pounds) = UnitCalc.kgToStLb(weightKg);
+      // One decimal, because that is what formatProfileWeight prints: at the
+      // two-decimal default a 13.99 lb remainder renders as a full stone.
+      final (stones, pounds) = UnitCalc.kgToStLb(weightKg, poundsDecimals: 1);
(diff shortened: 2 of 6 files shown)


jordan-lee-code:
Thank you for picking up the thread that the existing test explicitly left dangling. Seeing your own weight rendered as "9 st 14 lbs" when it is plainly 10 st is the kind of small wrongness that quietly erodes trust in everything else on the screen, so this is worth having.

Threading the display precision through `kgToStLb` rather than adding a second helper is the right call, and your reasoning for it is the reason: splitting the rollover rule from the rounding it depends on is what let them drift apart in the first place. The 30 to 250 kg sweep is convincing, and leaving `stLbToKg` and the parser on the two-decimal default keeps the stored values untouched.

One thing worth naming for the record rather than for you to fix: opening the weight field and saving it without editing still shifts the stored value by about 13 g, and that is unchanged from `develop` rather than anything this introduces.

CI green. Recommending merge.

jordan-lee-code (approved):

Issue #599, opened 2026-08-01
[Feature]: AI-assisted meal logging


simonoppowa:
Umbrella issue for adding AI-assisted meal logging. This carries the decision record so the individual sub-issues can stay short, and so the reasoning survives even if the later tiers are never built.

## The problem

Logging food that is already in the database is cheap — 5–6 taps, or 3 from the Recent list. Logging a *meal* means repeating that loop per item: a breakfast of eggs, toast and coffee is three full search-and-confirm cycles, roughly 18 taps and 30 keystrokes.

Separately and more painfully, food that **isn't** in the database costs 5 taps + ~15 keystrokes via Quick Add (which discards macros) or **11 taps + ~21 keystrokes** via the full custom-meal editor.

## Constraints this was designed against

Any AI feature here has to survive three things the app sells itself on:

1. **Privacy as a falsifiable claim.** The README does not say "we respect privacy", it says *"What leaves your device — these four destinations, nothing else"*, backed by a published signing fingerprint. That sentence is the product.
2. **Local-first by architecture.** No account, AES-256 Hive boxes, key in Keystore/Keychain.
3. **"Every number is cited."** Calorie targets follow IOM 2005, BMI follows WHO, macros follow WHO TRS 916, activity follows the 2024 Compendium — each linked from the in-app Sources & References screen.

## Shape: three tiers

```
tier 0  multi-item review UI + deterministic parser      → every user, no network
tier 1  + settings, key storage, Claude client, text     → users with an API key
tier 2  + camera entry point, base64 WebP in the request → users with an API key
```

Only **tier 0** is approved and scoped. It is tracked by the sub-issues on this one.

Tier 0 stands alone cleanly: the parser resolves everything against the existing food database, so every number it produces still comes from Open Food Facts / USDA / BLS. It generates no estimates, which defers the entire provenance apparatus the cloud tiers need — a new `MealSourceEntity` value, an export-schema change, and an "estimated" marker on every intake surface across 9 locales. **Nothing about the privacy table or the citation claim changes in tier 0.**

## Decisions

| # | Question | Decision |
|---|---|---|
| 1 | What job is AI hired for? | Cut logging friction. Insight/interpretation, food-DB matching, and competitive parity all deferred. |
| 2 | Where may inference run? | Local by default + opt-in bring-your-own API key. The project never sees user data and pays nothing. |
[...]

Issue #602, opened 2026-08-01
Multi-item meal review screen, entry point and strings


simonoppowa:
Part of the tier-0 work — see the parent issue for the full decision record. Depends on the parser and resolver sub-issues. **This is the one that actually ships the feature.**

## The screen

**New:** `lib/features/add_meal/presentation/screens/bulk_add_screen.dart`, plus `bulk_add_bloc.dart` / `_event.dart` / `_state.dart`. flutter_bloc, registered `registerFactory` in the locator — per-navigation, matching `AddMeal` / `Products` / `Food`.

- Multi-line text field, meal-type selector, date — then Parse.
- One row per resolved item showing the matched food name (tap to open a candidate picker), quantity and unit (prefilled from the parse, editable), and the resulting kcal.
- **Unresolved rows are visibly flagged**, with *skip* and *Quick Add instead* actions. They must not be silently dropped.
- **Log all** → for each item, call `MealDetailBloc.addIntake(context, unit, amountText, type, meal, day)`.

That last point matters: `addIntake` already handles `TrackedDay` bootstrap and macro accumulation, and it is already driven headlessly — `import_meal_scanner_screen.dart` calls it with no detail screen mounted, and so does the home drag-to-move in `intake_vertical_list.dart`. **Do not reach for `AddIntakeUsecase` directly**; that skips the TrackedDay bookkeeping and you would have to re-implement it.

Afterwards, refresh the Home and Diary blocs and `popUntil(mainRoute)`, exactly as `MealDetailBottomSheet.onAddButtonPressed` already does.

Lift the unit-dropdown construction and the `^\d+([.,]\d{0,2})?$` quantity validator from `lib/features/meal_detail/presentation/widgets/meal_detail_bottom_sheet.dart` rather than re-deriving them — including the `<= 10000` upper bound.

## Two things `addIntake` does not do for you

> **Added 2026-08-03.** Both are consequences of calling `addIntake` in a loop rather than through the bottom sheet, and neither is visible from `addIntake`'s signature.

[...]

Issue #601, opened 2026-08-01
Resolve parsed meal items against the existing food search


simonoppowa:
Part of the tier-0 work — see the parent issue #599 for the full decision record. Depends on #600 landing first.

> **Scope widened 2026-08-02.** Following the revisions in [this comment on #599](https://github.com/simonoppowa/OpenNutriTracker/issues/599#issuecomment-5160029398), this is no longer tier-0 plumbing. It is the **shared resolver both tiers feed**, and the database-first policy lands here rather than in the tier-1 client. Build it with the second producer in mind.

Take the parsed items and resolve each one to real food-database entries, reusing the existing search stack rather than building anything new.

**New file:** `lib/features/add_meal/domain/usecase/resolve_parsed_meals_usecase.dart`, registered in `lib/core/utils/locator.dart` as `registerFactory` (per-navigation, matching how `AddMeal` / `Products` / `Food` blocs are registered).

## Two producers, one contract

Both tiers converge on the same shape, so the resolver should accept the union rather than being written for the parser alone:

| Producer | Emits | Fallback macros |
|---|---|---|
| `parseMealText` (tier 0, #600) | `(query, quantity, unit)` | none — never |
| Model client (tier 1, not yet scoped) | `(name, quantity, unit)` | present, but consulted **only** on a database miss |

**The database always wins when it has a match.** A fallback macro payload is used only for items the search cannot resolve, and only those rows get the `estimated` provenance. Tier 0 supplies no fallback at all, so tier 0 produces no estimated rows and nothing about the citation claim changes.

## What it does

For each item, call both existing entry points on `SearchProductsUseCase`:

- `searchOFFProductsByString(query)`
- `searchFDCFoodByString(query)` — note this queries Supabase despite the name

Then merge with the existing `mergeAndRankMeals(off, sp, query)` from `lib/features/add_meal/util/meal_relevance_ranker.dart`, which already handles cross-source dedup, keeps the user's own custom meals and recipes in a tier above remote results, and relevance-sorts within each tier.

Keep the top handful as candidates so the review UI can offer a correction when the first match is wrong.

```dart
class ResolvedMealItem {
  final ParsedMealItem parsed;
  final List<MealEntity> candidates;  // ranked, may be empty
  final int selectedIndex;
}
```

[...]

PR #616, opened 2026-08-04
feat(add-meal): deterministic meal-text parser

### LINKED ISSUE #600: Deterministic meal-text parser
Part of the tier-0 work — see the parent issue for the full decision record.

Add a pure, dependency-free parser that turns one line of free text into a list of search intents. It performs no I/O and touches no UI, so it can land and be fully tested on its own.

**New file:** `lib/features/add_meal/util/meal_text_parser.dart`, beside the existing `meal_relevance_ranker.dart`.

Model it directly on `lib/core/utils/json_meal_importer.dart` — a pure static entry point, no I/O, and per-entry indexed error strings rather than silent drops.

```dart
class ParsedMealItem {
  final String query;       // "toast"  → goes to the existing food search
  final double? quantity;   // 100      → null when unstated
  final String? unit;       // "g"      → null when unstated
}

class MealTextParseResult {
  final List<ParsedMealItem> items;
  final List<String> errors;   // "Item 2: quantity must be greater than 0"
}

MealTextParseResult parseMealText(String input);
```

## Behaviour

- **Segment** the input on `,`, `;`, newline, and `+` — subject to the comma rule below.
- **Extract** a leading or trailing quantity+unit per segment — `100g toast`, `toast 100g`, `2 eggs`.
- **No quantity stated** → leave both `quantity` and `unit` null. The review row downstream then falls back to the defaults `MealDetailScreen._applyInitialSelection` already uses (serving if `meal.hasServingValues`, else 100 metric / 1 imperial), so this parser must not invent its own.
- **Validate** against the bounds the app already enforces elsewhere: quantity `> 0` and `<= 10000` (see `meal_detail_bottom_sheet.dart`). Reuse `FoodNameValidator.isValid` to reject segments containing no unicode letter.
- **Empty segments are skipped, not errors** — a trailing comma should not produce a complaint.

## The comma is both a separator and a decimal point — precedence rule

> **Corrected 2026-08-03.** An earlier version of this issue said to segment on `,` *and* accept `,` as a decimal separator, with no rule saying which wins. Those two instructions contradict each other and the test list below asked for both. `1,5 l milk` cannot satisfy both readings.

**The rule: a comma with a digit immediately before it and a digit immediately after it is a decimal point. Every other comma segments.**

- `1,5 l milk` → one item, quantity `1.5` — digits on both sides
- `100g toast, 2 eggs` → two items — space follows the comma
- `toast,2 eggs` → two items — no digit before the comma

This is a documented choice, not a discovered truth. It makes `100,2 eggs` parse as `100.2 eggs` rather than as two items, which is the one genuinely ambiguous case; seven of the nine shipped locales use the comma decimal, so resolving toward the decimal reading is the right default. The period decimal (`1.5 l milk`) is unaffected and always parses as a decimal.

## Normalise units — the app has exactly six, and `kg`/`l` are not among them

> **Added 2026-08-03.** `UnitDropdownItem` (`lib/features/meal_detail/presentation/bloc/meal_detail_bloc.dart:248`) defines exactly six units: `g`, `ml`, `g/ml`, `oz`, `fl.oz`, `serving`. `fromString` ends in `default: return UnitDropdownItem.gml`, so an unrecognised unit is silently coerced to `g/ml` rather than rejected. An earlier version of this issue told the parser to emit `kg` and `l`, which meant its own headline test case `1,5 l milk` would have logged **1.5 g/ml instead of 1500 ml — a silent 1000× error with no warning anywhere.**

Accept `kg` and `l` as *input*, then convert before emitting:

| Parsed | Emit |
|---|---|
| `kg` | `g`, quantity × 1000 |
| `l` | `ml`, quantity × 1000 |
| `g`, `ml`, `oz` | unchanged |

Apply the `<= 10000` bound **after** conversion, so `15 kg` is rejected rather than becoming `15000 g`.

**`unit` must always be either null or one of the six `UnitDropdownItem.toString()` values.** Assert this in a test — it is the invariant that stops a future unit addition reintroducing the same silent coercion.

## Keep it locale-independent

Key off digits and unit **symbols** (`g`, `kg`, `ml`, `l`, `oz`), not number-words and not `and` / `und` / `e`. That keeps the parser working across all 9 supported locales with no per-locale word lists — the food *names* go to the existing search, which already handles translations via `SPConst.translationLocaleOf`.

Please don't add English word-number parsing ("two eggs"). It works in exactly one of nine locales and sets a precedent that costs 9 files every time it grows.

## Tests

**New file:** `test/unit_test/meal_text_parser_test.dart`. Mirror the structure of the existing `test/unit_test/json_meal_importer_test.dart`.

- [ ] Leading quantity (`100g toast`) and trailing quantity (`toast 100g`)
- [ ] Period decimal (`1.5 l milk`) → quantity `1500`, unit `ml`
- [ ] Comma decimal (`1,5 l milk`) → **one** item, quantity `1500`, unit `ml`
- [ ] Comma as separator (`100g toast, 2 eggs`) → **two** items
- [ ] `toast,2 eggs` → two items (no digit before the comma)
- [ ] `kg` → `g` ×1000, and `l` → `ml` ×1000
- [ ] **Invariant: every emitted `unit` is null or one of the six `UnitDropdownItem` values**
- [ ] `15 kg` is rejected — the `<= 10000` bound applies after conversion
- [ ] Quantity with no unit (`2 eggs`)
- [ ] No quantity at all (`black coffee`) → both fields null
- [ ] Each separator: `,` `;` newline `+`
- [ ] Rejects `0` and anything `> 10000`
- [ ] Empty string and whitespace-only input
- [ ] A segment with no letters (`123`) is rejected via `FoodNameValidator`
- [ ] A single item with no separator at all

```sh
flutter test test/unit_test/meal_text_parser_test.dart
```

## Notes for a first-time contributor

This is self-contained: one new file plus one new test file, no UI, no state management, no code generation, and no new dependency. Run `just test` before opening the PR. The repo uses a 120-character line width and conventional commits (`feat:`, `fix:`, …) — see `AGENTS.md`.

Note that `just ci` currently cannot pass locally — it includes a `dart format --set-exit-if-changed` check against pre-existing formatting drift in committed files. Run `just test` and `just analyze` individually instead.

laurenlliu:
Closes #600.

Adds `parseMealText()` — a pure, dependency-free parser that turns one line of free text into a list of search intents for the tier-0 slice of #599. No I/O, no UI, no new dependency.

## What it does

- Segments on `,` / `;` / newline / `+`, with the comma decimal-vs-separator rule from the issue (a comma with a digit on both sides is a decimal point; every other comma is a separator).
- Extracts a leading or trailing quantity+unit per segment (`100g toast`, `toast 100g`, `1,5 l milk`).
- Converts `kg`→`g` and `l`→`ml` (×1000) before emitting, so the value that reaches `UnitDropdownItem.fromString` is never silently coerced to `g/ml` the way raw `kg`/`l` would be.
- Rejects segments with no letters via the existing `FoodNameValidator`, and enforces the same `>0 && <=10000` quantity bound `meal_detail_bottom_sheet.dart` already uses (checked *after* unit conversion).
- Asserts the unit invariant — every emitted unit is `null` or one of `UnitDropdownItem`'s six values — with a dedicated test sweeping a batch of inputs.

## Testing

- 30 new unit tests in `test/unit_test/meal_text_parser_test.dart`, covering every case in the issue's checklist plus a couple of edge cases I found while implementing (unrecognized unit tokens, negative numbers).
- Full suite: `flutter test` → 992/992 passing (no regressions).
- `flutter analyze` → no issues.
- `dart format --set-exit-if-changed` → clean.

## Notes for review

- Per the issue, I deliberately did **not** add English number-word parsing ("two eggs") — kept it to digits + `g`/`kg`/`ml`/`l`/`oz` symbols for locale independence across all 9 supported locales.
- Happy to adjust the exact error-message wording in `errors` if you'd prefer different phrasing than `Item N: ...`.

### DIFF
--- a/lib/features/add_meal/util/meal_text_parser.dart
+++ b/lib/features/add_meal/util/meal_text_parser.dart
@@ -137,6 +137,22 @@ String? _normalizeUnitSymbol(String raw) {
   }
 }
 
+/// The app's [UnitDropdownItem] (`meal_detail_bloc.dart`) has no `kg` or
+/// `l` — its `fromString` falls through to `g/ml` for anything it doesn't
+/// recognize, so emitting `kg`/`l` unchanged would silently turn `1,5 l
+/// milk` into 1.5 g/ml instead of 1500 ml. Converting here, before the
+/// value ever reaches that code, is what avoids the silent 1000x error.
+(double, String) _normalizeUnitAndQuantity(double quantity, String unit) {
+  switch (unit) {
+    case 'kg':
+      return (quantity * 1000, 'g');
+    case 'l':
+      return (quantity * 1000, 'ml');
+    default:
+      return (quantity, unit);
+  }
+}
+
 /// Turns one line of free text into a list of search intents. Pure and
 /// dependency-free: no I/O, no UI. See the class docs on [ParsedMealItem]
 /// and [MealTextParseResult] for the shape, and the doc comments on the
@@ -149,11 +165,18 @@ MealTextParseResult parseMealText(String input) {
 
   for (var i = 0; i < segments.length; i++) {
     final extracted = _extractQuantityAndUnit(segments[i]);
+
+    var quantity = extracted.quantity;
+    var unit = extracted.unit;
+    if (quantity != null && unit != null) {
+      (quantity, unit) = _normalizeUnitAndQuantity(quantity, unit);
+    }
+
     items.add(
       ParsedMealItem(
         query: extracted.query.trim(),
-        quantity: extracted.quantity,
-        unit: extracted.unit,
+        quantity: quantity,
+        unit: unit,
       ),
     );
   }
--- a/lib/features/add_meal/util/meal_text_parser.dart
+++ b/lib/features/add_meal/util/meal_text_parser.dart
@@ -73,7 +73,14 @@ const _unitSymbol = r'(?:kg|ml|oz|g|l)';
 
 /// A number, accepting either decimal separator. `JsonMealImporter` takes
 /// both the same way; [_parseQuantity] does the conversion.
-const _number = r'\d+(?:[.,]\d+)?';
+///
+/// The leading `-` is matched deliberately even though a negative quantity
+/// is never valid. Refusing it here does not reject the input — it just
+/// stops the number being recognized as a quantity at all, so `-5g sugar`
+/// would fall through to the food search as a literal query and the user
+/// would silently get the default amount instead of the bounds error that
+/// `0g water` produces. Matching it lets the `> 0` check below do its job.
+const _number = r'-?\d+(?:[.,]\d+)?';
 
 /// A quantity+unit token immediately before or after the food name, e.g.
 /// `100g toast` or `toast 100g`. `\s*` (not `\s+`) between the number and
--- a/lib/features/add_meal/util/meal_text_parser.dart
+++ b/lib/features/add_meal/util/meal_text_parser.dart
@@ -69,6 +69,74 @@ bool _isDigit(String char) {
   return code >= 0x30 && code <= 0x39;
 }
 
+/// A quantity+unit token immediately before or after the food name, e.g.
+/// `100g toast` or `toast 100g`. `\s*` (not `\s+`) between the number and
+/// the unit letters is what lets `1.5 l milk` — number, space, unit — match
+/// the same way as the no-space `100g toast` does.
+final _leadingQuantity = RegExp(r'^(\d+(?:\.\d+)?)\s*([a-zA-Z]*)\s+(.+)$');
+final _trailingQuantity = RegExp(r'^(.+?)\s+(\d+(?:\.\d+)?)\s*([a-zA-Z]*)\s*$');
+
+/// The result of attempting to pull a quantity+unit out of one segment.
+/// [unit] is the raw matched letters (not yet normalized/validated) or
+/// `null` when no unit token was present.
+class _Extracted {
+  final String query;
+  final double? quantity;
+  final String? unit;
+
+  const _Extracted({required this.query, this.quantity, this.unit});
+}
+
+/// Tries [_leadingQuantity] then [_trailingQuantity]. A unit token that
+/// doesn't match a recognized symbol (see [_normalizeUnitSymbol]) is
+/// treated as part of the food name instead of a unit, since a run of
+/// letters glued to a number is more often a product code than a unit
+/// this parser doesn't know — leaving it in the query keeps the segment
+/// searchable rather than silently discarding it.
+_Extracted _extractQuantityAndUnit(String segment) {
+  final leading = _leadingQuantity.firstMatch(segment);
+  if (leading != null) {
+    final rawUnit = leading.group(2)!;
+    if (rawUnit.isEmpty || _normalizeUnitSymbol(rawUnit) != null) {
+      return _Extracted(
+        query: leading.group(3)!,
+        quantity: double.parse(leading.group(1)!),
+        unit: rawUnit.isEmpty ? null : rawUnit.toLowerCase(),
+      );
+    }
+  }
+
+  final trailing = _trailingQuantity.firstMatch(segment);
+  if (trailing != null) {
+    final rawUnit = trailing.group(3)!;
+    if (rawUnit.isEmpty || _normalizeUnitSymbol(rawUnit) != null) {
+      return _Extracted(
+        query: trailing.group(1)!,
+        quantity: double.parse(trailing.group(2)!),
+        unit: rawUnit.isEmpty ? null : rawUnit.toLowerCase(),
+      );
+    }
+  }
+
+  return _Extracted(query: segment);
+}
+
+/// `null` when [raw] (case-insensitive) isn't one of the five symbols this
+/// parser recognizes as input. Deliberately just `g`/`kg`/`ml`/`l`/`oz` —
+/// see the file doc comment for why word-based units are out of scope.
+String? _normalizeUnitSymbol(String raw) {
+  switch (raw.toLowerCase()) {
+    case 'g':
+    case 'kg':
+    case 'ml':
+    case 'l':
+    case 'oz':
+      return raw.toLowerCase();
+    default:
+      return null;
+  }
+}
+
 /// Turns one line of free text into a list of search intents. Pure and
 /// dependency-free: no I/O, no UI. See the class docs on [ParsedMealItem]
 /// and [MealTextParseResult] for the shape, and the doc comments on the
@@ -80,7 +148,14 @@ MealTextParseResult parseMealText(String input) {
   final errors = <String>[];
 
   for (var i = 0; i < segments.length; i++) {
-    items.add(ParsedMealItem(query: segments[i]));
+    final extracted = _extractQuantityAndUnit(segments[i]);
+    items.add(
+      ParsedMealItem(
+        query: extracted.query.trim(),
+        quantity: extracted.quantity,
+        unit: extracted.unit,
+      ),
+    );
   }
 
   return MealTextParseResult(items: items, errors: errors);
--- a/lib/features/add_meal/util/meal_text_parser.dart
+++ b/lib/features/add_meal/util/meal_text_parser.dart
@@ -4,6 +4,11 @@ import 'package:opennutritracker/core/utils/food_name_validator.dart';
 /// meal_detail_bottom_sheet.dart.
 const _maxQuantity = 10000;
 
+/// The exact set of `UnitDropdownItem.toString()` values (meal_detail_bloc
+/// .dart). [ParsedMealItem.unit] must always be `null` or one of these —
+/// see the assert in [parseMealText] and the dedicated invariant test.
+const _validUnits = {'g', 'ml', 'g/ml', 'oz', 'fl.oz', 'serving'};
+
 /// One item extracted from a free-text meal description. [query] is what
 /// gets handed to the existing food search; [quantity] and [unit] are
 /// `null` when the input didn't state an amount, in which case the review
@@ -202,6 +207,13 @@ MealTextParseResult parseMealText(String input) {
         continue;
       }
     }
+
+    // This is the invariant that stops a future unit addition reintroducing
+    // the silent kg/l coercion _normalizeUnitAndQuantity exists to prevent.
+    assert(
+      unit == null || _validUnits.contains(unit),
+      'parseMealText emitted unrecognized unit "$unit"',
+    );
 
     items.add(ParsedMealItem(query: query, quantity: quantity, unit: unit));
   }
--- a/lib/features/add_meal/util/meal_text_parser.dart
+++ b/lib/features/add_meal/util/meal_text_parser.dart
@@ -1,0 +1,87 @@
+/// One item extracted from a free-text meal description. [query] is what
+/// gets handed to the existing food search; [quantity] and [unit] are
+/// `null` when the input didn't state an amount, in which case the review
+/// row downstream falls back to the defaults `MealDetailScreen
+/// ._applyInitialSelection` already uses.
+class ParsedMealItem {
+  final String query;
+  final double? quantity;
+  final String? unit;
+
+  const ParsedMealItem({required this.query, this.quantity, this.unit});
+
+  @override
+  String toString() =>
+      'ParsedMealItem(query: $query, quantity: $quantity, unit: $unit)';
+}
+
+/// Result of [parseMealText]. [errors] holds a one-line, item-indexed
+/// reason for each segment that could not be turned into a [ParsedMealItem]
+/// — empty segments (e.g. a trailing comma) are skipped silently and never
+/// produce an error.
+class MealTextParseResult {
+  final List<ParsedMealItem> items;
+  final List<String> errors;
+
+  const MealTextParseResult({required this.items, required this.errors});
+
+  bool get hasErrors => errors.isNotEmpty;
+}
+
+/// Splits [input] on `,` / `;` / newline / `+`, subject to the comma rule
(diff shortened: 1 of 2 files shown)


lib/features/add_meal/util/meal_text_parser.dart
```diff
@@ -0,0 +1,250 @@
...
+  if (char.isEmpty) return false;
+  final code = char.codeUnitAt(0);
+  return code >= 0x30 && code <= 0x39;
+}
+
+/// A quantity+unit token immediately before or after the food name, e.g.
+/// `100g toast` or `toast 100g`. `\s*` (not `\s+`) between the number and
+/// the unit letters is what lets `1.5 l milk` — number, space, unit — match
+/// the same way as the no-space `100g toast` does.
+final _leadingQuantity = RegExp(r'^(\d+(?:\.\d+)?)\s*([a-zA-Z]*)\s+(.+)$');
```
simonoppowa: **Blocking — multi-word foods silently lose their quantity.**

```
"2 eggs"                    ->  query="eggs"                     qty=2.0
"2 chicken breasts"         ->  query="2 chicken breasts"        qty=null
"3 slices bread"            ->  query="3 slices bread"           qty=null
"4 chocolate chip cookies"  ->  query="4 chocolate chip cookies" qty=null
```

`([a-zA-Z]*)` greedily takes the first word of the food name as a candidate unit. For `2 chicken breasts` the regex matches on the first attempt with `rawUnit = "chicken"`, and the guard below then discards the whole match because it isn't a unit symbol — without asking the engine for the no-unit parse.

`2 eggs` works only by accident: with a single trailing word there is nothing left for `\s+` to match, so the engine is *forced* to backtrack into the empty-unit reading. Add one more word and that pressure disappears.

Making the group optional **and** restricting it to the known symbols lets the engine find the no-unit reading itself:

```dart
const _unitSymbol = r'(?:kg|ml|oz|g|l)';
final _leadingQuantity = RegExp(
  '^($_number)\\s*($_unitSymbol)?\\s+(.+)\$',
  caseSensitive: false,
);
```

Worth checking the trailing regex on line 88 for the mirror problem rather than assuming it's clean — I believe it is, because its letter run is anchored by `$`, but it's worth a test either way.

This also matters downstream: #601 ranks candidates on token overlap, so leaving the digit glued to the query (`2 chicken breasts`) makes the match measurably worse than dropping it.

lib/features/add_meal/util/meal_text_parser.dart
```diff
@@ -0,0 +1,250 @@
...
+/// treated as part of the food name instead of a unit, since a run of
+/// letters glued to a number is more often a product code than a unit
+/// this parser doesn't know — leaving it in the query keeps the segment
+/// searchable rather than silently discarding it.
+_Extracted _extractQuantityAndUnit(String segment) {
+  final leading = _leadingQuantity.firstMatch(segment);
+  if (leading != null) {
+    final rawUnit = leading.group(2)!;
+    final unit = rawUnit.isEmpty ? null : _normalizeUnitSymbol(rawUnit);
+    if (rawUnit.isEmpty || unit != null) {
```
simonoppowa: This guard is where the match from line 87 gets thrown away. Once the unit group can only ever match a known symbol, the whole `rawUnit.isEmpty || unit != null` dance collapses to a lower-case:

```dart
unit: leading.group(2)?.toLowerCase(),
```

and `_normalizeUnitSymbol` disappears entirely — the regex is already the whitelist. That takes `_extractQuantityAndUnit` from 29 lines to 18.

lib/features/add_meal/util/meal_text_parser.dart
```diff
@@ -0,0 +1,250 @@
...
+
+/// A comma with a digit immediately before it and a digit immediately
+/// after it is a decimal point; every other comma is a separator. This
+/// rewrites decimal commas to `.` in place so the later split on `,` only
+/// ever cuts on separator commas.
+///
+/// `1,5 l milk` → `1.5 l milk` (digits on both sides, one item).
+/// `100g toast, 2 eggs` → unchanged (space follows the comma, two items).
+/// `toast,2 eggs` → unchanged (no digit before the comma, two items).
+String _normalizeDecimalCommas(String input) {
```
simonoppowa: **Should fix — this rewrites food names, not just quantities.**

It runs over the entire input before segmentation, so any digit-comma-digit anywhere is replaced — including inside the name:

```
"yoghurt 3,5% fat"      ->  query="yoghurt 3.5% fat"
"Omega 3,6,9 capsules"  ->  query="Omega 3.6.9 capsules"
```

The string handed to the food search now contains characters the user never typed. `3,5%` is a very common product descriptor in exactly the locales the comma rule exists to serve, so this will fire in normal use.

Rather than rewriting, have the separator regex decline to match a decimal comma in the first place:

```dart
final _separator = RegExp(r'[;+\n]|(?<!\d),|,(?!\d)');
```

"A comma not preceded by a digit, or not followed by one" is the same rule stated as a split condition. The comma-to-period conversion then happens only on the number actually matched, which keeps `1,5 l milk` working while leaving names alone.

lib/features/add_meal/util/meal_text_parser.dart
```diff
@@ -0,0 +1,250 @@
...
+      }
+      if (quantity > _maxQuantity) {
+        errors.add('Item $itemNum: quantity must be $_maxQuantity or less');
+        continue;
+      }
+    }
+
+    // This is the invariant that stops a future unit addition reintroducing
+    // the silent kg/l coercion _normalizeUnitAndQuantity exists to prevent.
+    assert(
```
simonoppowa, line 236: Nit, no action needed if you fix line 87 as suggested: as written this assert can never fire. `_normalizeUnitSymbol` already whitelists to `{g,kg,ml,l,oz}` and normalization maps `kg`→`g` and `l`→`ml`, so the emitted set is always a subset of `{g,ml,oz}`. The comment claims it guards against a future unit addition, but the whitelist gets there first.

With the regex-as-whitelist change it becomes reachable — add a symbol to the pattern without teaching `_normalizeUnitAndQuantity` about it and this fires. So the fix above gives the assert the teeth its comment already claims.

lib/features/add_meal/util/meal_text_parser.dart
```diff
@@ -0,0 +1,250 @@
...
+  // are skipped, not errors" — a trailing comma doesn't consume an index.
+  var itemNum = 0;
+
+  for (final segment in segments) {
+    itemNum++;
+    final extracted = _extractQuantityAndUnit(segment);
+    final query = extracted.query.trim();
+
+    if (!FoodNameValidator.isValid(query)) {
+      errors.add('Item $itemNum: not a valid food name');
```
simonoppowa, line 210: Not a defect and not for this PR — flagging it so it's on the record before #602 builds on it.

These strings are hardcoded English. That matches `JsonMealImporter`'s existing `'Entry $entryNum: …'` pattern, so it's consistent with the codebase and the issue specified this shape. But if they reach the UI in #602 they can't be localized, and the app carries 900 keys across 9 locales at zero drift.

A structured form (index + reason enum, formatted at the presentation layer) would keep that property. Decision for #602 — no change wanted here.

simonoppowa (changes requested):
Thanks for this — the spec is followed closely and the doc comments explain *why* rather than restating the code, which is rarer than it should be. The comma rule, the `kg`/`l` conversion and the bound-after-conversion ordering are all exactly as the issue asked, and resisting word-number parsing was the right call.

I ran the suite (30/30 green) and then drove `parseMealText` directly over a wider set of inputs than the tests cover. Two defects came out of that, both silent — no error, no flagged row, just a wrong search intent handed downstream:

```
"2 chicken breasts"     ->  query="2 chicken breasts"      qty=null
"2 boiled eggs"         ->  query="2 boiled eggs"          qty=null
"yoghurt 3,5% fat"      ->  query="yoghurt 3.5% fat"
"Omega 3,6,9 capsules"  ->  query="Omega 3.6.9 capsules"
```

Details inline. The first is blocking — multi-word foods are most of what people type, and the plan's own worked example is that shape.

One thing that is **not** on you: the file is `dart format`-clean at 80 columns, while `meal_relevance_ranker.dart` and `json_meal_importer.dart` next to it run to 110 and 96. `just format` passes no line-length flag so the tool defaults to 80, but AGENTS.md:156 claims a 120-character width "configured in `analysis_options.yaml`" — there is no `formatter:` block in that file, so that claim isn't true today. You did the tool-correct thing; I need to fix the repo, not this PR.

I have both fixes written and tested against your branch. Happy to push them here so you don't have to redo the work — just say the word, or take the analysis and do it your way, whichever you prefer.

simonoppowa:
Heads up on a base change, nothing to do with the review above: I've retargeted this from `develop` to `feature/ai-assisted-meal-logging`.

Tier 0 lands as three pieces — the parser (this PR, #600), the resolver (#601) and the review screen (#602) — and none of them does anything a user can see until all three are in. Staging them on an integration branch keeps `develop` free of a half-built feature and lets the three be reviewed against each other before the lot merges.

No action needed from you. Your commits are untouched and the diff is unchanged; only the merge target moved. CI will re-run against the new base.

simonoppowa:
I've pushed the two fixes to your branch as `63160d3b` — you'd said in the description you were happy to adjust things, and it seemed friendlier to hand you working code than a list of homework. Take it as a suggestion: revert it and do it your way if you'd rather, no hard feelings.

**What changed**

`_leadingQuantity`'s unit group is now optional *and* restricted to the five known symbols, so the regex engine finds the no-unit reading by itself instead of the code having to notice and retry. That fixes the multi-word case, and it let `_normalizeUnitSymbol` go entirely — the pattern is the whitelist now, so both branches of `_extractQuantityAndUnit` collapse to a lower-case. 29 lines down to 18.

The separator regex now declines to match a decimal comma rather than rewriting the input, and the comma-to-period conversion moved into `_parseQuantity`, where it only touches the number actually matched.

Eight regression tests in two new groups. Each carries a comment describing the failure it guards, so the reason survives the next refactor:

```
"2 chicken breasts"         ->  query="chicken breasts"         qty=2
"4 chocolate chip cookies"  ->  query="chocolate chip cookies"  qty=4
"yoghurt 3,5% fat"          ->  query="yoghurt 3,5% fat"        (unchanged)
"Omega 3,6,9 capsules"      ->  query="Omega 3,6,9 capsules"    (unchanged)
```

**One consequence worth knowing:** the assert I called unreachable in the review now has teeth. Because the regex is the whitelist, adding a symbol to `_unitSymbol` without teaching `_normalizeUnitAndQuantity` about it will fire it — which is what its comment always claimed it did.

**What I did not touch:** the formatting. That's the repo's problem, not this PR's, and I'll fix AGENTS.md separately.

38/38 in `meal_text_parser_test.dart`, and `flutter analyze` clean on both files. I have not run the full suite to completion against a healthy build — worth doing before this merges.

Note the base also moved to `feature/ai-assisted-meal-logging` (see the comment above).

---

*Edited to fix the code block above, which posted with literal `\n` instead of line breaks. A follow-up commit `cce1c9c9` also addresses Copilot's negative-quantity finding — see the review threads.*

lib/features/add_meal/util/meal_text_parser.dart
```diff
@@ -0,0 +1,250 @@
...
+  final code = char.codeUnitAt(0);
+  return code >= 0x30 && code <= 0x39;
+}
+
+/// A quantity+unit token immediately before or after the food name, e.g.
+/// `100g toast` or `toast 100g`. `\s*` (not `\s+`) between the number and
+/// the unit letters is what lets `1.5 l milk` — number, space, unit — match
+/// the same way as the no-space `100g toast` does.
+final _leadingQuantity = RegExp(r'^(\d+(?:\.\d+)?)\s*([a-zA-Z]*)\s+(.+)$');
+final _trailingQuantity = RegExp(r'^(.+?)\s+(\d+(?:\.\d+)?)\s*([a-zA-Z]*)\s*$');
```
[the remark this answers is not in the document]
simonoppowa: Agreed, and fixed in `cce1c9c9`. This was a real gap and I missed it in my own review.

`_number` now matches an optional leading `-`. That looks odd in isolation — a negative quantity is never valid — but refusing it in the pattern did not reject the input, it just stopped the number being seen as a quantity at all, so the segment fell through to the food search as the literal string `-5g sugar` and the review row would have applied the default amount. The user types -5 and silently gets 100. Matching the sign routes it into the existing `> 0` bound, which runs after unit conversion, so `-5kg flour` is caught too.

The comment on the pattern now says why the `-` is there, since it reads like a bug otherwise.

test/unit_test/meal_text_parser_test.dart
```diff
@@ -0,0 +1,306 @@
...
+      // describes unsigned digits), so the whole segment falls through to
+      // the query as-is rather than being parsed as quantity -5. It still
+      // passes FoodNameValidator (it contains letters), so this is
+      // accepted as an odd-looking but valid item, not an error.
+      final item = parseMealText('-5g sugar').items.single;
+
+      expect(item.query, '-5g sugar');
+      expect(item.quantity, isNull);
+      expect(item.unit, isNull);
+    });
```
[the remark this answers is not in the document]
simonoppowa: Right on both counts — the test asserted the wrong behaviour, and `> 0 && <= 10000` is what the issue specifies. Replaced in `cce1c9c9`: `-5g sugar` now expects `Item 1: quantity must be greater than 0`, the same error `0g water` produces.

Four more tests pin the boundaries, because the sign change touches parsing rather than just validation:

- `sugar -5g` — the trailing form is rejected the same way
- `-5kg flour` — rejected after the x1000 conversion, not before
- `-123` — still fails as an invalid food name, for having no letters rather than for its sign
- `low-fat milk, 100g Coca-Cola` — hyphenated names are untouched, since a `-` only reads as a sign when glued to a quantity's digits

PR #624, opened 2026-08-09
fix(l10n): make the Ukrainian and Chinese bulk-add hints parseable

### LINKED ISSUE #623: Bulk add: the Ukrainian and Chinese placeholder hints demonstrate unparseable syntax
Found while running the bulk-add screen through a wide set of inputs on a Pixel 6 and headlessly. Introduced by me in #619 — this is not a pre-existing problem.

## What is wrong

The placeholder in the bulk-add text field is the only place the input syntax is explained. Two of the nine shipped hints demonstrate a format the parser cannot read:

| Locale | Hint | `parseMealText` result |
|---|---|---|
| `uk` | `2 яйця, 100г тост, чорна кава` | 3 items, but the middle is `query="100г тост"` with **no quantity** |
| `zh` | `2个鸡蛋、100克吐司、黑咖啡` | **1 item**, the whole line unparsed |

The Ukrainian `г` is Cyrillic **ge** (U+0433), not Latin `g` — visually identical, functionally not a unit. The Chinese hint uses `克` for gram and `、` (ideographic comma) as separator; neither is a recognised unit symbol or separator.

The other seven parse into exactly the three items they depict.

A user who follows the example gets a row with digits stuck in the food name, sent to the food search as a literal query. It looks like the feature is broken, and the app told them to type it that way.

## Fix

#624 changes the two strings and adds a test that reads every ARB off disk and asserts each hint parses into three items, with at least two amounts recognised and no digits stranded in a query. It fails on both old strings with the reason spelled out, so this cannot regress.

Ukrainian just swaps to a Latin `g`. Chinese drops the count and shows two weights instead — see below for why.

## The larger point, which is not fixed

**The parser's design does not fit languages that do not put spaces between words.**

`parseMealText` finds a quantity by looking for a number adjacent to whitespace. Chinese writes `2个鸡蛋` — number, counter, noun, no spaces — so there is nothing to split on. Every workaround is bad:

- `2个鸡蛋` → one token, no quantity
- `2 个鸡蛋` → quantity 2, query `个鸡蛋`, the counter now polluting the food name
- `鸡蛋 2` → parses correctly but is not how anyone writes it

So for `zh` the feature currently supports weights with Latin unit symbols and nothing else. That is a real capability gap, not a typo, and it deserves a decision rather than a quiet workaround. Japanese and Korean would have the same problem if they were ever added.

A cheaper adjacent question: **should `、` and `，` (fullwidth comma) be separators?** They are punctuation rather than vocabulary, so adding them would not breach the no-per-locale-word-lists rule from #600 — and a CJK keyboard produces `，` by default, so a Chinese user typing naturally today gets one row no matter what the placeholder says. That change is small and probably worth making on its own.

## How it was missed

Every unit test passed. The hints were only ever read as translations, never as input — nothing connected the string to the parser that has to read it. The new test closes that specific gap.

simonoppowa:
Fixes #623.

The placeholder in the bulk-add field is the only place the input syntax is explained, so a hint the parser cannot read teaches a format that does not work. Two of the nine did:

| Locale | Was | Result |
|---|---|---|
| `uk` | `2 яйця, 100г тост, чорна кава` | middle item kept its digits: `query="100г тост"`, no quantity |
| `zh` | `2个鸡蛋、100克吐司、黑咖啡` | one unparsed item |

The Ukrainian `г` is Cyrillic ge (U+0433), not Latin `g` — visually identical, functionally not a unit. The Chinese hint uses `克` for gram and `、` as separator, neither recognised.

## The change

Ukrainian swaps to a Latin `g`; the food names are untouched.

Chinese becomes `100g 吐司, 250ml 牛奶, 黑咖啡` — two weights, no count. **A count cannot be demonstrated in Chinese**: the counter sits between number and noun with no spaces, so `2个鸡蛋` is one token and `2 个鸡蛋` merely moves `个` into the food name. Showing an unnatural word order to demonstrate a feature is worse than demonstrating the part that works.

## The test

Reads every ARB off disk and asserts each hint parses into the three items it depicts, with at least two amounts recognised and **no digits stranded in a query** — digits left in a food name are the signature of an amount that was not understood.

It fails on both old strings with the reason spelled out:

```
intl_zh.arb hint "2个鸡蛋、100克吐司、黑咖啡" parsed into 1 items, but it depicts three
```

Reading the ARBs directly rather than going through the generated localizations means it covers all nine locales without a widget test, and it will cover any locale added later for free.

## Verification

`flutter test` → 1055/1055. `flutter analyze` clean. ARB parity holds at 932 keys across all nine.

## Also here: CJK list separators

`、` `，` `；` are now separators, and the Chinese hint uses `，` rather than an ASCII comma.

Fixing the hint alone would have told a Chinese user to reach for a punctuation mark their keyboard does not produce — a CJK keyboard emits the fullwidth forms by default, so the feature was broken for them regardless of what the placeholder said.

This does not reintroduce per-locale word lists: the set is punctuation, fixed, tiny, and does not grow when a language is added. None of them carries a decimal meaning, so unlike `,` they need no lookaround.

Two things fell out for free:

- **U+3000, the ideographic space, is already matched by `\s`** — `100g　吐司` resolves with no further work.
- The original broken hint now yields **three separate rows** the user can fix individually rather than one unparsable row, even though the counts in it still cannot be read.

Tests cover each separator, the ideographic space, CJK and ASCII mixed in one line, and that a number on both sides of a fullwidth comma still splits. Japanese and Korean patterns were verified too, though neither locale ships yet.

## Still not addressed

The parser locates a quantity by looking for a number adjacent to whitespace, so **languages that do not space their words still cannot express counts**. `2个鸡蛋` remains unreadable — the counter sits between number and noun with no separator to key off. For `zh` this feature now supports weights with Latin unit symbols, in either comma style. That gap needs a decision rather than a workaround and stays recorded in #623.

### DIFF
--- a/lib/features/add_meal/util/meal_text_parser.dart
+++ b/lib/features/add_meal/util/meal_text_parser.dart
@@ -54,10 +54,22 @@ class MealTextParseResult {
 /// `Omega 3,6,9 capsules` reached the food search with characters the user
 /// never typed. The decimal comma is now converted only inside the number
 /// actually parsed, in [_parseQuantity].
-final _separator = RegExp(r'[;+\n]|(?<!\d),|,(?!\d)');
+///
+/// The CJK separators — `、` (ideographic comma), `，` and `；` (fullwidth
+/// comma and semicolon) — are here because a Chinese, Japanese or Korean
+/// keyboard emits them by default. Without them a user typing a list the
+/// only way their keyboard offers gets a single unparsed row no matter what
+/// the placeholder demonstrates.
+///
+/// They are punctuation, not vocabulary, so this does not reintroduce the
+/// per-locale word lists this parser was designed to avoid: the set is
+/// fixed, tiny, and does not grow when a language is added. None of them
+/// carries a decimal meaning, so unlike `,` they need no lookaround.
+final _separator = RegExp(r'[;+\n、，；]|(?<!\d),|,(?!\d)');
 
 /// Splits [input] into trimmed, non-empty pieces on `,` / `;` / newline /
-/// `+`, subject to the comma rule on [_separator].
+/// `+` and their CJK equivalents, subject to the comma rule on
+/// [_separator].
 List<String> _segment(String input) => input
     .split(_separator)
     .map((s) => s.trim())
--- a/lib/l10n/intl_uk.arb
+++ b/lib/l10n/intl_uk.arb
@@ -1105,7 +1105,7 @@
   "scannerLockOrientationTooltip": "Зафіксувати вертикально",
   "scannerUnlockOrientationTooltip": "Дозволити обертання",
   "bulkAddTitle": "Додати кілька позицій",
-  "bulkAddInputHint": "2 яйця, 100г тост, чорна кава",
+  "bulkAddInputHint": "2 яйця, 100g тост, чорна кава",
   "bulkAddParseLabel": "Знайти продукти",
   "bulkAddSubmitLabel": "Записати {count} позицій",
   "bulkAddNoMatchLabel": "Не вдалося знайти відповідник",
--- a/lib/l10n/intl_zh.arb
+++ b/lib/l10n/intl_zh.arb
@@ -1106,7 +1106,7 @@
   "scannerLockOrientationTooltip": "锁定为竖屏",
   "scannerUnlockOrientationTooltip": "允许旋转",
   "bulkAddTitle": "批量添加",
-  "bulkAddInputHint": "2个鸡蛋、100克吐司、黑咖啡",
+  "bulkAddInputHint": "100g 吐司, 250ml 牛奶, 黑咖啡",
   "bulkAddParseLabel": "查找食物",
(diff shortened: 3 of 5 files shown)


simonoppowa:
Confirmed on device: with the app switched to Chinese, the placeholder renders as `100g 吐司，250ml 牛奶，黑咖啡` (verified by @simonoppowa on a Pixel 6 — I could not reach the in-app language picker through UI automation).

That closes the loop on this change. It is now verified at every level:

- the ARB holds the corrected string
- `gen-l10n` emits it intact, fullwidth comma preserved
- the new test parses every locale's hint into the three items it depicts, and fails on both original strings
- the parser handles `，` `、` `；` and the ideographic space
- it renders correctly in the Chinese UI
- 11 German cases on-device show no regression from the separator change

The one thing still unverified is CJK text as *input* on a device — `adb shell input text` is ASCII-only, so the separator behaviour is covered headlessly rather than by typing 吐司 into the field on hardware.

Issue #622, opened 2026-08-09
Bulk add: a counted food with no unit is logged as grams ("2 eggs" → 2 g)


simonoppowa:
Follow-up from #602. Found while driving the bulk-add screen on a Pixel 6, not by any test — every unit test passes and the arithmetic is correct.

## What happens

Typing `100g Toast, 2 Eier, 1,5 l Milch` into the bulk-add screen produces:

| Row | Amount | kcal |
|---|---|---|
| Toast · K Classic | 100 g | 250 |
| Eier · Lidl | **2 g/ml** | **3** |
| Milch · K Classic | 1500 g/ml | 705 |

`2 Eier` is logged as **two grams of egg**. The diary entry reads `Eier — 0.07 oz — 3 kcal`.

Toast and Milch are right, because the user stated a unit. The egg row is the case where they stated a *count*.

## Why

`parseMealText` correctly emits `quantity: 2, unit: null` — it cannot know whether a bare number is a count, a weight or a serving, and #600 deliberately keeps it from guessing.

`BulkAddBloc._initialUnit` then fills the gap:

```dart
if (meal != null && meal.hasServingValues) {
  return meal.servingUnit ?? MealDetailUnits.gml;
}
return usesImperialUnits ? MealDetailUnits.oz : MealDetailUnits.gml;
```

So a food *with* serving data gets a sensible per-serving default. `Eier` from the Lidl OFF record has none, so it falls through to `g/ml` and the count becomes a weight.

This mirrors what `MealDetailScreen._applyInitialSelection` already does for a single food, and was deliberate — a bulk row and a hand-added row of the same food should start from the same number. The single-item screen gets away with it because the user is looking at one unit dropdown. On a bulk screen with several rows, a wrong unit on one row is easy to scroll past.

**Nothing is silently corrupted** — the amount and unit are both visible and editable on the row, and the kcal figure updates live, so the 3 kcal is there to be noticed. This is a defaults-and-visibility problem, not a data-integrity one.

## Why it matters more than it looks

"2 eggs" is the worked example in the feature's own plan and in the German placeholder text this screen ships with (`2 Eier, 100g Toast, schwarzer Kaffee`). Counting foods — eggs, slices, bananas, biscuits — are exactly what people type into a free-text box. A row that reads `3 kcal` is obviously wrong to a human, but a row reading `2 g` of something denser is not.

## Options

Recording these rather than picking one; this is a design decision.

[...]

PR #625, opened 2026-08-09
feat(add-meal): add deterministic meal text parser

GhagSagar23:
## Summary

Adds the approved Tier-0 offline parser for multi-item meal text. It splits supported item separators while preserving decimal commas, extracts leading or trailing quantities, normalizes `kg` to `g` and `l` to `ml`, and rejects invalid or oversized quantities.

Focused parser tests cover segmentation, decimal commas, quantity placement, unit conversion, bounds, malformed input, and supported-unit invariants.

## Scope

This is limited to the #600 parser milestone. Database resolution (#601), review and atomic logging UI (#602), and all cloud/AI tiers remain deferred.

## Verification

- `git diff --check` passed for both added files.
- The focused Flutter test and `flutter analyze` were not run: the required Docker runner could not start because the Colima Docker socket was unavailable.

Part of #599; implements its approved #600 Tier-0 milestone.

### DIFF
--- a/lib/features/add_meal/util/meal_text_parser.dart
+++ b/lib/features/add_meal/util/meal_text_parser.dart
@@ -1,0 +1,135 @@
+import 'package:opennutritracker/core/utils/food_name_validator.dart';
+
+const _maxQuantity = 10000;
+const _validUnits = {'g', 'ml', 'g/ml', 'oz', 'fl.oz', 'serving'};
+const _unitSymbol = r'(?:kg|ml|oz|g|l)';
+const _number = r'-?\d+(?:[.,]\d+)?';
+
+/// An item parsed from the Tier-0 multi-item entry field.
+class ParsedMealItem {
+  final String query;
+  final double? quantity;
+  final String? unit;
+
+  const ParsedMealItem({required this.query, this.quantity, this.unit});
+}
+
+/// Parsed items plus validation errors for input segments that cannot be
+/// safely passed to the existing food search.
+class MealTextParseResult {
+  final List<ParsedMealItem> items;
+  final List<String> errors;
+
+  const MealTextParseResult({required this.items, required this.errors});
+
+  bool get hasErrors => errors.isNotEmpty;
+}
+
+/// Commas surrounded by digits are decimal marks; every other comma, plus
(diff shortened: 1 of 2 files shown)


simonoppowa:
Thank you for this, and I'm sorry — **this was already implemented and merged three days ago**, and you had no way of knowing.

That's entirely our fault. #600 was still open, unassigned, and still labelled `help wanted` and `good first issue` when you picked it up. It stayed open because the work merged into a staging branch (`feature/ai-assisted-meal-logging`) rather than the default branch, so GitHub never auto-closed it. You did exactly what the issue invited you to do. I've now removed those labels and marked all three tier-0 issues as implemented so nobody else loses an evening the same way.

I read your code properly rather than just closing this. It follows the spec closely and it is good work — the separator lookarounds, the `kg`/`l` normalisation before the bounds check, the unit alternation that lets the regex find the no-unit reading, the indexed error strings. Those are the details that matter in this parser and you got them right.

**And you did one thing the merged version does not.** Your `_quantityAndUnitOnly` branch handles a segment that is only a quantity and unit:

```dart
final _quantityAndUnitOnly = RegExp('^($_number)\\s*($_unitSymbol)\\s*$', caseSensitive: false);
```

On `main` today, `100g` on its own parses as the food name `"100g"` with no quantity and is sent to the food search as a literal query. Yours returns an empty query, which then fails `FoodNameValidator` and is reported to the user instead. That is the better behaviour — a stray `100g` is a mistake worth telling someone about, not a search term. I confirmed the current gap by driving the app on a device.

**If you would like to open that as a small PR against the current parser, I would review it happily** — it is a genuine improvement, it is small, and it comes with a clear test case. `lib/features/add_meal/util/meal_text_parser.dart` on the `feature/ai-assisted-meal-logging` branch is where it now lives.

If you would rather work on something with more in it, [#622](https://github.com/simonoppowa/OpenNutriTracker/issues/622) is genuinely open and unclaimed: typing `2 eggs` currently logs two *grams* of egg, because a stated quantity with no unit falls back to `g/ml`. The issue lays out four options with the trade-offs, so it is a real design question rather than a chore. It is self-contained and reproducible in a unit test.

Sorry again for the wasted effort. Thanks for taking the time on it.

PR #638, opened 2026-08-13
fix(trends): respect configured day boundary

### LINKED ISSUE #628: [Bug]: Trends Page Not correctly sorting data by time
### Description

On the android version of the app, when the day starts at time is set to something other than midnight, the data on the trends page is not correctly sorted. Data from before midnight is assigned to the prior day, and data entered after midnight is sorted into the next day, rather than being split by the day starts at time.

### Steps to reproduce

1. Install the latest release for OpenNutriTracker on Android from Github
2. Set the day starts at time to something other than midnight (in my case 8 am)
3. log food and/or water consumption before midnight
4. log food and/or water consumption after midnight

### Expected behaviour

The calories and water consumption amount logged after midnight, but before the day starts at time, should be assigned to the previous day in the trends page.

### Actual behaviour

The calories and water consumption logged before midnight are assigned to the correct day, but any calories and water consumption logged after midnight, and before the 'day starts at' time are assigned to the next day.

### Platform

Android

### OS version

Android 17

### App version

2.0.2

### Install source

GitHub Release

### Feature area

Other

### Screenshots / recordings

Image
Image
Image
Image
Image

### Logs or stack trace

```markdown

```

### Additional context

_No response_

### Checklist

Charlie284:
## Summary

Trends now anchor their date windows to the configured logical day instead of wall-clock midnight. Food totals, water buckets, chart windows, and streak calculations share the existing `DayBoundaryCalc` boundary and use calendar-date arithmetic across DST transitions.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] CI / tooling
- [ ] Localization
- [ ] Other

## Related issues

Fixes #628

## Changes

- Use the configured day-start offset when loading the current and prior trends windows.
- Bucket water entries into the same logical days used by food and diary data.
- Keep chart and streak date ranges aligned with logical today and local calendar dates across DST.
- Cover midnight and 08:00 boundaries, entries around midnight and the boundary, both aggregation paths, and a spring DST transition.

## Screenshots / recordings

Not applicable; there are no visual changes.

## Test plan
- [ ] Manual check on Android

### Steps
1. Ran the focused day-boundary, trends, and refresh tests: passed, 52 tests.
2. Ran `flutter analyze`: passed with no issues.
3. Ran `just test`: passed, 966 tests.
4. Ran `just format`: completed. The pinned formatter also changes 259 files on an untouched `develop` checkout, so that unrelated repository-wide churn is not included here.
5. Ran `just ci`: stopped at its `format --set-exit-if-changed` gate for the same 259-file baseline mismatch; codegen, tests, and analysis were not reached by that recipe. The separate test and analysis commands above passed.

## Checklist
- [x] No new interactive widgets were added
- [x] No user-facing strings changed
- [x] No DBOs, DTOs, or environment fields changed

### DIFF
--- a/lib/core/utils/calc/day_boundary_calc.dart
+++ b/lib/core/utils/calc/day_boundary_calc.dart
@@ -135,8 +135,12 @@ class DayBoundaryCalc {
   }
 
   static DateTime _logicalDayOfTotalMinutes(DateTime moment, int totalMinutes) {
-    final shifted = moment.subtract(Duration(minutes: totalMinutes));
-    return DateTime(shifted.year, shifted.month, shifted.day);
+    final minutesSinceMidnight = moment.hour * 60 + moment.minute;
+    final dayDelta = minutesSinceMidnight < totalMinutes ? -1 : 0;
+    // The setting is a local wall-clock boundary, not an elapsed duration.
+    // Stepping the calendar date avoids shifting an extra hour when a local
+    // timezone enters or leaves daylight saving time.
+    return DateTime(moment.year, moment.month, moment.day + dayDelta);
   }
 
   /// A bare midnight is how this app spells "the calendar day named
--- a/lib/features/trends/presentation/bloc/trends_bloc.dart
+++ b/lib/features/trends/presentation/bloc/trends_bloc.dart
@@ -8,6 +8,7 @@ import 'package:opennutritracker/core/domain/usecase/get_tracked_day_usecase.dar
 import 'package:opennutritracker/core/domain/usecase/get_user_usecase.dart';
 import 'package:opennutritracker/core/domain/usecase/get_water_intake_usecase.dart';
 import 'package:opennutritracker/core/domain/usecase/get_weight_log_usecase.dart';
+import 'package:opennutritracker/core/utils/calc/day_boundary_calc.dart';
 
 part 'trends_event.dart';
 part 'trends_state.dart';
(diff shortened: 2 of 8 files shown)


ibrahim-iqbal (commented):
Read the whole thing end-to-end. This is the right fix for #628 and the DST-avoidance is well-executed.

**`DayBoundaryCalc._logicalDayOfTotalMinutes`** — the old `moment.subtract(Duration(minutes: totalMinutes))` was the actual bug. `Duration` is elapsed wall time, so on a DST spring-forward Sunday it added a spurious hour to the day boundary and the whole trends window shifted by a day for anyone whose configured offset happened to straddle the shift. The rewrite

```dart
final minutesSinceMidnight = moment.hour * 60 + moment.minute;
final dayDelta = minutesSinceMidnight < totalMinutes ? -1 : 0;
return DateTime(moment.year, moment.month, moment.day + dayDelta);
```

reads the local wall-clock components directly and lets `DateTime`'s own constructor handle the calendar rollover (negative days wrap into the previous month correctly on Dart), so DST cannot pull the boundary off. The comment saying it's "a local wall-clock boundary, not an elapsed duration" is exactly the invariant future readers need.

**`TrendsBloc.LoadTrendsEvent`** — three consistent moves:

- Fetching `config` before computing `today` lets `DayBoundaryCalc.currentLogicalDayMinutes(dayStartOffset)` anchor the trends window to the same boundary the diary and food totals already use. Previously the trends card silently disagreed with everything else on a day-offset user.
- `endOfToday` derived from `today.year/month/day` (rather than `now.year/month/day`) inherits the same anchoring — no gap where the range end straddles midnight relative to today.
- `today.subtract(Duration(days: ...))` → `DateTime(today.year, today.month, today.day - N)` for range starts and prior-week windows. Same DST reason as `_logicalDayOfTotalMinutes` — `Duration(days:)` is 24h × N in elapsed time, not N calendar days.

**Water bucketing** — the new `DayBoundaryCalc.logicalDayOfMinutes(e.dateTime, offset)` puts water into the same buckets as food and diary. Before, water was bucketed on the raw calendar day; a 01:30 glass of water for a user with a 03:00 offset previously counted as "yesterday" for food but "today" for water, so the water card and the food card disagreed on the same drink. Fixed.

**Chart-window UTC arithmetic** — the `DateTime.utc(...).difference(...).inDays` trick to compute calendar-day count between two local dates is exactly the workaround Dart requires here. `today.difference(earliest!).inDays` on local `DateTime` will return 6 (not 7) around a DST spring-forward day because one of the days is only 23h. Using UTC-labelled dates for the diff makes it exact-days. Non-obvious defensive move, worth the two-line change.

**Tests** — `trends_bloc_test` adds 170 lines covering midnight and 08:00 boundaries, entries around midnight, and the water-bucket alignment. `trends_calc_test` adds 18 lines that lock in the calendar-date behaviour. Between them they'd catch a regression on the old `Duration`-based approach on any device where the test runs a DST-relevant clock.

Small non-blocking note: the water loop passes `e.dateTime` to `logicalDayOfMinutes` — worth verifying `e.dateTime` is always the local zone. If any older entries were persisted in UTC, they'd bucket wrong. Not something this PR introduces, but a future audit target.

LGTM.

PR #648, opened 2026-08-14
feat: AI-assisted meal logging — multi-item entry, offline by default, optional model and camera (#599)

simonoppowa:
Brings the AI-assisted meal logging feature ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)) to `develop` as one reviewable change, staged on `feature/ai-assisted-meal-logging` since 9 August so it could be reviewed whole rather than in fragments.

**85 commits, 111 files, +31,662/−519.** Level with `develop`: #971 resolved the conflicts this branch had accumulated and #972 recorded the ancestry a squash had discarded, so `git log develop..feature` reads 0.

## What a user gets

An appbar action on the add-meal screen opens a multi-item screen. Type `100g Toast, 2 Eier, schwarzer Kaffee`, confirm one screen, and three intakes are logged — instead of three full search-and-confirm cycles. **This works with no key, no account and no network beyond the food search the app already does.**

If the user supplies their **own** API key, the same box also accepts free text — `zwei Eier und drei Scheiben Brot`, `I had a chicken caesar salad and a latte`, `2个鸡蛋，200ml牛奶` — and a camera action appears that reads a meal from a photo.

## The claim this feature does not break

**Every calorie and macro still comes from Open Food Facts / USDA / BLS.** Nothing here estimates nutrition.

That holds on every path, and it is enforced rather than documented:

- The deterministic parser produces only a *search query* plus an amount. It has nowhere to put a calorie.
- The model's tool schema has **no macro fields**, so it cannot return one — asserted by a test that fails if anyone adds one.
- Whatever a model returns passes `validateParsedMealItems`, the same bounds the regex enforces. A model cannot write to the diary under looser rules than a parser.

So no `MealSourceEntity` value was added, no export-schema change, no "estimated" marker. The provenance apparatus agreed for a macro-emitting tier stays deferred, and **[#250](https://github.com/simonoppowa/OpenNutriTracker/issues/250) does not need reopening for this.**

A device pass found what that guarantee does *not* cover, and it is worth stating plainly: `Spiegeleier` resolved to **Haribo Spiegeleier — the gummy sweets**. Real numbers from a real Open Food Facts record, no model-emitted macro, every promise kept, and the entry still wrong. That is a food-matching problem, not a provenance one ([#847](https://github.com/simonoppowa/OpenNutriTracker/issues/847), fixed).

## Tier 0 — offline, no key, no network

| | |
| :-- | :-- |
| `meal_text_parser.dart` | segments a line, extracts amounts, locale-independent |
| `resolver_relevance.dart` | ranking for resolution; the existing ranker scored `eggs` against `Egg` at 0.0 |
| `resolve_parsed_meals_usecase.dart` | resolves each item against the existing search, per-source failure isolation |
| `bulk_add_screen.dart` + bloc | the review surface; every row editable, nothing logs unconfirmed |
| `meal_quantity_converter.dart` | shared with meal detail so the two write paths cannot drift |

The parser is deliberately **locale-independent** — it keys off digits and unit symbols, never number-words or `and`/`und`/`e`, so it behaves identically in all nine locales without per-language word lists. Word-numbers are the model's job precisely because that is what does not generalise.

## Four destinations, and the user picks

What began as one hardcoded provider is now a choice behind one interface:

| Destination | Key | Models |
| :-- | :-- | :-- |
| Anthropic | user's own | one curated |
| OpenAI | user's own | two curated |
| OpenRouter | user's own | four curated, vendor named per row |
| **A server the user runs** | optional | free text, or fetched from the endpoint |

The fourth is the one that changes the feature's character: an OpenAI-compatible endpoint on the user's own hardware means the text never leaves their network. It brings its own problems, each answered:

- **Plaintext is refused to anything that is not private** ([#758](https://github.com/simonoppowa/OpenNutriTracker/issues/758)). `http://` to a public host is refused *before anything is sent*, and the notice says so. `http://` to a private address is allowed and warned about.
- **A capability probe** ([#784](https://github.com/simonoppowa/OpenNutriTracker/issues/784)) asks the endpoint what it can actually do — meal text, photos — and remembers the answer, because a small local model that reads text but not images is the normal case.
- **A cold model gets time to load** ([#776](https://github.com/simonoppowa/OpenNutriTracker/issues/776)): 120 s, six times what a hosted API gets.
- **JPEG rather than WebP** ([#747](https://github.com/simonoppowa/OpenNutriTracker/issues/747)), per destination, because llama.cpp cannot decode WebP at all.

## Consent, and what leaves the device

The key lives in the platform keystore, is write-only in the UI, and can be paused without being deleted. **With no key configured, nothing changes anywhere** — no camera action, no network destination, no behaviour difference.

Before anything is stored, the user reads a disclosure naming their chosen destination and must agree to it ([#835](https://github.com/simonoppowa/OpenNutriTracker/issues/835)). Withdrawing consent removes the key. The photo path repeats the disclosure at the moment the camera would open — the last point at which declining still costs nothing.

Only the typed line (or the photo) and the app language are sent. Never the diary, profile or history. The project never sees the key and does not pay for its use. The README's destination table moves from three to four.

**On the typed path**, every failure — no network, rejected key, rate limit, changed API, an outright bug — falls back to the deterministic parser. Failures that will not fix themselves say so; a blip stays silent, because a notice that fires for transient failures becomes one people learn to ignore.

**A photo has nothing underneath it.** There is no offline way to turn a picture into food names, so `ReadMealPhotoUseCase` carries the opposite policy on purpose: a failure there is a dead end and is reported rather than absorbed. An earlier version of this paragraph said "every failure" without that qualification, which read as a guarantee the photo path does not make.

## Verification

- **1,849 tests pass** on this head — measured, not tallied from the branches that fed it. 31 new test files against `develop`.
- **ARB keys**: **1022 in every one of the nine locales** — exact parity, and this branch adds **+86 to each**. (An earlier revision of this description said the counts differed per locale; that was an artifact of a line-based count that also matched keys nested inside `@`-metadata blocks. Counted as parsed JSON, they have never differed.)
- **Live API probes** across nine locales against every provider: zero disagreements with the deterministic parser, zero invariant violations.
- **Two device acceptance passes on a Pixel 6** ([#830](https://github.com/simonoppowa/OpenNutriTracker/issues/830)), the second against this branch head with all four providers live.

Device testing keeps finding what unit tests cannot. The first pass found a thousandfold litre under-count, a unit substitution bypassing a row warning, and a correct empty model answer being overridden. The second found seven more: the Haribo match above, a Cancel button that did not cancel a provider switch, a doubled comma in four locales, a failure notice landing at the wrong end of a long Settings screen, a setup check that denied having run after four minutes of running, a quoted wait that understated its worst case by half, and a refusal notice outliving the address it judged. **All eight are fixed and merged** ([#847](https://github.com/simonoppowa/OpenNutriTracker/issues/847), [#848](https://github.com/simonoppowa/OpenNutriTracker/issues/848), [#849](https://github.com/simonoppowa/OpenNutriTracker/issues/849), [#850](https://github.com/simonoppowa/OpenNutriTracker/issues/850), [#851](https://github.com/simonoppowa/OpenNutriTracker/issues/851), [#852](https://github.com/simonoppowa/OpenNutriTracker/issues/852), [#853](https://github.com/simonoppowa/OpenNutriTracker/issues/853), [#854](https://github.com/simonoppowa/OpenNutriTracker/issues/854)).

## Not yet verified, and honestly

- **[#830](https://github.com/simonoppowa/OpenNutriTracker/issues/830) cases 15–18 are open.** The three photo cases need a person and real food at the lens; the rejected-key case needs a deliberately bad key in a provider slot, which overwrites a real one.
- **The capability probe's passing path has never been observed on hardware.** It has only ever been seen timing out. [#854](https://github.com/simonoppowa/OpenNutriTracker/issues/854) closed the automated gap — every failure kind is now pinned, and a verdict is proven to survive a restart — but cases 21–23 still need a real OpenAI-compatible server on the tester's network.
- **Non-English strings are machine translations.** The doubled comma in cs/de/sk/uk was the visible symptom of that; a native read of the own-server disclosures is still worth doing.

## Before merging

- **[#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599) will not auto-close.** `Closes` only fires on the default branch.
- **The `feature/**` CI trigger** in `default_workflow.yml` was added so PRs into this branch got checks at all. Keep it if the branch lives on; drop it if this is the end of the line.
- Version bump and changelog for the new screen.
- ~~This branch is three commits behind `develop`.~~ Level with it now (#971, #972).

## Not included, deliberately

A model emitting macros is not here and cannot be until **decision #5** is settled against [#250](https://github.com/simonoppowa/OpenNutriTracker/issues/250) — settled 2026-08-14 as *a model may never emit nutrition values*. [#587](https://github.com/simonoppowa/OpenNutriTracker/issues/587) raises the same question through a share intent and is held on the same answer.


simonoppowa:
## Device pass on this head — one defect, filed as #973

Ran the typed model path on a Pixel 6 (`oriole`, `de-DE`) against `20e9a9b4` — full flavour, debug, `2.1.0+62` — installed over an existing configuration so the stored keys and consent were the ones a real user would have. Anthropic was the active provider.

Input `zwei Eier und drei Scheiben Brot`: digit-free on purpose, so the locale-independent parser cannot quantify it and only the model can answer. It did — *"Von KI gelesen — prüfe jede Zeile"* — and read both items and both counts correctly.

The counts then lost their unit:

| Row | Match | Amount | Energy |
| :-- | :-- | :-- | :-- |
| Eier | Classic, K-Classic, Kaufland | **2 g/ml** | 3 kcal |
| Brot | Backerkronung | **3 g/ml** | 8 kcal |

Both rows warned `Einheit für diese Menge prüfen`; **Save all** logged them anyway. Breakfast reads **11 / 657 kcal** for two eggs and three slices of bread. Filed as **#973** (`backlog:high`).

**The provenance claim held.** Every number came from Open Food Facts and Supabase — `portions_by_food_ids`, `portion_labels_by_food_ids` in the log. No model-emitted macro anywhere. This is #847's shape one field over: real numbers from a real record, every promise in the description kept, and the entry still wrong. It lands on the portion work already in this build (#967–#970), where a count that maps to no portion falls back to the one unit that silently rescales the row.

### What this pass did not cover

- **OpenAI and OpenRouter were not run**, blocked by **#974**. Switching provider needs the AI Assist row on the You tab, and it could not be driven: `settings-ai-assist` never appears in the UIAutomator tree at any scroll position, and focus traversal skips it from both directions — as do `settings-food-sources`, `settings-health-sync` and `settings-import-custom-food`, while `settings-delete-all-data` in the same group is exposed normally. The row is painted and usable by hand. Whether that is an accessibility gap in `SettingsScreen(embedded: true)` or a device artifact, #830's stated method — driving by Semantics identifier over ADB — cannot reach these rows on this build, and no test references `settings-ai-assist` or covers the embedded form. Filed as **#974**, where the first job is establishing with TalkBack whether this also puts the AI Assist dialog out of reach for screen-reader users — which would bear directly on the consent story in this description.
- #830 cases 15–18 and 21–23 remain open for the reasons already recorded there: they need a person at the lens, a deliberately bad key, or a real OpenAI-compatible server.

simonoppowa:
## #974 resolved: confirmed real, but **not** a blocker for this PR

Updating what I wrote above, where I left the accessibility question open.

**It is user-facing.** With TalkBack on, swiping right from `Tag beginnt um` jumps straight to *App-Daten exportieren*. Four Settings rows cannot be reached without sight — food databases, Health Connect, custom food import, and AI Assist. For this feature that means no key, no provider, and no consent screen.

**It is not caused by this branch.** The same scan on the installed `.develop` build (`2.1.0-develop`, `develop` without the AI feature) shows the identical gap, and `develop` already carries the same settings structure, the same embedded form, and the same rows. #648 adds one more row to a band that was already unreachable.

So nothing here argues against merging #648. The honest statement is narrower and worth having on the record: this PR does not introduce the defect, but it does place a **privacy disclosure and a credential entry point** inside the affected band, which raises the cost of leaving it unfixed. Tracked separately as **#974** (`backlog:high`), against the app rather than this branch.

The merge-relevant finding on this PR remains **#973**.

simonoppowa:
## #979 — a provider with a stored key cannot be switched to

Found trying to switch to openrouter for #830 case 12, and unlike #974 this one **is** introduced by this branch.

Tapping a provider row writes it immediately, and then every exit from the dialog undoes it: **OK is not rendered** for a provider that already has a key (`if (!_loading && (!_hasKey || _provider == AiProvider.ownServer))`), **ABBRECHEN** restores the previous provider, **Back** routes to the same handler, and `barrierDismissible: false` rules out the fourth. With keys stored for more than one hosted provider — the ordinary state once someone has configured a second — no switch between them is possible.

The two provider switches earlier in this pass looked like they worked because the harness force-stopped the app before the dialog could revert. That was an accident of the driver, not the UI working.

Worth noting how it stayed hidden: `ai_assist_dialog_test.dart` has a `leaving without confirming (#848)` group pinning that Cancel restores the provider, but nothing asserts a confirm path exists for a keyed provider, and no test asserts the OK button's presence at all. The one test exercising the switch and OK together is for `ownServer` — the single provider the gate lets through.

Filed as **#979** (`backlog:high`). Choosing between destinations is the operation the four-destination design exists to offer, so this reads as merge-relevant rather than a follow-up.

PR #672, opened 2026-08-15
fix(android): expose per-app language picker via localeConfig

### LINKED ISSUE #626: Android per-app language picker is unavailable: localeConfig is not declared
Found while trying to view the Chinese UI on a Pixel 6 (Android 17, SDK 37) to verify a localization fix.

## What happens

```
$ adb shell cmd locale set-app-locales com.opennutritracker.ont.opennutritracker --locales zh-CN
$ adb shell cmd locale get-app-locales com.opennutritracker.ont.opennutritracker
Locales for com.opennutritracker.ont.opennutritracker for user 0 are [zh-CN]
```

The override is accepted and stored. After a force-stop and relaunch the app still renders in German — the system language. The setting has no effect.

The cause is in the manifest:

- `android:localeConfig` is not declared on `<application>` in `android/app/src/main/AndroidManifest.xml`
- there is no `res/xml/locales_config.xml` (the only file in `res/xml/` is `data_extraction_rules.xml`)

Android's per-app language feature reads that declaration to learn which locales an app supports. Without it the platform stores the preference and ignores it, and **the "Language" entry does not appear under Settings → Apps → OpenNutriTracker** at all.

## Why this is worth fixing but is not urgent

**The app is not untranslatable from the UI.** There is an in-app picker — `_SettingsTile` with `settingsLanguageLabel` at `settings_screen.dart:304-311`, opening `_showLanguageDialog`, persisting through `setSelectedLocale` and read back at `main.dart:83`. Nine locales ship and all of them work.

So this is about discoverability and platform convention, not capability:

1. **Users look in the OS place.** Since Android 13 the per-app language picker is where people expect to change one app's language, and for this app it is absent.
2. **The in-app picker is not easy to find.** It lives inside a collapsed `ExpansionTile` on the profile page, several scrolls down. I could not reliably reach it by UI automation, which is a hint about how a user fares.
3. **Testing a single locale requires changing the whole device language.** That makes verifying a translation far more disruptive than it should be, for maintainers and translators alike.

## What the fix involves

Add `res/xml/locales_config.xml` listing the nine shipped locales (`cs de en it pl sk tr uk zh`) and reference it with `android:localeConfig="@xml/locales_config"` on `<application>`. AGP 8.1+ can generate the file from `androidResources.generateLocaleConfig` if that is preferred over hand-maintaining it, which also keeps it from drifting as locales are added.

**The part that needs thought, and why this is not a one-line change:** the app already keeps its own `selectedLocale` in config and applies it over the platform locale. Once the OS-level override also exists, the two can disagree — a user could pick Chinese in the app and Polish in Android settings. Whichever way that is resolved should be deliberate. The simplest coherent answer is probably to let the OS setting win and have the in-app picker write through to it via `LocaleManager`, but that is a design decision.

## Not related to AI

Surfaced during work on the bulk-add feature but entirely independent of it.

t-khush:
## What / Why
Fixes #626

- Android 13+ per-app language (Settings -> Apps -> OpenNutriTracker -> Language) does not appear without `android:localeConfig`.
- `adb shell cmd locale set-app-locales` is accepted but ignored, so QA needs full device language switch.
- In-app picker works but lives inside a collapsed ExpansionTile, low discoverability.

## Solution
- Add `android/app/src/main/res/xml/locales_config.xml` listing the 9 shipped locales: `cs de en it pl sk tr uk zh` plus `zh-CN`/`zh-Hans` aliases so the OS maps Chinese variants correctly.
- Reference it from `<application>` with `android:localeConfig="@xml/locales_config"` in `AndroidManifest.xml`.

## Scope & trade-offs
This is the minimal platform-convention fix described in #626. No Kotlin/gradle changes, AGP 8.1+ auto-supported.

The issue notes a design question: once both sources exist, app `selectedLocale` vs OS override can disagree (e.g. app=zh, OS=pl). Current behavior: in-app remains source of truth; OS override is honored only if user uses system picker. A follow-up can make the in-app dialog write through via `LocaleManager` so the two stay in sync — deliberately out of scope here to keep this PR small and mergeable.

## Testing done
- `aapt2 dump xmltree` shows `localeConfig` present.
- Listing includes all 9 arbs under `lib/l10n`.
- Build check: manifest parses, file path exists (no new dependencies).

Closes #626


jordan-lee-code:
Thank you, Khush, and thank you especially for naming the loose end rather than leaving it to be discovered. You were right that `localeConfig` is the missing platform convention and right that the two sources of truth would disagree once both existed, and that honesty is what made this straightforward to finish.

I have taken the follow-up on rather than asking you to. A picker that appears in Android's Settings and silently does nothing for anyone who has ever used the in-app one felt like a worse place to leave people than no picker at all, so both now write to both sides: choosing in the app tells the OS, an OS override wins at startup and is saved back so Settings agrees, and changing it in Android Settings while the app is running arrives as `didChangeLocales` and is adopted straight away. An install that predates this seeds the system from its saved choice rather than being overwritten by it.

It goes through the framework `LocaleManager` rather than `AppCompatDelegate`, because `FlutterActivity` extends `FragmentActivity` and `setApplicationLocales` would have quietly done nothing there. No new Gradle dependency and nothing on the iOS side touched.

Two notes on the mechanics. I had to force-push, because the branch sat on a squash-merged `main` commit so `develop` was not in its history and repointing would have shown this as re-landing the whole 2.0.2 release. Your commit is unchanged and still yours, only its parent moved. I also collapsed the three Chinese rows to one: we ship a single Simplified ARB, so offering `zh`, `zh-CN` and `zh-Hans` as separate choices asks someone to pick between three things that are the same thing.

Analyzer clean, suite green at 997. Still to do before merge is a look at the real OS picker on hardware, to confirm the rows read "English" and "Deutsch" rather than the raw tags.

jordan-lee-code (approved):
Reviewed and validated, including the localeConfig wiring checked against the built APK. See the review comment above.

Issue #670, opened 2026-08-15
What must the formal privacy policies say about the OpenRouter path?


simonoppowa:
## Question

There are **two** iubenda policies, both linked from the app and the README, and both go stale the moment OpenRouter ships:

- EN — `privacyPolicyURLEn` → https://www.iubenda.com/privacy-policy/53501884
- DE — `privacyPolicyURLDe` → https://www.iubenda.com/privacy-policy/53922100

[What does the privacy claim become when the user picks the destination?](https://github.com/simonoppowa/OpenNutriTracker/issues/660) fixed what must be true in the README and the app. This ticket decides what must be true in the formal document, which is a different audience with different obligations — and which the README itself points at as the authority.

Decide what the policies must state:

- **OpenRouter as a recipient**, plus a downstream provider that varies by model. The curated path pins to a named vendor; the advanced free-text path cannot name one in advance.
- **Photographs as a data category** leaving the device to a third party. The current policy predates the photo feature entirely.
- **The forwarded account-level identity.** [#664](https://github.com/simonoppowa/OpenNutriTracker/issues/664) established this reaches Anthropic and OpenAI on every request and cannot be suppressed. It is a personal-data flow the policy does not currently describe.
- **Divergent retention.** Anthropic's unqualified ephemerality applies only to the direct path; OpenRouter's analogue carries an abuse-detection, security, billing and legal-compliance carve-out ([#654](https://github.com/simonoppowa/OpenNutriTracker/issues/654)).
- **Who is controller for this traffic.** Under BYO-key the user holds the account and the billing relationship. `docs/ai-legal-constraints.md` (**local to the maintainer's working tree, deliberately not committed**) already examined this against EDPB Guidelines 07/2020 and recorded it as *"a genuine judgement call, not a settled question"* — so the decision here is which position to take and to state, not to re-derive it.

Also practical, and possibly limiting: **iubenda generates these from declared services.** Whether its model can express "a provider the end user selects at runtime" at all, or whether that needs free-text sections, shapes what is achievable.

Editing the documents is manual work for the maintainer and is not part of this map; deciding what they must contain is.

**Not legal advice, and this ticket should not pretend otherwise** — its output is a list of what changed and what must therefore be represented, in a form a lawyer can act on.

simonoppowa:
## Resolution

**Not legal advice.** This is a list of what changed and what must therefore be represented, for a lawyer to act on. Where a position has to be taken rather than derived, it is marked as a position.

[...]

Issue #668, opened 2026-08-15
Which models go on the curated list, and how is it kept current?


simonoppowa:
## Question

[How does the user choose an OpenRouter model?](https://github.com/simonoppowa/OpenNutriTracker/issues/658) settled the mechanism — a curated list plus an advanced free-text field — but not the contents.

Decide:

- **Which models are blessed.** They must clear four bars: `tools` + `tool_choice` + `image` advertised *per endpoint* rather than at model level (the model-level flag is a union and over-reports); a first-party provider to pin to; a policy that survives [#656](https://github.com/simonoppowa/OpenNutriTracker/issues/656); and empirical evidence from [#666](https://github.com/simonoppowa/OpenNutriTracker/issues/666) that forcing is actually honoured.
- **Whether the text path and the photo path get different lists.** 341 models advertise tools; only 215 also take an image. A text-only model is perfectly usable for reading a typed meal.
- **How the list is kept current** — hardcoded and shipped with releases, fetched and cached, or hardcoded with a staleness check. Models are deprecated and added continuously, and a solo maintainer owns whatever is chosen.
- **The model-to-provider pinning map**, which is required regardless and is not derivable from the slug: `qwen/*` is served first-party by Alibaba, `google/*` maps to two provider slugs.

The default entry is an exact id, never an alias — settled in [#658](https://github.com/simonoppowa/OpenNutriTracker/issues/658) after `openai/gpt-5.6` was observed resolving server-side to `openai/gpt-5.6-sol`.

---

## Added by [the forced-tools probe](https://github.com/simonoppowa/OpenNutriTracker/issues/666)

`supported_parameters` **under-reports as well as over-reports**. `amazon/nova-lite-v1` does not advertise `tool_choice`, yet returned a valid forced tool call with a clean schema when `require_parameters` was omitted — and 404s when it is set.

So the list cannot be derived from the catalogue flags in either direction, and `require_parameters: true` will reject models that demonstrably work. Membership has to be established empirically, per model, which is a bounded but real curation cost.

---

## Added by [the non-food probe](https://github.com/simonoppowa/OpenNutriTracker/issues/669)

Capability flags are not fitness. `openai/gpt-5.4-nano` advertises `tools`, `tool_choice` and `image`, passes every filter, and returned **empty for both food-bearing photos** in the non-food slice — a user photographing a real meal would be told "No food found in that photo" and get nothing.

[...]

Issue #661, opened 2026-08-15
What do the 27 strings that name Anthropic become?


simonoppowa:
## Question

Three ARB keys name Anthropic in user-visible text — `aiAssistDisclosureLabel`, `aiAssistKeyFieldLabel`, `bulkAddPhotoDisclosureLabel` — across nine locales, so 27 strings.

Decide whether they become provider-neutral ("your provider"), parameterised with the provider name as a placeholder, or split per provider. Weigh: a placeholder keeps the sentence specific and true but adds ICU plumbing to a disclosure that must stay legally precise in nine languages; neutral wording is one edit but vaguer exactly where precision matters most.

Includes the settings field label, which currently reads "Anthropic API key" and would be wrong for an OpenRouter key.

---

## Added by [what happens when the model cannot force a tool call](https://github.com/simonoppowa/OpenNutriTracker/issues/659)

A sixth photo-failure case is needed: *this model cannot do this*. It is none of the existing five — "try another photo" is wrong, "try again later" is wrong, and it is not a permission problem.

Only the photo path needs wording. On the text path the read falls back to the deterministic parser, and that fallback is deliberately silent.

So this ticket now covers the three Anthropic-naming keys **plus** one new string for the capability refusal, pointing at the model setting rather than at the photo.

---

## Added by [what the privacy claim becomes](https://github.com/simonoppowa/OpenNutriTracker/issues/660)

The disclosure now has to distinguish **two cases**, and the two surfaces differ:

- the **settings dialog** covers both providers, including the advanced free-text path whose destination cannot be named;
- the **photo sheet** now only ever appears on a nameable path, since the free-text field is text-only.

Three facts must be expressible in whichever wording is chosen:

1. OpenRouter is a broker — a request touches it *and* a provider.
2. An account-level identity is forwarded to some providers and cannot be disabled.
3. Anthropic's unqualified "ephemeral" retention quote applies to the Anthropic-direct path **only**; OpenRouter's analogue carries an abuse-detection/billing/legal carve-out.

Also to decide here: whether the serving provider is surfaced per request, now that `X-OpenRouter-Metadata: enabled` is always sent and the app therefore knows it.

simonoppowa:
## Resolution

### The premise the ticket started with was wrong

This was filed as "make 27 strings provider-neutral or parameterised". Reading them in full killed both options:

[...]

Issue #660, opened 2026-08-15
What does the privacy claim become when the user picks the destination?


simonoppowa:
## Question

The README does not say "we respect privacy" — it names four destinations and invites you to check. The AI section quotes Anthropic's retention policy verbatim. A user-selected OpenRouter model makes the destination a variable, and possibly a variable the app cannot name.

Decide what replaces the current wording in the four-destinations table and in the in-app disclosure, given whatever the research tickets establish. Weigh honesty against legibility: "your chosen provider, whose policy we cannot quote" is truthful and much weaker than what is there now.

Also decide whether an unknown-retention provider should be allowed to carry a *photo*, or whether the photo path stays Anthropic-only — the disclosure currently promises the picture is sent for one request and kept by nobody.

---

## Added by [How does the user choose an OpenRouter model?](https://github.com/simonoppowa/OpenNutriTracker/issues/658)

The advanced free-text model field **cannot be pinned** — an arbitrary id has no entry in the provider map — so on that path the destination is genuinely unpredictable and the app cannot name it in advance, even with `X-OpenRouter-Metadata` enabled.

So the privacy wording has to cover three cases, not two:

1. Anthropic direct — the destination is known and its policy is quotable, as today.
2. OpenRouter with a curated model — pinned to a named first-party vendor, so two destinations, both nameable.
3. OpenRouter with a hand-entered model — OpenRouter plus a provider it selects, which may vary per request.

simonoppowa:
## Resolution

### 1. The enumeration survives, and photos only go where the vendor is named

"Four destinations, nothing else" stays an enumeration rather than becoming a hand-wave. The table gains OpenRouter **and** the vendor serving the request — both nameable, because [#663](https://github.com/simonoppowa/OpenNutriTracker/issues/663) pins a curated model to the vendor its slug names.

**The advanced free-text field is text-only.** It is the one place the app cannot say in advance where a request goes, so it does not carry photographs.

The asymmetry is deliberate. A photo is the highest-stakes payload this app can send and the only one that cannot be un-sent — it can contain a kitchen, a face, a room. Text is a meal description the user typed. If exactly one path has an unnameable destination, it should be the one carrying the least.

### 2. The identifier is accepted and disclosed, not worked around

[...]

PR #676, opened 2026-08-15
feat(settings): let the user pick a provider, and make OpenRouter reachable

simonoppowa:
**This is the whole feature, in one diff.** It stacks on [#673](https://github.com/simonoppowa/OpenNutriTracker/pull/673) (the OpenRouter client) and [#674](https://github.com/simonoppowa/OpenNutriTracker/pull/674) (per-provider credentials), so those two commits appear here as well. Merge them first and this diff shrinks to the settings surface alone; review it as it stands and you see everything at once.

Builds wayfinder [#658](https://github.com/simonoppowa/OpenNutriTracker/issues/658), [#660](https://github.com/simonoppowa/OpenNutriTracker/issues/660), [#661](https://github.com/simonoppowa/OpenNutriTracker/issues/661) and [#668](https://github.com/simonoppowa/OpenNutriTracker/issues/668). Map [#653](https://github.com/simonoppowa/OpenNutriTracker/issues/653) is now fully resolved and this is the last of its build work bar the privacy policies.

## Provider first, everything else beneath it

The key, the model list and the disclosure all mean different things depending on who is answering, so they are shown as consequences of one choice rather than as four independent settings. Splitting them across separate Settings rows would let someone save a key for a provider they are not using and see nothing wrong.

Switching to a provider with no key goes **quietly unavailable** rather than erroring — the camera action hides itself and the text path falls back to the deterministic parser. That is the state `MealPhotoUnavailable` already describes, arriving by a new route, so it needed no new machinery.

## The disclosure is a shared base plus a provider paragraph

Not two full variants. The sentences true either way — diary never sent, nutrition always from the databases, key never leaves the device — exist **once**, so a correction cannot land in one variant and miss the other in a language nobody on the project reads. Whole sentences also concatenate safely across nine locales in a way an interpolated clause does not.

The OpenRouter paragraph is not the Anthropic one with a different noun. It has to state a **broker and a downstream vendor**, an **account-level identity forwarded on every request that cannot be switched off**, and retention with an **abuse / security / billing / legal carve-out**. Only `aiAssistKeyFieldLabel` is parameterised, because that really is the same sentence with a different vendor in it.

`aiAssistDisclosureLabel` is removed rather than left beside the new keys: a stale string naming one provider unconditionally is exactly the drift this split exists to prevent.

## Two curated models, and sonnet leads on measurement

| | staple identified, hard photos |
| --- | --- |
| **anthropic/claude-sonnet-5** | **16/16** |
| anthropic/claude-haiku-4.5 | 12/16 |

Haiku's four misses were confident, specific and wrong — `banh mi` for a basket of tacos, `Decorative candy arrangement` for oatmeal. Those are not empty answers; they resolve to *something* in the food search and have to be caught in the review screen. At roughly **$0.0033 against $0.0017** a photo, cost does not defend the weaker one.

Both are pinned to Anthropic with `allow_fallbacks: false`, so the vendor named in Settings is the vendor that answered. Unpinned, a request for `anthropic/claude-haiku-4.5` was served by Amazon Bedrock on all three attempts of a probe.

The list is hardcoded because membership requires a behavioural screen the app cannot run — the catalogue does not carry the property the list is selected on. A stale entry 404s, which is already classified as a capability refusal pointing at this screen.

## No advanced free-text field in this increment

That keeps every destination nameable and pinned, and it keeps two known problems **unreachable** rather than shipped:

- a typo'd model id returns **400**, which the taxonomy reads as a rejected *image* — wrong advice, and the free-text field is where typos happen;
- `strict: true` has no primary-source guarantee of uniform enforcement, and combined with `require_parameters: true` could 404 a model that would otherwise work.

Both are fine on the curated list — 447 live calls, zero occurrences — and both need fixing before that field ships.

## The wiring is asserted, not assumed

`mealItemsApiFor` is a free function rather than a private locator helper specifically so this can be tested. Choosing OpenRouter really produces an OpenRouter request, pinned to the vendor the screen named; a stored model id is honoured; a retired one falls back instead of 404ing forever; and **one provider's key never reaches the other's endpoint**.

Five mutations run against those tests, all five killed: building the wrong client for a provider, dropping the pin, ignoring the stored model, trusting a retired id, and promoting haiku to default.

## The README enumeration survives as an enumeration

It gains OpenRouter *and* the vendor serving it, states the broker hop, the forwarded identity and the weaker retention, and **scopes Anthropic's "ephemeral" quote to the direct path** instead of stretching it over both. The two existing sentences about identifiers are untouched and simply not generalised to the AI path.

## Verification

- **1284 tests**, `flutter analyze` clean on `lib` and `test`.
- 16 dialog tests, 11 wiring tests, both new.
- One thing the tests found that is worth knowing: the dialog is now tall enough that the enable switch and Remove button start **below the fold**. It scrolls, and the tests scroll to reach them like a user would — but it is worth a look on a real handset.

## Still outstanding, deliberately

The formal privacy policies ([#670](https://github.com/simonoppowa/OpenNutriTracker/issues/670)) are not touched here — that is maintainer work with a lawyer, and it is worth knowing the policies are already behind what shipped *before* this branch: both declare five recipients, none of them Open Food Facts, Supabase or Anthropic.

Also unrelated but adjacent: `privacyPolicyURLDe` is declared and never opened — both entry points show the English policy whatever the locale.


simonoppowa:
## Verified on a Pixel 6

Develop flavour installed side by side; the production app and its data were never touched. Device locale is German, so this also exercised the translations.

### The whole chain, end to end

A real request reached OpenRouter, pinned to Anthropic, on `anthropic/claude-sonnet-5`:

| step | result |
| --- | --- |
| key saved to the OpenRouter slot | `AiApiKeyTag.openrouter` present; no `.anthropic`, no legacy tag |
| settings tile names the destination | **"An — OpenRouter"** |
| camera action offered on bulk-add | yes — it only renders when a key is enabled |
| request built and sent | reached OpenRouter, no fallback line in the log |
| model split a sentence the parser cannot | `q=Eier` and `q=Glas+Milch` — two searches from one separator-free line |
| rows resolved against the food databases | 2 rows |
| AI review notice | *"Von KI gelesen — prüfe jede Zeile"* |

The deterministic parser produces **one** row from that input, because it splits on `,;\n+` and there is no separator. Two rows is proof the model answered.

### The disclosure and picker, in German

Confirmed on hardware: the OpenRouter paragraph states the broker, the forwarded identity (*"eine kontobezogene Kennung, und das lässt sich nicht abschalten"*) and the retention carve-out (*"Missbrauchserkennung, Sicherheit, Abrechnung oder Erfüllung rechtlicher Pflichten"*). Switching to Anthropic drops the identity clause and the carve-out, hides the model radios, states the single pinned model, and keeps the shared paragraph. The vendor is named once as *"Bereitgestellt von Anthropic"*.

### Two rows came back at 3 kcal and 0 kcal, and that is correct

`2 Eier` resolved to 2 **g** of egg, `1 milch` to 1 **ml**. The matched Open Food Facts records carry no parseable `servingQuantity`, so `_initialUnit` deliberately refuses to label them `serving` — that would read "2 serving" and still log 2 g, renaming [#622](https://github.com/simonoppowa/OpenNutriTracker/issues/622) rather than fixing it. It falls through to weight and flags the row instead: both showed *"Einheit für diese Menge prüfen"*. The guard fired exactly as designed.

### Three layout defects found here that no widget test could see

1. **The key field was entirely below the fold** on the OpenRouter path — its disclosure is three sentences longer. `find.byType(TextField)` was green throughout, because a widget scrolled out of view is still in the tree.
2. **The disclosure was clipped mid-word with no affordance**, so it read as finished.
3. **The scrollbar added to fix (2) clipped the text horizontally** — the fix reintroduced a smaller version of its own bug.

All three are fixed and pinned by tests that measure the real thing (the field's rect against a Pixel 6 viewport, `thumbVisibility` on the Scrollbar), both mutation-checked.

### One finding worth its own decision

**A rejected key is completely invisible on the text path.** The first live attempt failed with `401`, and the only reason anyone knows is `adb logcat`:

```
WARNING: OpenRouterMealItemsApi: Interpreter call failed with 401
INFO: ReadMealTextUseCase: Interpreter unavailable, using the parser: provider returned 401
```

The screen showed a plausible result from the deterministic parser. A user in that state would never learn their key is wrong, while every request still costs a round trip.

The photo path distinguishes auth and says so, because it has no fallback. Text swallows it. Falling back silently is right; falling back silently *forever* on a permanently actionable failure is not. Not fixed here — it predates this branch and deserves its own decision.

### Not covered

The photo path on device. It needs a camera grant and the system picker, neither of which an automated driver should be doing on someone's phone.

PR #712, opened 2026-08-18
fix(scanner): report a not-found barcode as not found

EugeneSusla:
### Problem

Scanning a barcode that Open Food Facts does not have shows **"Error while fetching product data"** with a Retry button, rather than the "Product not found" message the app already ships.

`ScannerBloc` classifies the failure with:

```dart
if (exception == ProductNotFoundException) {
```

That compares the caught exception *object* against the *type*, which is never equal. The `productNotFound` branch is therefore unreachable, and every OFF 404 falls through to the generic error state. `errorProductNotFound` and its wiring in `scanner_screen.dart` were already in place and simply never reached.

The Retry button is especially misleading here: `OFFDataSource` deliberately does not retry a not-found (`shouldRetry: (e) => e is! ProductNotFoundException`), so retrying a 404 can never succeed.

Reproduces with any barcode absent from OFF — for example an in-store GS1-128 label, whose GTIN is restricted-circulation and so never registered globally.

### Change

- `exception == ProductNotFoundException` → `exception is ProductNotFoundException`.
- Adds `ScannerBloc` tests covering both failure classifications.

The test asserts the barcode actually reached the search use case, and that the settled state is a failure rather than filtering the stream for one, so a swallowed exception fails with a readable diff instead of passing on the bloc's catch-all.

### Verification

- New test verified red before the fix, green after.
- `flutter analyze` clean; full suite green (964 tests).

### DIFF
--- a/lib/features/scanner/presentation/scanner_bloc.dart
+++ b/lib/features/scanner/presentation/scanner_bloc.dart
@@ -31,7 +31,7 @@ class ScannerBloc extends Bloc<ScannerEvent, ScannerState> {
           ),
         );
       } catch (exception) {
-        if (exception == ProductNotFoundException) {
+        if (exception is ProductNotFoundException) {
           emit(
             const ScannerFailedState(ScannerFailedStateType.productNotFound),
           );
--- a/lib/features/scanner/presentation/scanner_state.dart
+++ b/lib/features/scanner/presentation/scanner_state.dart
@@ -25,7 +25,7 @@ class ScannerLoadedState extends ScannerState {
   });
 
   @override
-  List<Object?> get props => [product];
+  List<Object?> get props => [product, usesImperialUnits];
 }
 
 class ScannerFailedState extends ScannerState {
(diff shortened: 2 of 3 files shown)


jordan-lee-code:
Lovely catch, Eugene, and the kind that hides in plain sight for years. Comparing an instance against a `Type` made the whole branch unreachable, so every 404 got a Retry button that `OFFDataSource` guarantees can never succeed. Being told to try again when trying again cannot possibly work is a small cruelty, and I am glad it is going.

I have pushed one commit on top rather than asking you to, since it turned up while confirming yours. `ScannerFailedState` left `type` out of `props`, so a not-found and a fetch error compared equal, and `emit` drops a state equal to the current one. `ScannerLoadedState` left out `usesImperialUnits` the same way. Neither is reachable today, because a loading state always sits between two failures, and the tests say so explicitly rather than pretending otherwise. They are just the sort of thing that resurfaces later as a screen showing the wrong message with no obvious cause.

Analyzer clean, full suite green, and the equality tests fail without the fix. Recommending merge.

jordan-lee-code (approved):

Issue #748, opened 2026-08-20
Does ATS reach dart:io on iOS? Answer it in the simulator job


simonoppowa:
## Question

[Does the cleartext gate actually apply to this app's HTTP stack?](https://github.com/simonoppowa/OpenNutriTracker/issues/746) measured Android and found the answer is **no**: cleartext to a LAN address connects today with nothing declared in the manifest. iOS was left unverified for want of an Apple device.

It matters as much as Android did. If ATS also does not reach `dart:io`, then on both platforms this feature is already possible and the whole cleartext question is about what the app *claims*. If ATS **does** apply, iOS and Android diverge, and the options in `docs/ai-cleartext-http-policy.md` have to be chosen per platform rather than once.

**There is a route that needs no hardware.** This repo already runs an `ios-integration-tests` job on a simulator, and ATS applies in the simulator. So the same probe #746 used can answer it:

- A temporary integration test that `http.get`s a plain-HTTP address and reports whether it connects or throws, running inside the real app package under the real `Info.plist`.
- Reaching a LAN address from a simulator is awkward; the host's own `127.0.0.1` is reachable from the simulator directly, which is enough to answer the ATS question — but note that Apple, like Android, may treat loopback specially, so say which case was measured rather than generalising from one.
- Record the exact exception if it throws. `NSErrorFailingURLStringKey` and an ATS-specific message would be conclusive; a generic `SocketException` would not.

Two things to keep in mind from #746, where they cost time:

- Confirm what the **merged** `Info.plist` in the built app actually contains, not what the source declares — the Android equivalent was worth checking and could have invalidated the result.
- The **local-network permission** prompt is a separate gate that applies regardless of the ATS answer, and on a simulator it may not appear at all. Do not read its absence as evidence about a device.

Resolved when there is a measured yes or no for iOS written down, with the case that was measured named.

simonoppowa:
## Measured on iOS: ATS does **not** reach this app's HTTP stack either

From the `ios-integration-tests` job on the throwaway [#750](https://github.com/simonoppowa/OpenNutriTracker/pull/750), now closed unmerged:

```
PROBE ats-loopback http://127.0.0.1:61083/     -> CONNECTED status=200 body=ats-probe-ok
PROBE ats-lan      http://192.168.64.11:61088/ -> CONNECTED status=200 body=ats-probe-ok
PROBE platform     ios Version 18.5 (Build 22F77)
```

[...]

Issue #746, opened 2026-08-20
Does the cleartext gate actually apply to this app's HTTP stack?


simonoppowa:
## Question

[#734](https://github.com/simonoppowa/OpenNutriTracker/issues/734) resolved with a finding sharp enough to deserve its own ticket, because it decides whether the other options are even on the table.

Both platform vendors scope their cleartext policy to their own networking APIs and disclaim the layer beneath. Android: *"there's no expectation that the `Socket` API honors this flag."* Apple: *"ATS doesn't apply to calls your app makes to lower-level networking interfaces."*

This app reaches the network through `package:http` → `dart:io`, which is raw and secure sockets. The research found **no cleartext check** in the pinned Dart SDK, the `http` package or the Flutter framework, and no `NetworkSecurityPolicy` reference in the engine binary.

**If that is right, cleartext to a LAN address already works today**, and the whole question changes from *what may the app do* to *what should the app claim*.

Settle it by measurement, not by reading:

- Build a debug APK with **no** manifest change and point a plain `package:http` GET at an `http://` address on the LAN. Record whether it connects or throws, and the exact exception if it throws.
- Repeat on iOS with no `Info.plist` change, on a device at the shipping deployment target (15.5).
- Try loopback separately from a LAN address on both — Android documents a localhost special case, and they may not behave alike.
- Note the **iOS local-network permission prompt** if it appears, since that gate is independent of cleartext and applies either way.

The endpoint from [#739](https://github.com/simonoppowa/OpenNutriTracker/issues/739) will serve, but any HTTP server on the LAN answers this — it does not need a model behind it, so this ticket is not blocked on that one.

Resolved when both platforms have a measured yes or no written down. The decision about what to declare stays with the options table in `docs/ai-cleartext-http-policy.md`.

simonoppowa:
## Measured on Android: the cleartext gate does **not** reach this app's HTTP stack

Run on a Pixel 6, twice, with **no manifest change**. Confirmed first that the merged debug manifest (`build/app/intermediates/merged_manifests/fullDebug/.../AndroidManifest.xml`) declares neither `usesCleartextTraffic` nor `networkSecurityConfig` — so the debug build carries the same posture as release, and the result is not an artefact of building debug.

```
PROBE lan-cleartext      http://192.168.188.21:8099/probe.txt
  -> CONNECTED status=200 body=cleartext-probe-ok
PROBE loopback-cleartext http://127.0.0.1:8099/probe.txt
[...]

PR #750, opened 2026-08-21
probe: does ATS reach dart:io on iOS? (#748 — do not merge)

simonoppowa:
Throwaway probe for [#748](https://github.com/simonoppowa/OpenNutriTracker/issues/748). **Not for merge** — close it once the answer is recorded.

[#746](https://github.com/simonoppowa/OpenNutriTracker/issues/746) measured the Android half on a Pixel: a plain-HTTP GET to a LAN address connects today with nothing declared in the manifest, so Android's cleartext policy does not reach `package:http` → `dart:io`. iOS was left unverified for want of Apple hardware.

This PR exists only to reach the `ios-integration-tests` job, which runs on a simulator where ATS applies. The answer is in that job's log, on lines starting `PROBE`.

The test is **self-contained** — it binds its own HTTP server inside the test, so nothing needs to run on the runner and the workflow is untouched. It measures loopback and a real non-loopback interface separately, because Apple may treat loopback specially and a loopback-only result would not generalise.

It asserts nothing that can fail: a red probe would trip the job's retry loop and buy a second cold build for nothing.

Already validated on Android, where it reproduces #746 by a different mechanism (self-bound server, the phone's own wifi address):

```
PROBE ats-loopback http://127.0.0.1:45593/     -> CONNECTED status=200 body=ats-probe-ok
PROBE ats-lan      http://192.168.188.22:39585/ -> CONNECTED status=200 body=ats-probe-ok
PROBE platform     android CP2A.260705.006
```

### DIFF
--- a/integration_test/zz_ats_probe_test.dart
+++ b/integration_test/zz_ats_probe_test.dart
@@ -1,0 +1,105 @@
+// TEMPORARY probe for #748. Delete once the result is recorded.
+//
+// Question: does Apple's App Transport Security reach this app's HTTP stack?
+// #746 measured the Android half and found it does not — a plain-HTTP GET to a
+// LAN address connects today with nothing declared in the manifest. Both
+// vendors scope their policy to their own networking APIs and disclaim the
+// layer beneath, and this app goes through `package:http` -> `dart:io`, which
+// is BSD sockets rather than NSURLSession.
+//
+// Self-contained on purpose: it binds its own server inside the test, so it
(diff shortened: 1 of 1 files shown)


simonoppowa:
Answered. `PROBE ats-lan http://192.168.64.11:61088/ -> CONNECTED status=200` on iOS 18.5 — ATS does not reach `dart:io` either. Recorded on [#748](https://github.com/simonoppowa/OpenNutriTracker/issues/748); closing unmerged as intended.

PR #775, opened 2026-08-21
fix(add-meal): convert Open Food Facts micronutrients into the app's units

### LINKED ISSUE #716: [Bug]: Micro grams (μg) are displayed wrong
### Description

For product (at least from OpenFoodFacts) that have nutrients which are displayed in μg will (I assume) display grams instead of μg. Which results in just displaying  0.0μg.

This occurs on different places of the App, eg. Product search, but also at the 'Todays nutrients' sheet. Which displays eg. Vitamin D always with 0.00μg.

### Steps to reproduce

1. Press the plus button
2. Search for a OpenFoodFacts Product which have nutrients that are measured in micro grams (Vitamin D)
3. Check the Micronutrients sheet

### Expected behaviour

Display the right micro gram value.

### Actual behaviour

Displays 0.0μg

### Platform

iOS

### OS version

iOS, Android

### App version

2.0.2

### Install source

Apple App Store

### Feature area

Micronutrients

### Screenshots / recordings

Image

### Additional context

I assume its because OpenFoodStats returns the values always in gram, and there is no conversion happening on the UI side. Calculations should probably also checked, in regard to the daily nutrients summary.

### Checklist

jordan-lee-code:
## Summary

Open Food Facts normalises every `<nutrient>_100g` field to grams, whatever unit the packaging used. `MealNutrimentsEntity` carries each micronutrient in the unit the app displays it in, milligrams for the minerals and micrograms for vitamins A, D and B12, and `fromOffNutriments` was copying the values straight across. Every mineral was stored a thousand times too small, and those three vitamins a million times too small.

That is what Genfood reported in #716: vitamin D reading `0.00µg` on every Open Food Facts product, in search results and in the Today's nutrients sheet alike, because 5µg had been saved as 0.000005.

## Why it is worth more than the sheet it was reported against

`NutrientPanelTotals.fromIntakes` is deliberately unit-agnostic. It scales the stored per-100 value and hands the result to the panel, and the panel compares it against `DriReference` amounts in mg and µg. So a food scanned from Open Food Facts was contributing almost nothing to any micronutrient goal, quietly, without ever looking broken. `ComputeRecipeNutritionUseCase` sums the same fields, so a recipe mixing an Open Food Facts ingredient with a Supabase one was adding grams to milligrams.

## Changes

- Two small private helpers on the factory convert grams to milligrams (cholesterol, the seven minerals, vitamin C, vitamin B6, niacin) and grams to micrograms (vitamins A, D and B12). Both null-pass, matching the surrounding `asDoubleOrNull()` shape.
- Macros and the mono/poly/trans lipids are grams on both sides and are untouched.

The other two sources were already correct, which is what made the mismatch legible once you looked at them side by side: `fromSpFoodSummary` pivots into the app's units and `fromFdcNutriments` reads FDC values that are natively mg and µg. The same field meant milligrams from one database and grams from another.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] CI / tooling
- [ ] Localization
- [ ] Other

## Related issues

Closes #716

## Test plan

- [x] Unit tests added or updated
- [x] `flutter analyze` clean

Detail:

- The existing micronutrient fixtures now carry the grams Open Food Facts actually sends, so the test reads the conversion in both directions at once. Two of the expectations had been pinning the bug (`sodium100 == 0.12`, against the FDC path's `120` for the same food).
- New: values copied from live API responses rather than invented, so the fixture cannot drift from what the API sends. Iron for `3168930010265` and sodium for `3017620422003`, both reported with `_unit: "g"`.
- New: the same food mapped from Open Food Facts and from FDC now comes out identical, field for field. That is #716 in one assertion.
- New: a declared zero survives as zero rather than becoming a null the panel ignores.
- `flutter test`: 1093 passing. `flutter analyze`: no issues.

## What this does not do

Meals logged before this keep the values they were saved with, so old entries read as they do today and nothing gets worse. Repairing that history means rewriting someone's intake records on upgrade, and that deserves its own change and its own review, so it is filed separately rather than folded in here.

## Checklist

- [x] Code follows project style
- [x] No new interactive widgets were added
- [x] No user-facing strings changed
- [x] No DBOs, DTOs, or environment fields changed

### DIFF
--- a/lib/features/add_meal/domain/entity/meal_nutriments_entity.dart
+++ b/lib/features/add_meal/domain/entity/meal_nutriments_entity.dart
@@ -153,33 +153,73 @@ class MealNutrimentsEntity extends Equatable {
       saturatedFat100:
           (offNutriments.saturated_fat_100g as Object?).asDoubleOrNull(),
       fiber100: (offNutriments.fiber_100g as Object?).asDoubleOrNull(),
-      // #237: Extended lipid profile
+      // #237: Extended lipid profile. These three are grams either way, so
+      // they pass through like the macros above.
       monounsaturatedFat100:
           (offNutriments.monounsaturated_fat_100g as Object?).asDoubleOrNull(),
       polyunsaturatedFat100:
           (offNutriments.polyunsaturated_fat_100g as Object?).asDoubleOrNull(),
       transFat100: (offNutriments.trans_fat_100g as Object?).asDoubleOrNull(),
-      cholesterol100:
-          (offNutriments.cholesterol_100g as Object?).asDoubleOrNull(),
+      cholesterol100: _gToMg(
+        (offNutriments.cholesterol_100g as Object?).asDoubleOrNull(),
+      ),
       // #237: Minerals
-      sodium100: (offNutriments.sodium_100g as Object?).asDoubleOrNull(),
-      potassium100: (offNutriments.potassium_100g as Object?).asDoubleOrNull(),
-      magnesium100: (offNutriments.magnesium_100g as Object?).asDoubleOrNull(),
-      calcium100: (offNutriments.calcium_100g as Object?).asDoubleOrNull(),
-      iron100: (offNutriments.iron_100g as Object?).asDoubleOrNull(),
-      zinc100: (offNutriments.zinc_100g as Object?).asDoubleOrNull(),
-      phosphorus100:
-          (offNutriments.phosphorus_100g as Object?).asDoubleOrNull(),
+      sodium100: _gToMg(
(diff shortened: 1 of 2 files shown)


Genfood:
Thanks. 🙏 

Maybe a stupid question, or I don't understand the source code well enough, but wouldn't it be much cleaner to work only with grams regardless of the nutrients?
While I understand this might be not the scope of this PR, nor up to you to decide, I can see a lot of problems that can occur because we treat some nutrients differently.

I would propose to only display some nutrients in micro gram on the UI side, and the rest of the calculations, as well as storing them should happen in grams.

What do you think about this?

PR #810, opened 2026-08-24
fix(l10n): stop re-pushing a per-app language override the user cleared

EugeneSusla:
Follow-up to #672. The launch-time reconciliation pushes the saved in-app language to the OS whenever the OS reports no override. That push was meant as a one-time migration for installs that predate the wiring, but nothing records that it happened — and "no override" is also exactly what the OS reports after the user picks **System default** in Android's per-app picker. Their clear is silently re-pushed on the next cold start, so the only way off the override is the in-app picker, with nothing suggesting so. Two neighbouring cases share the flaw:

- An OS override the app does not ship (say `ja`) folds to the same null, so the saved code was pushed **over** the user's OS-level choice — overwriting it rather than ignoring it.
- Clearing the override while the app is alive did nothing on screen: `_adoptSystemLocale` early-returned on null, the inert-picker failure in the clearing direction.

### The fix

Config gains `localeSyncSeeded` (Hive index 37), recording exactly one fact: **the OS has been seen holding an override.** `reconcileAppLocale` decides the ambiguous cases on the raw system tag instead of the folded language code:

- Supported override → system wins, saved if different, seeded recorded.
- Unshipped override → ignored on both sides (the OS demonstrably holds a value, so seeded is still recorded — switching it to System default later reads as the clear it is).
- Failed channel read → nothing is decided; a transient failure must never read as a deliberate clear. `AppLocaleService.getApplicationLocale` now reports `readFailed` distinctly instead of folding `PlatformException` into "no override" (`MissingPluginException` stays a clean "no override": on iOS/desktop the absence is permanent and true).
- No override, not seeded → the one-time migration push, **verified by reading the tag straight back**: on Android 13+ the value echoes back and seeded is recorded before the call returns, so a clear arriving any time later is honoured; on platforms with no per-app override the read-back stays empty and the branch repeats as a harmless no-op instead of ever mistaking the platform for a user who cleared their override.
- No override, seeded → the user picked System default: the in-app override is cleared to follow it.

`_adoptSystemLocale` now runs the same `reconcileAppLocale` instead of hand-rolling a second copy of the decision table, which gives the runtime path the seeded guard and the failed-read guard for free — and fixes the clearing direction while the app is alive. It is also serialized against itself: the migration push re-fires `didChangeLocales` mid-run, and two interleaved runs would double-apply one config snapshot; an early arrival queues a replay rather than being dropped.

### Verification

- Unit tests cover the full decision table: push-echo seeding, the cleared-override clear, the unshipped-tag ignore, the override-less-platform repeat (pushes asserted), the failed-read guard, and the existing supported-override cases.
- `flutter analyze` clean; full suite green (1093 tests).

### DIFF
--- a/lib/core/data/data_source/config_data_source.dart
+++ b/lib/core/data/data_source/config_data_source.dart
@@ -218,6 +218,14 @@ class ConfigDataSource {
     await _update((c) => c.selectedLocale = locale);
   }
 
+  Future<bool> getLocaleSyncSeeded() async {
+    return _readMerged().localeSyncSeeded ?? false;
+  }
+
+  Future<void> setLocaleSyncSeeded() async {
+    await _update((c) => c.localeSyncSeeded = true);
+  }
+
   Future<void> setConfigShowMicronutrients(bool show) async {
     await _update((c) => c.showMicronutrients = show);
   }
--- a/lib/core/data/dbo/config_dbo.dart
+++ b/lib/core/data/dbo/config_dbo.dart
@@ -152,6 +152,14 @@ class ConfigDBO extends HiveObject {
   // nothing has been deleted.
   @HiveField(36)
   List<String>? healthDeletedExternalIds;
+  // One-shot marker for [reconcileAppLocale]'s migration push: set once the
+  // saved language has been offered to Android's per-app language picker (or
+  // the picker was seen holding a value). Null means the push has not
+  // happened yet. Without it, an override the user cleared in the OS picker
+  // is indistinguishable from one that was never seeded, and gets silently
+  // pushed back on the next launch.
+  @HiveField(37)
+  bool? localeSyncSeeded;
 
   ConfigDBO(
     this.hasAcceptedDisclaimer,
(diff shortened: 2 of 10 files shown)


simonoppowa:
## Hold this out of 2.2.0 — there is a Hive field-index collision in it

Flagged by a release-readiness audit before [#988](https://github.com/simonoppowa/OpenNutriTracker/pull/988), not by review of the fix itself.

The change carries a **Hive field-index collision**, which means resolving it also means regenerating `config_dbo.g.dart`. A release merge is the worst moment for that: a wrong field index is a silent data-corruption class of bug, and it would ship alongside a release that already carries two major features neither of which has reached users.

The underlying fix — not re-pushing a per-app language override the user cleared — is real and worth landing. This is about **when**, not whether.

Recommend: resolve the index collision on its own, with its own regeneration and review, after 2.2.0 is out.

simonoppowa:
Carried onto today's `develop` as #1230, with your commit and authorship intact — the fix is as you wrote it. Two things had to move underneath it in the five weeks it waited: the Hive field is 40 now (37–39 were taken by other work), and the locale lists resolve against the shipped-language list that #1188 introduced. The review of the carried version also turned up one thing worth adding: after the runtime reconcile rewrites the config, the singleton `SettingsBloc` is reloaded, otherwise the Language tile could still show the cleared language and confirming the picker would push it back. Details in #1230's description.

Thank you for the fix and for the care in the decision table — it survived four adversarial passes with nothing but nits. Closing this in favour of #1230; if you see anything there that doesn't match your intent, say so on that PR.

PR #831, opened 2026-08-25
fix: accept Android document-provider export paths

### LINKED ISSUE #763: [Bug]: Export reports failure when Android auto-renames an existing ZIP
## Description

When `opennutritracker-export.zip` already exists in the selected Android destination, the system document provider auto-renames the new file, the export is written successfully, but OpenNutriTracker reports **Export/Import error**.

This is related to, but not a duplicate of, #504. In this case the exported file is non-empty and valid; only the app's reported result is wrong.

## Steps to reproduce

1. Ensure Downloads already contains `opennutritracker-export.zip`.
2. Open **You → Settings → Export / Import Data**.
3. Select **JSON** and tap **Export**.
4. Choose Downloads and save using the suggested filename.
5. Android creates an auto-renamed file such as `opennutritracker-export (2).zip`.
6. Return to OpenNutriTracker.

## Expected behaviour

The app reports a successful export when the document provider successfully writes the auto-renamed ZIP.

## Actual behaviour

The app displays `Fehler beim Export/Import` even though the new ZIP exists and is valid.

## Platform

- Android

## OS version

Android 17 / API 37 on Pixel 6 (`oriole`)

## App version

2.0.2, built from source at `145fca91764c872544f91bb0e317b979ef946417` (PR #752)

## Install source

Built from source

## Feature area

Export / import

## Logs or stack trace

The newly written file was 55,619 bytes. `unzip -t` passed for all seven entries:

- `user_activity.json`
- `user_intake.json`
- `user_tracked_day.json`
- `user_recipes.json`
- `recipe_images/<id>.webp`
- `weight_log.json`
- `custom_activity_templates.json`

Importing the same ZIP back into the app succeeded.

Likely area: `ExportWriteVerifier` may be checking the originally requested path while the Android document provider writes or returns an auto-renamed destination.

## Additional context

Found during the full Pixel regression for PR #752. Export code and `file_picker` were not changed by that dependency rollup.

## Checklist

jpdexter101-lang:
## What this fixes

Fixes #763.

On Android, `file_picker` 11 writes through the selected Storage Access Framework `content://` URI, then returns `uri.path`. That value looks like `/document/primary:Download/...` but is not a real `dart:io` filesystem path. When Downloads auto-renames an existing export to `(2).zip`, the export verifier tries to stat the provider path and reports a valid export as failed.

This change treats Android document-provider paths the same as `content://` URIs while retaining strict byte-length verification for real filesystem paths.

## Verification

- `dart format lib/core/utils/export_write_verifier.dart test/unit_test/export_write_verifier_test.dart`
- `flutter test test/unit_test/export_write_verifier_test.dart` — 7 tests passed
- `git diff --check` — passed
- `flutter analyze` was attempted, but the checkout is missing the repository's `lib/generated/l10n.dart` localization output and reports broad pre-existing generated-code errors; the focused test suite passes.

### DIFF
--- a/lib/core/utils/export_write_verifier.dart
+++ b/lib/core/utils/export_write_verifier.dart
@@ -24,11 +24,16 @@ class ExportWriteVerifier {
   const ExportWriteVerifier._();
 
   /// Throws a [StateError] if [savedPath] can be statted and its length
-  /// differs from [expectedByteLength], or if a non-content:// path cannot
+  /// differs from [expectedByteLength], or if a real filesystem path cannot
   /// be statted at all.
   static void verify(String savedPath, int expectedByteLength) {
-    if (savedPath.startsWith('content://')) {
-      // Not a path dart:io can stat - nothing more we can check.
+    // file_picker 11 returns Uri.path for Android document-provider saves,
+    // rather than the original content:// URI. Uri.path looks like
+    // /document/primary:Download/file.zip, but it is not a dart:io path that
+    // can be statted. The provider has already accepted the bytes by this
+    // point, so treating it as a missing local file creates a false failure.
+    if (savedPath.startsWith('content://') ||
+        savedPath.startsWith('/document/')) {
       return;
     }
 
--- a/test/unit_test/export_write_verifier_test.dart
+++ b/test/unit_test/export_write_verifier_test.dart
@@ -56,6 +56,18 @@ void main() {
       );
     });
 
+    test('does not throw for file_picker Android document-provider paths', () {
+      // file_picker 11 strips the content:// scheme and returns Uri.path,
+      // which looks like this even though it is not a local filesystem path.
+      expect(
+        () => ExportWriteVerifier.verify(
+          '/document/primary:Download/opennutritracker-export (2).zip',
+          4,
+        ),
+        returnsNormally,
+      );
+    });
+
     test('throws when a normal path cannot be statted at all', () {
       // A missing filesystem path means the write never landed — unlike a
       // content:// URI, there is no legitimate unreadable case to excuse.


simonoppowa:
## Not a 2.2.0 blocker, and one note if you do land it

Reviewed in a release-readiness audit before [#988](https://github.com/simonoppowa/OpenNutriTracker/pull/988).

The defect is real and somewhat broader than the description says — the false failure is not limited to the path shape the title names. It is a good fix.

It is not a release gate, though: it affects export rather than anything 2.2.0 introduces, and users hitting it have been hitting it since before v2.0.2.

If you do want it in: it is small (2 files, no `pubspec` change, no schema change), but it comes from a fork, so its workflow run needs approving first — **merge only on green**, since nothing in `main`'s branch protection will stop a red merge.

simonoppowa:
Closing as superseded, with thanks — the defect you named was real, and it is gone from `develop` by a different route.

The dependency migration in #1048 moved the app to file_picker 12, whose `saveFile` returns a `Uri` rather than a path string. `ExportWriteVerifier.verify` now takes that `Uri` and returns without checking anything unless the scheme is `file:` — so a document-provider save, which arrives as `content://…`, can no longer be mistaken for a missing local file. The case your test pins ("a path that is not a dart:io location") is covered on `develop` by the `content://` test in `test/unit_test/export_write_verifier_test.dart`; the `/document/…` string shape your fix keyed on no longer reaches the verifier at all, because there is no string any more.

Your patch targets the `String` signature, which is why it can no longer merge. Nothing further needed from you — and if you still see a false "could not save" on an Android document-provider export with 2.3.0, that would be a new bug worth its own issue.

Issue #836, opened 2026-08-25
Removing the last AI credential withdraws the agreement


simonoppowa:
## Parent

Part of the AI-assisted meal logging feature ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)), staged on `feature/ai-assisted-meal-logging` ahead of [#648](https://github.com/simonoppowa/OpenNutriTracker/pull/648).

## What to build

An agreement that cannot be withdrawn is not much of an agreement, and one that outlives the thing it authorised is worse — a stored "yes" with nothing left to say yes to.

Withdrawal should not need its own control. Removing the stored credential is already how a user turns the feature off, so that act should clear the agreement too: adding a credential again asks again.

**The wrinkle worth getting right.** Credentials are stored per provider, so more than one can be configured at once while the agreement itself is single and global. Clearing it when *any* credential is removed would leave a still-configured provider usable with no agreement recorded — the exact state this work exists to prevent. So the agreement is cleared when the **last** remaining credential goes, not the first.

## Acceptance criteria

- [ ] Removing the last configured provider's credential clears the agreement
- [ ] Removing one credential while another provider is still configured leaves the agreement standing
- [ ] After the agreement is cleared, saving a credential again shows the consent screen
- [ ] No new control is added for withdrawal, and no state exists where a credential is usable with no agreement recorded

## Blocked by

- #835

simonoppowa:
Fixed in #839, with review findings closed in #841. Both merged.

The agreement is cleared when the **last** credential goes, not the first — credentials are per provider while the agreement is single, so clearing on any removal would have left a still-configured provider usable with nothing recorded.

That uncovered an older bug in the same decision: `clear()` asked whether any *key* remained, the narrow question #755 retired elsewhere and `setEnabled` already documents having learned. Someone running their own server alongside a hosted key lost the feature by removing the hosted key. It asks `_isUsable` now.

#841 also made the agreement part of what `enabled` means, so a credential can never be usable with no agreement recorded — the invariant no longer lives only in UI flow.

Issue #834, opened 2026-08-25
Give the per-provider AI disclosure one source


simonoppowa:
## Parent

Part of the AI-assisted meal logging feature ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)), staged on `feature/ai-assisted-meal-logging` ahead of [#648](https://github.com/simonoppowa/OpenNutriTracker/pull/648).

## What to build

Prefactor, no behaviour change. Choosing which disclosure paragraph belongs to which provider currently happens privately inside the AI settings dialog. A second surface is about to need exactly the same words, and a copy would drift the moment either is edited — leaving two statements about where a user's data goes, disagreeing.

Lift that selection to somewhere both surfaces read from, so there is one answer to "what do we tell the user about this destination".

The own-server case is the one to be careful with: its paragraph already depends on whether the address is encrypted, so what is lifted is a decision, not a lookup table.

## Acceptance criteria

- [ ] The settings dialog renders exactly what it renders today, for all four providers, including the encrypted and plaintext own-server variants
- [ ] The per-provider selection is reachable from outside that dialog
- [ ] A test pins each provider to its own paragraph, so a future edit cannot silently collapse two of them together
- [ ] No new strings, no ARB changes

## Blocked by

- None — can start immediately.

simonoppowa:
Fixed in #837, merged. The per-provider selection moved beside `aiAssistSubtitle`, which is already shared for the same reason, and the dialog's private copy was deleted rather than left alongside it. A test pins each provider to its own paragraph so two cannot silently collapse into one.

PR #840, opened 2026-08-25
fix: close the review findings on the AI consent flow

simonoppowa:
Addresses the review of #838 / #839, both of which merged while these fixes were being written — so they land here, on #837's branch, which now carries all three tickets.

Four of the five findings accepted, one rejected with evidence.

## Accepted

**The screen promised something the code does not do.** `_selectProvider` persists the provider and *then* repaints the disclosure, so "each provider's terms are shown **before you switch**" was false. Nothing is sent by switching, so the sentence now states the guarantee that is true: the terms are in Settings, and nothing leaves until the feature is used.

**The own-server paragraphs never mentioned photos.** The camera is offered for a server the user runs once its photo probe passes, yet both variants listed only the typed text — understating what leaves **in the settings dialog too**, not just on the agreement screen. Fixed in all nine locales, with a test asserting every provider's paragraph mentions the photo, so a fifth destination cannot be added without one.

**The profile was not named** among the things that never leave. It is now, beside the diary.

**`enabled` did not consult the agreement.** It asked whether the provider was configured and whether the pause switch was off, nothing else. Not reachable through the dialog, which asks before it writes — but "the UI never does that" is weaker than what #836 promises, and this is the layer that owns the state. It now sits in the expression that already refuses to call an unusable provider on.

**Also taken:** the duplicated usability question is now one helper. Worth saying why it mattered rather than treating it as tidiness — that duplication is exactly how `setEnabled` and `clear` came to disagree about what counts as configured, with the narrow key-only test surviving in one long after the other had learned better.

## Rejected

**A Bloc for the consent screen.** `AGENTS.md` says every screen has one, but practice scopes that to stateful feature screens: seven Scaffold-bearing screens have no Bloc, three of them in the same directory as this one — `food_sources_screen`, `kcal_goal_info_screen`, `nutrient_visibility_screen`. A consent screen holds no state; the answer is the pop value. A Bloc here would be an empty state and one event that pops.

## What it cost

Six test fakes now seed the agreement alongside the credential they seed, because a store holding a key and no agreement is a state the app cannot reach — those arrangements described a world that does not exist. That churn is the honest price of putting the invariant in storage rather than leaving it to UI flow.

Two existing assertions had to say what they meant rather than what happened to be true: one listed the exact stored tags where it meant "reading writes nothing back", and the summary's read budget went from five to six, because the agreement is genuinely one more key. The budget still guards what it was written to guard — that reads do not grow per value.

## Verification

| | |
|---|---|
| full suite | **1790 pass** |
| `flutter analyze` (bare, whole project) | **clean** |

### Mutation check

| reverted | outcome |
|---|---|
| own-server forgets photos again | **caught** — *ownServer does not mention the photo it sends* |
| `enabled` stops consulting the agreement | **caught** — *a credential without an agreement is not usable* |


simonoppowa:
Superseded by #841.

This one never got CI: it was opened against `feat/ai-disclosure-one-source`, which merged but was not deleted, so GitHub did not retarget it. Repointing the base fires `pull_request: edited` and reopening fires `reopened` — neither created a workflow run here, and only a `synchronize` does.

Rebuilding the same two commits on the current `feature/ai-assisted-meal-logging` tip avoided a twelve-file merge as well: the feature branch now carries #834–#836 squashed, so merging the base in had both sides adding that content independently, including nine ARBs where one side held the corrected strings and the other the pre-fix ones.

Issue #894, opened 2026-08-27
Is "anonymous crash reports" true given a per-install id?


simonoppowa:
Part of #867

## Question

Is "anonymous crash reports" true, given a stable per-install identifier and a city?

[#889](https://github.com/simonoppowa/OpenNutriTracker/issues/889) measured a live event in this project's Sentry org and found, alongside the crash itself:

- **`user.id`** — a 36-character UUID, set by the native SDKs, **ungated by `sendDefaultPii`**, and carried onto Dart events by `load_contexts_integration.dart`.
- **`user.geo`** — `country_code`, `city`, `region`, resolved by Relay from the connection IP **even though `user.ip_address` is `null`**, and not covered by the org's *Prevent Storing of IP Addresses* setting.

The consent label ([`intl_en.arb`](lib/l10n/intl_en.arb), `dataCollectionLabel`) reads:

> Send **anonymous** crash reports to help fix bugs. No food log, weight, or personal data is included.

[#874](https://github.com/simonoppowa/OpenNutriTracker/issues/874) made that label the basis of an **Art.6(1)(a) consent**, on the reasoning that it was freely given, specific, informed and withdrawable. *Informed* is the element under pressure here.

This is the same shape as the breadcrumb leak ([#877](https://github.com/simonoppowa/OpenNutriTracker/issues/877)) — a true-sounding promise that measurement contradicts — with two differences that cut in opposite directions: it is **weaker**, because a UUID and a city are not diary content; and **harder to fix**, because the identifier comes from the SDKs rather than from this project's own logging.

Decide:

- Is a **stable identifier bound to a city** anonymous in any sense the label can carry? Note it persists across crashes, so events from one installation are linkable to each other.
- If not: does the **label change**, does the **behaviour change**, or both? A `beforeSend` hook could strip `user.id` and `user.geo`; whether that costs anything Sentry is actually used for is part of the question.
- Does this **reopen #874's consent conclusion** or #887's decision not to re-consent? Both were reasoned on the label as written.
- Does the **README** sentence need a third correction? [PR #891](https://github.com/simonoppowa/OpenNutriTracker/pull/891) fixed *"no user or device identifier"* for the Supabase path only, and a persistent UUID is an identifier in the plainest sense rather than something a recipient derives.

The [spec](docs/privacy-policy-spec.md) is written and lists this under **Left open**; it does not block applying the rest.

simonoppowa:
[...]

Issue #892, opened 2026-08-27
Delete all my data leaves the AI API key behind


simonoppowa:
**Settings → Delete all my data does not remove the user's AI API keys.** They stay in the platform secure store after the wipe.

`DeleteAllUserDataUsecase.deleteAll()` ([`lib/core/domain/usecase/delete_all_user_data_usecase.dart`](lib/core/domain/usecase/delete_all_user_data_usecase.dart)) closes Sentry and clears eight Hive boxes for the active profile. It never touches `AiCredentialStorage`, whose `clear()` has exactly one caller — [`ai_assist_dialog.dart:249`](lib/features/settings/presentation/widgets/ai_assist_dialog.dart:249), scoped to a single provider and reachable only by going into AI settings deliberately.

So a user who wipes their data still has a paid credential — and, for the user-run provider, the endpoint URL — sitting in the Keychain/Keystore, with no indication anything was left behind.

This looks like an oversight rather than a decision. Map [#732](https://github.com/simonoppowa/OpenNutriTracker/issues/732) reasoned the endpoint URL should live in the keystore *beside* the key partly because splitting them "would diverge on *delete all my data*" — which assumes delete-all reaches them both. It does not reach either.

## The work

- `deleteAll()` clears the credential store for **all** providers, not one.
- Decide deliberately whether it also clears the **stored model and endpoint** — [#737](https://github.com/simonoppowa/OpenNutriTracker/issues/737) and [#738](https://github.com/simonoppowa/OpenNutriTracker/issues/738) already established that changing the endpoint clears the model, so the invariants exist; this is about whether a wipe returns the app to a genuinely un-configured state.
- Keep the deliberate exclusions intact — other profiles, the shared Open Food Facts cache, and the shared content libraries are excluded on purpose, each with a comment saying why. This is not a licence to widen the wipe.
- A test pinning it, in the spirit of the other invariant tests in `test/unit_test/`.

**Gate: before the AI feature ships.** Nothing is released yet, so this is a release-blocker rather than a hotfix. Raised by [#885](https://github.com/simonoppowa/OpenNutriTracker/issues/885) on map [#867](https://github.com/simonoppowa/OpenNutriTracker/issues/867), which decided the privacy policy will describe the erasure path — and the description has to be true when it lands.

simonoppowa:
Fixed by #908, merged into `feature/ai-assisted-meal-logging` as `6475798f`.

Verified against all four points this issue asked for:

[...]

Issue #876, opened 2026-08-27
How does a policy absorb transfers to an address we cannot name?


simonoppowa:
Part of #867

## Question

How does a generated policy absorb optional, user-keyed transfers — including one to an address the project cannot know?

The baseline decision is that corrections describe **current** processing, so none of the AI providers is disclosed yet. But the corrected policy should take them when they ship without being rewritten, and the hardest case is already specified: map [#732](https://github.com/simonoppowa/OpenNutriTracker/issues/732) settled a fourth provider that is **a server the user runs**, at a URL the app is told at runtime and the project never sees.

- Can a recipient be disclosed as **conditional** — *"if you supply your own key and enable this, your text or photo goes to the vendor you chose"* — or does the generator only express unconditional collection? If it cannot, the policy has to name recipients that most users never reach, which is its own kind of untrue.
- For the **user-run endpoint**, is the project a controller of a transfer whose destination the user chose and the project cannot identify? #732 decided the truthful public claim is *"no third party receives it"*, never *"it never leaves the device"* — the policy must not weaken that into either overclaim.
- Does the **OpenRouter broker path** need disclosing as two recipients rather than one, given the README already says a request lands at OpenRouter *and* the serving vendor, with an account-level identity forwarded that cannot be switched off?
- Does the **user's own API key** — held in the Keystore/Keychain, sent only as the header it authenticates — count as personal data the app processes, or is it out of the policy's reach entirely?
- Is a **placeholder shape** the right answer at all, or should the spec instead state the condition under which the policy must be revisited before the feature ships?

Answer the shape, not the wording. Blocked by the adjudication and the expressiveness research.

simonoppowa:
**The user is the controller of these transfers, so the policy owes an explanation rather than a recipient disclosure — and the constraint that looked hardest stops applying.**

## The user controls this path

They obtain a key from a vendor they contract with and pay, enter it, enable the feature, choose the provider and the model, and on the fourth provider supply the endpoint address as well. You ship software that runs on their device, sees nothing, and has no relationship with the recipient — much as the author of an email client is not the controller of mail sent through the user's own account.

This is a genuinely different posture from Open Food Facts, and the difference is agency, not convenience. There you chose the destination and the transfer happens for everyone by default; here nothing leaves until the user has done four deliberate things, one of which is naming an address you never see.

[...]

Issue #867, opened 2026-08-27
Map: the published Data policy against what the app actually does


simonoppowa:
## Destination

A **locked spec** for correcting both iubenda privacy policies — [EN 53501884](https://www.iubenda.com/privacy-policy/53501884) and [DE 53922100](https://www.iubenda.com/privacy-policy/53922100) — so that they truthfully and GDPR-Art.13-sufficiently describe what OpenNutriTracker **v2.0.2** actually does, including **which document a user is served**, and are shaped to absorb the unreleased AI providers without a second rewrite. Every decision settled and written down, so one session can apply it in iubenda and one PR can fix the code. Applying it is not part of this map.

## Notes

**Domain.** Flutter/Dart, Clean Architecture; repo conventions in `AGENTS.md`. Sessions should reach for `/grilling` and `/domain-modeling`; research tickets are resolved by a research subagent.

**The baseline is v2.0.2, released 2026-08-02 — and the AI feature does not exist in it.** Anthropic, OpenAI, OpenRouter and the user-run endpoint all live on unreleased feature branches; so does the README's Privacy section itself (`main`'s README has no destinations table). So the corrections describe **current** processing — a policy claiming transfers no user can trigger is its own kind of untrue — while the *shape* must take the AI providers when they ship. Every finding carries a **live** or **pending** tag.

**Amendment, made while resolving [#876](https://github.com/simonoppowa/OpenNutriTracker/issues/876):** the AI passage goes into the policy **now**, ahead of the feature shipping, rather than waiting. This departs from the baseline above and was taken knowingly. It holds up because the passage explains a **conditional capability and who is responsible for it** rather than asserting that processing occurs, and because it aligns the policy with the README that already documents the feature. The cost is stated rather than smoothed over: a reader on `v2.0.2` will find a description of something their app does not have.

**iubenda is a fixed constraint.** No answer counts as settled until it is expressible in iubenda: a catalogue entry where one exists, a custom service or custom clause where it does not. "What should it say" must always also answer "and where does that go".

**GDPR/EU is the binding framework.** Controller: Simon Oppowa, Regensburg. Other regional sections are left as iubenda renders them and get no tickets of their own.

[...]

PR #895, opened 2026-08-27
docs: the spec for correcting the published Data policy

simonoppowa:
The published privacy policy described an app that does not exist. This is the spec for making it describe the one that does.

Closes the audit opened as #867. Twenty tickets established what the two iubenda documents get wrong and what they should say instead; this document is the executable form of those decisions, with the ticket carrying the argument cited beside each one.

## What the audit found

The two policies named five processors — TestFlight, Google Play Beta Testing, Google Play Store, App Store Connect, Sentry. Every user of the shipped app also talks to **Open Food Facts** and to **the project's own Supabase backend**, and neither appeared anywhere.

It got more specific from there:

- **Storage permission and Reminders permission were declared and are false.** From a real merged release manifest: no storage permission exists at all, and "Reminders" names an OS API this app has never touched. Meanwhile `CAMERA`, `VIBRATE` and `ACCESS_NETWORK_STATE` are in the build and in no policy.
- **The German document was unreachable.** Both call sites opened the English URL, so a German user acknowledged a document in a language they may not read.
- **Crash reports were carrying the diary** — food searches, water and weight entries, scanned barcodes — against a consent label promising none of it. Live in `v2.0.2`. Fixed in #879.
- **Every food search is logged beside a postal code and a TLS fingerprint** for 24 hours, in a log this project controls.
- **Sentry stores a stable per-install UUID and the city of every crash**, neither gated by `sendDefaultPii`, against a label calling the reports *anonymous*.
- **The README was wrong twice**, in the shipped copy, about what requests actually carry. Corrected in #891 — and the map retired the rule that had made the README the document the policy defers to, since deferring to it would have laundered those errors into the legally-operative text.

## What the spec contains

Structured to be applied rather than read:

| § | |
| :-- | :-- |
| 1–2 | The iubenda service list and the category changes — including that all four permission categories come from a single custom service, so one deletion does it |
| 3 | What each recipient clause must state: role, payload, country, retention |
| 4 | AI assistance as explanatory prose rather than a recipient entry, because the user is the controller of those transfers |
| 5–6 | Legal basis per purpose, and why the onboarding checkbox is an acknowledgement that nothing rests on |
| 7–8 | Local storage, and the owner block |
| 9 | The five repo and app changes that are part of the correction |
| 10–11 | Sequencing, and a verification checklist |

## What it deliberately does not do

**It states no fresh judgement.** Anything still undecided sits under *Left open* — including which iubenda catalogue entries exist for the three added categories, and whether the crash-reporting label may keep the word "anonymous" (#894).

**It does not decide whether an Impressum is owed.** That is a separate obligation under German telemedia law concerning the maintainer's own exposure, and it wants qualified advice rather than a reading. The spec notes only that if one is owed, the postal address belongs there once rather than duplicated into two privacy policies that never required it.

**Two sentences are blocked on work outside it.** The erasure clause needs #892 — today "delete all my data" leaves the AI API keys in the Keychain — and the AI passage assumes the shape #876 settled.

## Reviewing it

The claims are all traceable: each decision links its ticket, and the supporting research is in `docs/privacy-*.md`. The most load-bearing findings were measured rather than reasoned — the merged manifest, the gateway log, a live Sentry event — and those measurements are quoted in the tickets.


simonoppowa:
Closing — this targeted `main`, and `CONTRIBUTING.md:14` says plainly that **all pull requests must target `develop`**, with `main` reserved for a single batched release merge. I reasoned from the default branch instead of reading the contributing guide.

Reopened against `develop` as https://github.com/simonoppowa/OpenNutriTracker/pull/898 with the same two commits, cut cleanly from `develop`. Retargeting this one in place was not an option: the branch was based on `main`, force-push is blocked by a repository rule, and `main` carries five commits `develop` does not — so the diff came out at 130 files.

Issue #890, opened 2026-08-27
Is Cloudflare a named sub-processor under a Supabase DPA?


simonoppowa:
Part of #867

## Question

Is Cloudflare a named sub-processor under a DPA covering this Supabase project?

[#869](https://github.com/simonoppowa/OpenNutriTracker/issues/869) measured Cloudflare's presence in the request path directly: `cf_*` fields on every gateway record with nothing configured on the project, `server: cloudflare` on the wire, and an edge PoP chosen by proximity to the user. [#873](https://github.com/simonoppowa/OpenNutriTracker/issues/873) then adjudicated Supabase a **processor** with Cloudflare beneath it and the project as controller — which makes Cloudflare a sub-processor, and sub-processors need a paper trail.

That trail has not been found. Supabase's public sub-processor page returns **404** at the obvious URL.

Establish:

- Where Supabase actually publishes its **sub-processor list**, and whether Cloudflare appears on it with a stated purpose and location.
- Whether a **DPA** is in force for this account — Supabase's standard terms may incorporate one automatically, or it may need signing from the dashboard's legal or documents section.
- Whether the DPA's **Art.28(2)/(4) sub-processor terms** cover changes to that list, and how notice is given.
- Whether Supabase's terms say anything about the **Cloudflare layer specifically**, given #869 found it is not optional or configurable — the project cannot avoid it while using the managed platform.

Work from Supabase's own published terms and documentation. Do not sign anything; if a DPA needs accepting, say so and stop.

simonoppowa:
**Yes — Cloudflare, Inc is a named sub-processor, and the DPA is already in force. Nothing needs signing.**

The 404 was a wrong URL, not a missing document. The live list is at **https://supabase.com/legal/customer-resources/subprocessor-list** (HTTP 200) — which is the exact URL the DPA itself names as operative in clause 6.2. `https://supabase.com/legal/subprocessors` genuinely 404s; it was never the right path.

### 1. The list

The page wraps a PDF: [Subprocessor List — Updated June 1, 2026](https://supabase.com/legal/subprocessor-list/June-1-2026.pdf). Cloudflare appears verbatim as:

> | Cloudflare, Inc | Provision of hosting services |

**Purpose stated: "Provision of hosting services". Location stated: none.** The PDF has exactly two columns — *Name of Sub-processor* and *Description of Processing*. No country column exists, for Cloudflare or for any of the 23 other entries. So Supabase publishes **who**, never **where**.

[...]

Issue #889, opened 2026-08-27
Does Sentry attach the client IP on iOS despite sendDefaultPii?


simonoppowa:
Part of #867

## Question

Does Sentry set the client IP to `{{auto}}` on Apple platforms regardless of `sendDefaultPii`?

[#870](https://github.com/simonoppowa/OpenNutriTracker/issues/870) verified from the SDK source that `sentry_client.dart` reads `user?.ipAddress ?? (_options.sendDefaultPii ? '{{auto}}' : null)`, and that Relay only infers `geo` when `ip_address` is present — so on the face of it `sendDefaultPii = false` withholds the IP.

But Sentry's **Apple** documentation says the flag *"won't stop Sentry from collecting users' IP addresses via the client connection"*, and #870 could not resolve whether that reaches the Flutter SDK, which layers a Dart client over the native Cocoa SDK. If the native layer attaches it independently, the Dart-side reasoning does not hold on iOS.

This matters concretely now: [#878](https://github.com/simonoppowa/OpenNutriTracker/issues/878) confirmed the org is in the **US region**, so an IP attached on iOS is a third-country transfer of an identifier the policy would otherwise say is not sent.

Measure it rather than reason about it. There is precedent — [#748](https://github.com/simonoppowa/OpenNutriTracker/issues/748) settled an equivalent iOS question by adding a check to the `ios-integration-tests` job rather than waiting for a device. Options: inspect the event Sentry actually receives from an iOS build, or read the sentry-cocoa source for where `ip_address` is set and whether the Flutter SDK's options propagate to it.

Answer with what iOS actually sends, and whether Android differs.

simonoppowa:
**No. Neither iOS event path attaches the IP — and the Apple documentation that raised this question is stale.** `sendDefaultPii = false` holds on iOS, though not for the reason #870 gave. Android was never at risk. Two things the ticket did not ask turned out to matter more; see the end.

## The decisive evidence

**The flag does reach the native SDK.** `autoInitializeNativeSdk` defaults to `true` (`sentry_flutter_options.dart:33`), and `sendDefaultPii` is one of the keys in the `initNativeSdk` options map — `'sendDefaultPii': options.sendDefaultPii,` (`sentry_native_channel.dart:60`) — read straight onto the native options in `SentryFlutter.swift:46`. So the layering worry has no basis.

[...]

Issue #877, opened 2026-08-27
Is the search-term breadcrumb leak fixed in code or disclosed?


simonoppowa:
Part of #867

## Question

The app leaks food-search terms into Sentry as print breadcrumbs. Is that fixed in code, or disclosed in the policy?

Established by [What does Sentry receive at 100% trace sampling?](https://github.com/simonoppowa/OpenNutriTracker/issues/870) and verified line by line:

- `DebugPrintIntegration` is added unconditionally by `SentryFlutter.init` and returns early **only in debug mode** — so it is live in exactly the builds where Sentry runs.
- `enablePrintBreadcrumbs` defaults to `true` (`sentry_options.dart:332`) and the app overrides nothing: it sets only `dsn` and `tracesSampleRate`.
- `LoggerConfig` sets `Logger.root.level = Level.ALL` and pipes every record into `debugPrint`.
- [`off_data_source.dart`](lib/features/add_meal/data/data_sources/off_data_source.dart) logs `'Fetching OFF results from: $searchUrl'` — and the search URL contains the term the user typed.

So for a user who has opted into crash reporting, **their typed food searches ride along as breadcrumbs on the next event of any kind**. The in-app consent string says the opposite: `dataCollectionLabel` — *"Send anonymous crash reports to help fix bugs. No food log, weight, or personal data is included."*

This map's destination is a policy spec, so the decision it owes is which way the contradiction resolves:

- **Fix in code** — stop logging full URLs, or set `enablePrintBreadcrumbs = false`, or both. The consent string then becomes true as written and the policy says nothing new. Cheapest, and it is the option that keeps a promise already made rather than retracting it.
- **Disclose** — leave the behaviour and widen the policy and the consent string to admit that search terms may accompany a crash report. Honest, but it weakens a claim the project currently makes.
- **Both, in order** — disclose for the shipped version, fix going forward.

Weigh at least: whether a food search term is personal data at all (which is [the adjudication ticket](https://github.com/simonoppowa/OpenNutriTracker/issues/873)'s business, though the consent string's promise is broken on the plainer reading either way); whether breadcrumbs carry anything else worth catching, since `Level.ALL` means *every* log line, not just this one; and whether the fix belongs in the Sentry options or in the logging call sites, since the two differ in what else they suppress.

Not blocked: the code fix is available now regardless of how the personal-data question lands.

simonoppowa:
[...]

Issue #872, opened 2026-08-27
Which permissions does the released build actually request?


simonoppowa:
Part of #867

## Question

Which permissions does the released build actually request, including the ones it acquires transitively?

Both policies list **Storage permission** and **Reminders permission** among the personal data collected. The manifest on `main` declares only `INTERNET`, `RECEIVE_BOOT_COMPLETED` and `POST_NOTIFICATIONS`; iOS declares `NSCameraUsageDescription` and `NSPhotoLibraryUsageDescription` and nothing about Reminders. On the face of it the policy **overclaims** — it tells users more is taken than is taken — which is a defect in its own right even though it errs toward disclosure.

Before deleting anything, establish the real set at `v2.0.2`:

- the **merged** Android manifest, not the app's own — plugins inject permissions, and `image_picker`, the file export path, or a notifications plugin may add storage or alarm entries that never appear in `android/app/src/main/AndroidManifest.xml`;
- the iOS **usage-description keys** actually present in the built app, and whether any plugin adds `NSRemindersUsageDescription`;
- whether *Reminders* in the policy plausibly refers to the **local-notification** permission rather than the iOS Reminders app, in which case the entry is misnamed rather than false;
- whether the export/import path touches the **Storage Access Framework**, which needs no declared permission and so would make *Storage* wrong in a third way.

Record the true set with evidence for each. The removal or renaming decision follows in the rendering pass.

simonoppowa:
## Verdict

The policy **overclaims on both entries, in different ways**.

- **"Storage permission" is false in every sense — remove it.** There is no storage permission in the released build: none declared, none injected by any plugin, none `maxSdkVersion`-gated, and `permission_handler` isn't even a dependency. Export/import goes through SAF document pickers, which by design need none.
- **"Reminders permission" is a misnaming, not a fabrication — rename it.** The app's own UI calls the feature "Daily Reminder"; it's delivered by `flutter_local_notifications` over `POST_NOTIFICATIONS`. No Reminders/EventKit API is touched on either platform. Deleting it outright would swing from overclaiming to underclaiming a permission the app does prompt for at onboarding.

And the same audit turned up the opposite defect: **`CAMERA`, `VIBRATE` and `ACCESS_NETWORK_STATE` are all in the released build and appear in neither policy.**

## Route taken

[...]

PR #897, opened 2026-08-27
docs: the research behind the privacy-policy spec

simonoppowa:
The six research documents behind the privacy-policy spec (#895), brought onto `main` so its citations resolve.

Each arrived as its own branch and is cherry-picked here unchanged, so the individual commits and their reasoning survive. One annotation was added on top — see below.

| Document | Question it answers | Ticket |
| :-- | :-- | :-- |
| `privacy-fdc-reachability.md` | Does the app still call the USDA FoodData Central API? (No — the client is wired but has zero callers) | #868 |
| `privacy-permission-audit.md` | What does the released build actually request? (Six permissions, from a real merged release manifest) | #872 |
| `privacy-sentry-payload.md` | What does Sentry receive at `tracesSampleRate = 1.0`? | #870 |
| `privacy-iubenda-expressiveness.md` | What can the generator be made to say? | #871 |
| `privacy-supabase-subprocessors.md` | Is Cloudflare a named sub-processor under a DPA? | #890 |
| `privacy-sentry-ios-ip.md` | Does iOS send the client IP despite `sendDefaultPii`? | #889 |

## Why these are worth keeping

The load-bearing findings in the spec were **measured rather than reasoned**, and these documents are where the measurements live: the merged release manifest that showed no storage permission exists, the gateway log record pairing a search term with a postal code and a TLS fingerprint, the sub-processor PDF, the live Sentry event. Each claim carries a quote and a `file:line` or URL.

They also record what was *rejected*, which the spec cannot — a spec states conclusions, and half the value of an audit is knowing which plausible answer was wrong and why.

## One document is annotated rather than corrected

`privacy-sentry-payload.md` concluded that Sentry infers geolocation only when `ip_address` is present on the event. **That is wrong.** #889 disproved it by reading a live event in this project's own organisation: `user.geo` carried country, city and region while `user.ip_address` was `null`, because Relay resolves geo from the *connection* IP before the filtering step.

The final commit adds a warning callout at the top of that document instead of editing the claim away. The point of keeping research in the repo is that its reasoning can be checked, and a document silently corrected teaches nobody which step was wrong. Everything else in it stands — including the print-breadcrumb finding that led to #877, which turned out to be a live diary leak in the shipped release.

## Notes

Docs only; no code paths touched. The six source branches remain pushed and can be deleted once this merges.

### DIFF
--- a/docs/privacy-fdc-reachability.md
+++ b/docs/privacy-fdc-reachability.md
@@ -1,0 +1,165 @@
+# Is the USDA FoodData Central path reachable at runtime?
+
+Research for [#868](https://github.com/simonoppowa/OpenNutriTracker/issues/868), on map [#867](https://github.com/simonoppowa/OpenNutriTracker/issues/867). Written 2026-08-27. Baseline: tag `v2.0.2` (commit `9ab14fe3`), which is what `main` is at. All line numbers below are `v2.0.2` unless stated otherwise.
+
+## Short answer
+
+**Not reachable. `api.nal.usda.gov` receives nothing from the released app under any condition — no setting, no flag, no fallback, no hidden screen.** The direct FDC client is instantiated on every launch and injected into `ProductsRepository`, but the one repository method that would use it, `getFDCFoodsByString`, has **zero callers anywhere in the tree** — not in `lib/`, not in tests, not in tooling. The trail breaks at exactly one line: the use case named `searchFDCFoodByString` calls `getSupabaseFoodsByString`, not `getFDCFoodsByString`. The "FDC" naming throughout the search stack is a label for *which corpus the results came from*, not for *which host was contacted*.
+
+The README's claim that a search "reaches that backend and stops there" is **accurate as written**. USDA FoodData Central is **not** a live recipient and does not belong on the recipients list.
+
+## What the direct client would send, if it were reached
+
+Establishing the payload first, because it is what would have to be disclosed if any hop connected.
+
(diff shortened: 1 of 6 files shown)


simonoppowa:
Closing unmerged. These files should not be in the repo, and the reasoning that put them here was weaker than it looked.

**The findings are already durable elsewhere.** Every one of these six investigations was posted as a resolution comment on its ticket — verdict, decisive evidence, quotes with a `file:line` or URL — and [#867](https://github.com/simonoppowa/OpenNutriTracker/issues/867) indexes all twenty. That is the record anyone will actually go looking for, and it is linkable and permanent without adding anything to the tree.

**And these documents are exactly what the audit spent twenty tickets warning about.** 2,336 lines of prose that nothing checks. `privacy-supabase-subprocessors.md` describes a sub-processor list dated 1 June 2026; `privacy-iubenda-expressiveness.md` describes a generator UI that will change. Both will be wrong within a year and nothing will notice — the same failure mode found in the README, in `AGENTS.md`, and in the crash-reporting consent label. [#883](https://github.com/simonoppowa/OpenNutriTracker/issues/883) concluded that measured behaviour beats prose; merging more unchecked prose cuts directly against that.

The pattern showed up before this even merged: one document had to be annotated as partly wrong within hours of being written, when [#889](https://github.com/simonoppowa/OpenNutriTracker/issues/889) disproved [#870](https://github.com/simonoppowa/OpenNutriTracker/issues/870)'s geolocation conclusion.

**The repo had already answered this.** Two earlier maps left their research on throwaway branches, and roughly eighteen `docs/ai-*.md` files from those efforts have never been committed anywhere. That is the established practice and it was right.

**What changed instead:** [#895](https://github.com/simonoppowa/OpenNutriTracker/pull/895) now cites the **tickets** rather than file paths (`2c79b208`), so nothing in the spec depends on these files. The six `research/*` branches stay pushed as a cold archive of the fuller working — the sources that led nowhere, the alternate readings — for anyone who wants to re-verify a measurement rather than trust it.

PR #913, opened 2026-08-27
DO NOT MERGE — throwaway: verify the iOS retry gate

simonoppowa:
Throwaway for #912. **Do not merge; closed once both scenarios report.**

Scenario 1 (this commit): both attempts forced to fail fast. Expected — `ios-integration-tests` green (by design), `ios-integration-tests-retry` runs and is green, `ios-integration-tests-result` **red**. If the gate is green here it is worthless and #912 is wrong.

Scenario 2 (next commit): first forced to fail, second real. Expected — retry fires on a new runner, real suite passes, gate green.

Unrelated heavy jobs are `if: false` here to keep the throwaway cheap.

### DIFF
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -260,129 +260,45 @@ jobs:
         # App Store install and isn't what CI should validate.
         run: flutter build ios --no-codesign --release --flavor full
 
+  # Two calls on two runners, not two attempts on one. See the header of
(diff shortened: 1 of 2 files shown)


simonoppowa:
Both scenarios verified — see #912. Scenario 1: both attempts fail → gate red (`first runner: false`, `second runner: false`). Scenario 2: first fails, retry runs the real suite on its own runner and passes → gate green (`first runner: false`, `second runner: true`).

Issue #900, opened 2026-08-27
Keep the per-install id and city out of crash reports


simonoppowa:
Crash reports carry a **stable per-install identifier** and the **city** of every crash. Neither is needed to fix a bug, and both contradict what users were told.

Measured on a live event in this project's Sentry organisation ([#889](https://github.com/simonoppowa/OpenNutriTracker/issues/889)):

```
user.id  : 36-character UUID   ← stable across launches, ungated by sendDefaultPii
user.geo : { country_code, city, region }
user.ip_address : null          ← the IP itself is correctly withheld
```

The consent label promises *"**anonymous** crash reports … No food log, weight, or personal data is included"*, and [#874](https://github.com/simonoppowa/OpenNutriTracker/issues/874) made that label the basis of an Art.6(1)(a) consent. [#894](https://github.com/simonoppowa/OpenNutriTracker/issues/894) decided the data is **pseudonymous, not anonymous** — a persistent identifier makes every crash from one installation linkable to every other — and that the fix is to remove the identifier rather than to describe it.

## The work

**1. Strip the identifier, client-side.** Add a `beforeSend` hook in `lib/core/utils/sentry_config.dart` that clears `event.user`. That also restores an existing public claim instead of requiring a third correction of it: the README's *"no user or device identifier"* becomes true for Sentry again. Pin it with a test alongside the existing `sentry_config_test.dart` cases, which already pin `enablePrintBreadcrumbs` and `sendDefaultPii`.

**Known cost, accepted deliberately:** `user.id` is what powers Sentry's *users affected* count. Without it you cannot tell one installation crash-looping from a hundred installations hitting the same bug.

**2. Chase `user.geo`, server-side.** A client hook **cannot** remove it — the client never sends it. Relay derives it from the connection IP at ingest, and #889 established the organisation's *Prevent Storing of IP Addresses* setting does not cover it. The candidate is an **Advanced Data Scrubbing** rule under Security & Privacy. **Verify against a real event rather than assuming it worked** — the same method that found this in the first place.

**3. Correct the consent label.** Drop *"anonymous"*. Say what is actually sent and what is not.

## Sequencing

**Blocked on [#893](https://github.com/simonoppowa/OpenNutriTracker/pull/893)**, which is the PR that brings `sentry_config.dart` onto `develop`. Editing that file before it lands there would collide.

[...]

PR #917, opened 2026-08-27
fix(core): send crash reports, not a report every time the app opens

simonoppowa:
`tracesSampleRate` was `1.0`.

A traces sample rate is what turns transaction sending on — `isTracingEnabled()` is `tracesSampleRate != null || tracesSampler != null` — and `enableAutoPerformanceTracing` defaults to `true`. So app-start and screen-load transactions were produced for **ordinary sessions, at a 100% sample rate, with no crash involved**.

## Why that is a problem rather than a preference

The switch the user agreed to says:

> Send crash reports to help fix bugs. No food log or weight data is included.

— `dataCollectionLabel`, `intl_en.arb:335`

A report every time the app opens is not that. The published Data policy scopes it the same narrow way: *"When **a crash** sends a technical report […]"*.

There is a second consequence that is easy to miss. The policy concedes that *"Sentry's servers may derive an approximate location (country, region and city) from the connection"*. At `1.0` that concession applies **once per session** rather than once per crash — a materially different picture of how often a coarse location is observed, and it cannot be fixed client-side, because Relay derives it server-side at ingest after `beforeSend` has run (see the doc comment on `_stripUser`, and #900).

## The change

One line, plus the reasoning next to it. `null` is what switches performance monitoring off, so no transaction is ever sampled or sent.

Written as an **explicit null rather than a deleted line**, matching `enablePrintBreadcrumbs` directly above it: this file exists so that "the one setting which must not drift can be pinned by a test rather than by a reviewer noticing an absent line", and an absent setting reads as an oversight where this one is a decision.

`sentry_config_test.dart` gains a case asserting both the null rate and `isTracingEnabled() == false`.

## The alternative, deliberately not taken

Keep the tracing and widen the consent string and both policy documents to "crash **and performance** reports". That stays available and nothing here is hard to reverse — **say so and I will do it instead**. It was not the default choice because it asks the user for more than they were asked for, and nothing in this repository reads the performance data.

## Scope note

This targets `develop`, per `CONTRIBUTING.md:14`. Worth knowing that `origin/main` also carries `tracesSampleRate = 1.0` (it is three commits past `v2.0.2` and has `sentry_config.dart` with `enablePrintBreadcrumbs = false` but no `beforeSend`), so whichever branch the next release is cut from, it currently ships 100% tracing.

Found while auditing the published Data policy against the source. The policy half of this finding — the sentence that says only a crash sends anything — is being drafted separately.

### DIFF
--- a/lib/core/utils/sentry_config.dart
+++ b/lib/core/utils/sentry_config.dart
@@ -21,10 +21,34 @@ import 'package:sentry_flutter/sentry_flutter.dart';
 /// line can leave the device.
 void configureSentryOptions(SentryOptions options, {required String dsn}) {
   options.dsn = dsn;
-  options.tracesSampleRate = 1.0;
+  options.tracesSampleRate = _tracesSampleRate;
   options.enablePrintBreadcrumbs = false;
   options.beforeSend = _stripUser;
 }
+
+/// Null, which is what switches performance monitoring off: `isTracingEnabled`
+/// is `tracesSampleRate != null || tracesSampler != null`, so a null rate means
+/// no transaction is ever sampled or sent.
+///
+/// It was `1.0`. A traces sample rate is what turns transaction sending on, and
+/// `enableAutoPerformanceTracing` defaults to true, so app-start and
+/// screen-load transactions were produced for ordinary sessions at a 100%
+/// sample rate — with no crash involved.
+///
+/// The switch the user actually agreed to says *"Send crash reports to help fix
+/// bugs"*. A report every time the app opens is not that. It also changes how
+/// often Sentry observes the coarse location it derives server-side from the
+/// connection: once per session rather than once per crash, which is a
+/// materially different picture from the one the consent describes.
+///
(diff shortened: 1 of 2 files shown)


simonoppowa:
Confirmed: tracing stays off, rather than widening the consent string and the policy to "crash and performance reports". This PR is the fix as written; the alternative it offers is declined.

Two consequences recorded elsewhere so they don't get lost:

- **No policy edit is needed for this finding.** Once this ships, the published sentence — *"When **a crash** sends a technical report […]"* — becomes true as published, and the consent string stays exactly as it is. Noted on #919, where the draft replacement text is now marked do-not-apply.
- **#900's scrubbing rule is still needed.** It was going to strip the `user_city` / `user_country_code` span tags as well as the event geo; with no transactions there are no span tags, but the geo on ordinary crash events is what that issue is actually about and is unaffected by this change.

If tracing is ever turned back on, both come back together: the policy sentence needs widening and the span tags reappear.

Issue #927, opened 2026-08-27
The Health Connect rationale intent lands on the home screen


simonoppowa:
Found while preparing the Play **Health apps declaration** for 2.1.0. The manifest declares the Health Connect rationale correctly; the activity it points at does not handle it.

## What is declared

`AndroidManifest.xml` has both halves, and both are right:

- an `<intent-filter>` on `.MainActivity` for `androidx.health.ACTION_SHOW_PERMISSIONS_RATIONALE`
- an `<activity-alias>` `ViewPermissionUsageActivity` targeting `.MainActivity`, for `VIEW_PERMISSION_USAGE` with category `HEALTH_PERMISSIONS`, guarded by `android:permission="android.permission.START_VIEW_PERMISSION_USAGE"`

That matches Google's own snippet, permission guard included.

## Why it does not work

Both point at `MainActivity`, which does not look at the intent that started it. On `develop` the class is a `FlutterFragmentActivity` whose only job is the locale `MethodChannel` — no `onNewIntent`, no intent inspection, no `getIntent()` read anywhere in the file. Nothing on the Dart side reads it either.

So Health Connect launches the app and the user lands on the diary home screen with no explanation of what was asked or why. The requirement is met in the manifest and unmet in the product.

This is the kind of thing a reviewer checks by tapping it.

## What it should do

Route the two actions to a screen that explains the health import — which data is read, what it is used for, that nothing is written back, that it stays on the device — and links the privacy policy's health section. `HealthSyncScreen` is the natural destination, or a purpose-built page if the settings screen carries too much unrelated state to deep-link into.

The two actions can share one destination. `VIEW_PERMISSION_USAGE` is what Health Connect opens from its own settings for an app that already holds permissions; `ACTION_SHOW_PERMISSIONS_RATIONALE` is what it opens from the grant dialog. The same explanation answers both.

Worth checking while in there: `MainActivity` is `launchMode="singleTop"`, so a second launch arrives through `onNewIntent` rather than `onCreate`, and a naive implementation that only reads the intent at startup will work the first time and silently stop working after that.

## Related

Companion to #926 — same submission, different requirement. That one is about what the app says before it asks; this is about what it says when the user goes looking afterwards.

Issue #926, opened 2026-08-27
Show a prominent disclosure before asking for Health Connect permissions


simonoppowa:
Found while preparing the Play **Health apps declaration** for the 2.1.0 release, which is the first to carry workout import (#651). Blocks the submission rather than the build.

## The gap

Play's User Data policy requires an in-app **prominent disclosure** with affirmative consent *before* an app requests access to sensitive data. Health and fitness data is sensitive under that policy, and Health Connect access is reviewed against it.

`health_sync_screen.dart` goes straight to the platform prompt. `_onAutoImportChanged` flips the toggle and calls `HealthImportRepository.requestPermissions()` immediately — the system dialog is the first thing the user sees. The screen does have an info affordance, but it opens the WHO and compendium citations behind the calorie-credit maths, which is a different subject entirely.

Nothing about this is misleading — the import genuinely is read-only, on-device, and opt-in. The problem is that the app never *says* so before asking, and the reviewer checks for the saying.

## What the disclosure has to cover

Per the policy, it must appear before the permission request, be dismissible only by an affirmative action, and state:

- which data is accessed — workouts, and body fat percentage for the calorie-credit suggestion
- what it is used for — filing finished workouts as diary activities and suggesting how much of their energy to credit
- that nothing is written back to the health store
- that the data stays on the device

The last two are already true and already asserted in `NSHealthShareUsageDescription`; this is about saying them on Android, in the app, before the prompt.

## Not to be confused with

The Health Connect *rationale* screen, which is a separate requirement and separately broken — see #927. The rationale answers "why does this app want my data" when the user goes looking in Health Connect. This one is the disclosure the app owes before it ever asks.

## Ordering

Health Connect access is allow-listed per package name and is independent of app version, so workout import is inert until Google approves the declaration. That means this does not block shipping 2.1.0 — but the declaration cannot be submitted until it is done, and Google publishes no review turnaround, so the approval clock does not start until then.

simonoppowa:
Shipped in #929 (merged 2026-08-27).

`lib/features/settings/presentation/widgets/health_disclosure_dialog.dart` is on `develop` and `main`, and the gate this issue asked for is in place — `health_sync_screen.dart:144` runs `if (!await _confirmDisclosure()) return;` before the `requestPermissions()` call at line 149, so the disclosure cannot be skipped.

Closing as delivered. `Fixes #N` does not auto-close against `develop`, which is why this stayed open.

Issue #925, opened 2026-08-27
Revoke the USDA FDC API key that shipped in every release binary


simonoppowa:
Follow-up from #881, which removed `FDC_API_KEY` from the repository. Removing it from the source does not invalidate it — the key itself is still live, and it has been distributed.

## What shipped

`FDC_API_KEY` was injected through `envied` into `lib/core/utils/env.dart` and compiled into every release build. `envied` obfuscates rather than encrypts, which it documents plainly; the value is recoverable from a shipped artifact by anyone who cares to.

Checking the tags, it is present in `lib/core/utils/env.dart` in **21 releases** — every tag from `beta` through `v2.0.2`. The removal is on `develop` and reaches users only with 2.1.0. `.github/workflows/default_workflow.yml` on `main` still carries five references to the secret.

The Actions secret `FDC_API_KEY` is dated 2023-07-24 and has never been rotated.

As #881 established, the client this key existed for — `FDCDataSource` — had no callers. The app has never made a request to `api.nal.usda.gov`, so there is no traffic to migrate and nothing breaks when the key stops working.

## What did not happen

**The key is not in git history.** Every historical value of `FDC_API_KEY` in `.env` and `.env.example`, across all six commits that touched those files, is the same 8-character placeholder. `.env` was renamed to `.env.example` in 2a4db305 as a precaution, but there was never a real value there to leak. A scan of the wider history for a key-shaped value next to `FDC_API_KEY` also came back empty.

So the exposure is confined to the compiled binaries and to the Actions secret — not to anything readable in this repository.

## Severity

Low, but not zero. An api.data.gov key gates access to a free public dataset and carries a rate limit rather than a spend. The realistic downside is somebody else consuming this project's quota, or the key being used in a way that gets it throttled or blocked upstream — which would be attributed to this project.

It is a credential that was published, was never used, and is still valid. That is enough to retire it.

## What needs doing

1. Revoke or regenerate the key at api.data.gov, under the account it was issued to. FDC keys are issued and managed by api.data.gov rather than by the FDC site itself.
2. Delete the `FDC_API_KEY` Actions secret once 2.1.0 has shipped and no workflow on `main` references it. Deleting it while `main` still injects it in five places would break the release workflow.
3. No client change is needed — #881 already removed every consumer.

## Ordering

[...]

PR #934, opened 2026-08-28
Release 2.1.0: workout import, crash-report and privacy-policy corrections, food search over POST

### LINKED ISSUE #764: [Bug]: Delete all data leaves the daily reminder scheduled
## Description

Deleting all user data resets the app to onboarding but leaves the previously scheduled daily reminder registered with Android. The user can therefore still receive a nutrition reminder after explicitly deleting their profile and configuration.

## Steps to reproduce

1. Complete onboarding and enable **Daily reminder**.
2. Verify the reminder is scheduled (the test used the default 08:00 reminder).
3. Open **You → Settings → Delete all my data**.
4. Confirm **Delete everything**.
5. Observe that the app returns to onboarding.
6. Inspect Android's scheduled alarms, or wait for the reminder time.

## Expected behaviour

Delete-all clears the user's configuration and cancels every scheduled OpenNutriTracker notification/alarm before returning to onboarding.

## Actual behaviour

The app returns to onboarding, but Android still lists:

```text
com.opennutritracker.ont.opennutritracker/
com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver
```

for the old 08:00 daily reminder. Uninstalling the disposable package removed the alarm.

## Platform

- Android

## OS version

Android 17 / API 37 on Pixel 6 (`oriole`)

## App version

2.0.2, built from source at `145fca91764c872544f91bb0e317b979ef946417` (PR #752)

## Install source

Built from source

## Feature area

Settings / notifications / onboarding reset

## Additional context

Found during the full Pixel regression for PR #752. The delete/reset path was not changed by that dependency rollup.

## Checklist

### LINKED ISSUE #882: Keep food search terms out of the Supabase gateway log
Every food search a user runs is recorded in the Supabase API gateway log next to their IP address, city, postal code, ISP and a TLS fingerprint. Retained one day, EU region, readable only by the maintainer — but the pairing is avoidable and worth avoiding.

## Why it happens

`sp_food_data_source.dart` searches with PostgREST's `.textSearch(...)`, which is a **GET** — so the term lands in the query string:

```
/rest/v1/food_summary?select=*&name=wfts(english).<the user's term>&limit=100
```

The gateway log stores that URL twice over, as `request.url` and `request.search`, in the same record as `request.headers.cf_connecting_ip`, `request.cf.city`, `request.cf.postalCode`, `request.cf.asOrganization` and `request.cf.botManagement.ja3Hash`.

## Why it is fixable

Enumerating the ~50 fields the gateway records turns up **no request-body field**. A PostgREST **RPC call is a POST with a JSON body**, so a search issued as `.rpc('search_food_summary', {...})` against a Postgres function would not put the term in the log at all.

The IP and the geolocation would still be recorded — those come from Cloudflare and are not ours to switch off. But they would no longer sit beside *what the user searched for*, and that pairing is the part worth removing.

## Shape of the work

- A `search_food_summary(term text, sources text[], limit int)` function in the [OpenNutriTracker-Backend](https://github.com/simonoppowa/OpenNutriTracker-Backend) repo, matching the current ranking behaviour (`rankAndTruncateFoodsByName`, the `_candidatePoolSize` over-fetch, the source filter).
- The same for the `food_translation` lookup, which has the identical problem on the localized path.
- Swap the two call sites in `sp_food_data_source.dart`.
- Verify against the gateway log that the term no longer appears.

Not urgent and not a broken promise — the privacy policy names no backend at all today, which is being fixed separately. Found while resolving https://github.com/simonoppowa/OpenNutriTracker/issues/873 on map https://github.com/simonoppowa/OpenNutriTracker/issues/867, which decided to disclose the logging and raise the reduction here rather than fold it into the policy spec.

simonoppowa:
> [!IMPORTANT]
> **The #900 blocker is cleared.** The Sentry `user.geo` scrubbing rule is applied and now *verified against real events* — geo was populated across six countries up to 28 Aug ~06:00–07:30 UTC and has been empty in every event since, with no geo tag and no `city` anywhere in the payload. See the [verification on #900](https://github.com/simonoppowa/OpenNutriTracker/issues/900#issuecomment-5453669362); deployment note 5 below is updated.
>
> #933 and #945 have both landed on `develop` and are included.

## Summary

Release 2.1.0 (build 62). Everything merged into `develop` since 2.0.2 (#604) went to `main`: opt-in workout import from Health Connect and Apple Health, a run of corrections to what the app sends to Sentry and to what it tells users it sends, food search moving off the URL and into a POST body, the first health-data coverage the privacy policy has ever had, and a batch of user-facing bug fixes.

`main` is **87 commits** behind `develop`, but most of that is 2.0.2 content it already holds as the single squashed commit of #604. The real delta is **33 PRs**: #590, #615, #649, #651, #672, #700, #712, #714, #718, #752, #772, #786, #827, #828, #829, #881, #893, #899, #901, #902, #904, #905, #909, #911, #916, #917, #918, #921, #923, #924, #928, #929, #931.

Two of those carry nothing new for `main`. It took three commits directly after the 2.0.2 merge (#879, #880, #891), and nothing brings a direct-to-`main` change back down; #893 and #899 cherry-pick them onto `develop`, so from `main`'s point of view they are already-present content. They matter because the crash-reporting work below builds on them.

## Type of change
- [x] Bug fix
- [x] New feature
- [x] Refactor / cleanup
- [x] Documentation
- [x] CI / tooling
- [x] Localization

## Related issues

Closes #764, #766, #768, #823, #868, #882, #926, #927, and the `develop` half of #883.

Refs #580, #629, #651, #867, #871, #887, #888, #900, #915, #919, #920, #925.

### Workout import from Health Connect / Apple Health (#651, #924, #929)

The headline feature, and opt-in throughout: nothing is read until the user turns it on in Settings → Health sync.

- **#651** adds the import itself — read-only access to workouts and body fat percentage, filed into the diary as ordinary editable activities, crediting a user-tunable share of the reported energy.
- **#924** bounds the tombstone list that stops a deleted workout being re-imported, so it stops growing forever. A read-path bug surfaced while writing the test: pre-release tombstones were invisible until the next write, which would have re-imported workouts the user had already deleted.
- **#929** closes both in-app gaps Google checks for. A **prominent disclosure** now gates the permission request rather than merely preceding it — cancel, a tap outside, a killed activity all read as a refusal and ask the platform for nothing. And the **Health Connect rationale intent** now goes somewhere: the manifest declared it correctly, but `MainActivity` never read the intent that started it, so the user asked "why does this app want my data" and landed on the diary.

### Privacy and crash reporting (#893, #901, #902, #917, #921, #881)

- **#901** stops crash reports carrying a per-install identifier; **#902** stops the setting calling them "anonymous", which they were not.
- **#917** turns Sentry performance tracing off rather than widening the consent string — the deliberate choice between two ways to make the code and the promise agree.
- **#921** shows existing users a one-time notice that the policy changed, recorded on close rather than on open so someone who kills the app mid-dialog is told again.
- **#881** removes the unused direct USDA FDC client and its API key, which shipped in every release binary for a client that never ran.

### Food search moves into a POST body (#911, closes #882)

Search terms travelled in the URL, where the gateway logged them alongside the caller's IP, city and connection fingerprint. Three Postgres functions now take them in the request body instead. **Verified live before this PR:** all three exist in production with the expected signatures and `EXECUTE` granted to `anon` and `authenticated` — which matters, because PostgREST calls as `anon`.

### Fixes

#615 stone weights showing a full stone of pounds · #649 servings the app could not scale · #712 a not-found barcode reported as not found · #714 the German privacy policy opening, and unopenable links reported · #718 navigation stack surviving an absent unwind target · #828 a deleted profile's reminder outliving it · #829 weekday headings clipped at large text sizes.

### Other user-facing changes

#590 suggests a meal type from the current time · #672 exposes the Android per-app language picker.

### Docs and localization

#918 stops eight locales telling users their recipes are deleted — they are not, and English had been right since #471 · #916 corrects the README: a *language* code leaves the device, not a country code, and only on word searches · #899 corrects what the README claims about the wire · #904 and #905 describe the localization workflow as it actually works · #928 stops `policyRevision`'s comment claiming a health section was published before it was · #786 removes the retired GitHub Pages site.

### CI / tooling / deps

- **#923** commits the rendered privacy policy so git is its change record, and checks the two languages have not drifted — iubenda keeps no version history and custom clauses do not propagate between languages.
- **#909** fails the build when a locale is missing a key. `gen-l10n` exits 0 on a missing translation and silently falls back to English; #918 in this same release is that failure caught by hand.
- **#931** stops the project-board automation racing itself when an issue is filed with labels.
- **#827** bounds each integration attempt so a hang can be retried; **#700** fixes the pod fallback to run on the bash macOS runners actually have; **#752** consolidates sixteen Dependabot PRs into one reviewable rollup.

### Version

**#772** bumps to 2.1.0+62 and raises `minSdkVersion` 24 → 26.

## Test plan
- [ ] Described steps below were followed locally
- [ ] Manual check on Android
- [ ] Manual check on iOS

The suite runs **1160 tests, all passing on the merge base**, verified on merged `develop` — up from 1147 before #929. The new l10n completeness gate reports `{}` against #929's three new strings, which is the one combination CI never exercised on its own, since #909's run predated those keys.

Several fixes are mutation-checked against their own tests rather than assumed — #929 records which test fails when each half of it is reverted.

Two gaps are known and not closed here. #651's iOS path has never run on hardware. And #922 asks whether iOS Local Network Privacy gates the own-server AI provider — not this release's feature, but the same class of question as #769 below.

### Steps
1. `just test` and `flutter analyze` on the merge result.
2. On an Android 8.1+ device with Health Connect populated: grant, import, revoke, re-import. Confirm dedupe across repeated imports, that credited kcal equals reported × multiplier, and that a workout lands on the correct diary day under a non-midnight day boundary.
3. Tap the rationale affordance in Health Connect from **both** entry points, and confirm a second tap in the same session still works — the activity is `singleTop`, so that is the path that would silently stop working.
4. Upgrade an install carrying a 2.0.2 profile: the policy notice appears exactly once, "Read the policy" does not dismiss it, and it does not appear on fresh onboarding or after Try Demo.
5. A real food search returns what it did before, and the term no longer appears in the gateway log.
6. Set body weight to stones and check a value near a boundary (63.49 kg should read 10 st 0 lbs).
7. Raise system text size to 1.5× and check the diary's weekday headings are not clipped.

## Checklist
- [x] New interactive widgets include `Semantics(identifier: '...')` where needed
- [x] Localization updated in ARBs **and** `lib/generated/` when user-facing strings change
- [x] Codegen ran if DBOs/DTOs/env changed (`just build`)

## Deployment notes

0. ~~**Publish the corrected privacy-policy clauses first.**~~ **Done.** Four clauses are live in both documents and captured in #933: the composed Open Food Facts paragraph (findings 4, 6, 11), its retention sentence (15), the Supabase request contents (5), and a new health and fitness clause (7). `tool/policy_snapshot.dart` reports `purposes: 7 en, 7 de`, `services: 9 en, 9 de`, both agreeing, exit 0.

1. **Finding 12 becomes owed the moment this merges.** Its clause describes the `beforeSend` hook that strips the installation identifier. That is unreleased today, so publishing it would be false for everyone on 2.0.2 — which is why it was held. Once this ships, the published Sentry clause is the stale one. Text is drafted on #920.

2. **`minSdkVersion` rises 24 → 26.** Android 7.0, 7.1 and 8.0 users stop receiving updates at 2.0.2 and the store will not tell them why. This needs a line in the release notes — as does the fact that **workout import ships inert until Google approves the Health apps declaration**, since Health Connect access is allow-listed per package name independent of app version.

3. **iOS: the HealthKit capability is now enabled on the App ID (#769) — but both provisioning profiles are Invalid as a result.** Done on 28 Aug in the Developer portal, `HEALTHKIT` only; `HealthKit Estimate Recalibration` deliberately left off because `Runner.entitlements` does not declare it. `match AppStore com.opennutritracker.ont.opennutritracker` and the development profile both read **Invalid**, so **iOS signing is broken until the distribution profile is regenerated and pushed to the match repo — tracked in #947, which is now the hard iOS blocker.** CI cannot do it: `MATCH_READONLY` is hardcoded `'true'` in `ios-package` and `ios-deploy` and `workflow_dispatch` takes no inputs. The App ID is `com.opennutritracker.ont.opennutritracker` — #769's body names `com.opennutritracker.ont`, which is wrong. Enabling it in the portal was not sufficient: `ios-package` runs match in `readonly` mode, so a regenerated profile must be pushed to the match repo by hand. `ios-build` uses `--no-codesign`, so CI cannot catch any of this and the release merge is the first thing that can fail. Do a `workflow_dispatch` dry run first, and enable the capability through the **web portal, not Xcode's capability editor**, which can add `healthkit` to `UIRequiredDeviceCapabilities` and silently drop iPad support. `PrivacyInfo.xcprivacy` needs **no** change — HealthKit is not a required-reason API, and Apple defines "collect" as transmitting off device, which imported workouts never are.

4. **Revoke the USDA FDC key, then delete the secret — in that order (#925).** It shipped in 21 tags for a client that never ran. It is **not** in git history. Do not delete the `FDC_API_KEY` secret until this merges: `main`'s workflow still injects it in five places.

5. **#900 is done — no action left.** The `user.geo` country/region/city Sentry derives server-side at ingest is removed by an Advanced Data Scrubbing rule on the organisation, verified against real events rather than assumed. `beforeSend` cannot reach it — Relay fills it in after the hook runs — which is why this half was server-side. One tidy-up remains and blocks nothing: two rules are configured (`user.geo` and the undocumented `$user.geo.**` that #939 flagged), one of which is redundant.

6. **Two workflows only take effect once they reach `main`** — workflow files run from the default branch. `policy-snapshot.yml` (#923) and the project-board concurrency fix (#931) both start working at this merge, not before.

7. The probe branch `ci/prove-retry-on-hang` from #827 could not be deleted by push (`GH013`) and needs removing through the UI.


simonoppowa:
## Two iOS blockers found in App Store Connect, one of them dated tomorrow

Surfaced while capturing the current store declarations for #936. Neither is a code problem and neither was on this PR's deployment notes.

1. **The Apple Developer Program membership expires 29 Aug 2026.** That is tomorrow as of this comment. If auto-renewal is not on, the app is removed from sale and no submission is possible until it is renewed. Only the Account Holder can do it.

2. **The Developer Program License Agreement has been updated and must be reviewed and accepted** before existing apps can be updated or new ones submitted. Also Account-Holder-only. This one silently blocks the submission rather than the build — the same failure shape as deployment note 3, where `ios-build` runs `--no-codesign` and cannot see a signing problem.

Both gate shipping 2.1.0 to iOS, and neither affects the Play release, so they do not block merging this PR — only the iOS half of what follows it.

## Also worth knowing before submission

App Store Connect states: *"To make changes to the privacy policy, create a new app version. All other changes will be immediately available."*

So the **privacy policy URL is versioned** while data-type answers are not. Apple currently points at `53501884` (English) and Play at `53922100` (German). If the store URLs need correcting — and #936 found they contradict both each other and #875's decision — the Apple side has to ride along with a version, which means deciding it **before** submitting rather than after.

simonoppowa (commented):
## Review: Release 2.1.0

**Approve** — reviewed the full `fddba395..a34df46c` delta (90 commits, 163 files) locally, with `flutter analyze` clean and the full test suite passing (1,164 tests). All required CI checks green. I found no blocking code issues. Detailed findings below.

### Verification performed locally

- `flutter analyze` — no issues.
- `flutter test` — all 1,164 tests pass, including the new `import_workouts_usecase_test.dart` (profile-race, watermark, tombstone, day-boundary coverage).
- Codegen (`build_runner`) clean at the PR head; `hive_registrar.g.dart` consistent with the new DBO fields.
- `git merge-tree origin/main HEAD` — merges cleanly, zero conflicts.
- All three workflow YAMLs parse; ARB parity verified (943 keys × 9 locales).

### Workout import (#724, #743, #768, #929, #944…)

- The profile-switch race (#944) is properly closed: `_inFlight`/`_inFlightProfileId` serialization plus `_stillOnProfile` checks, and the atomic `_writeBoth` two-box write in `ConfigDataSource`.
- Tombstone design is sound: externalId → workout start time, pruned at `oldestUsefulTombstone`, legacy migration via `_mergedDeletedWorkouts`. `DeleteUserActivityUsecase` now writes tombstones; delete-all can't resurrect workouts because it wipes the per-profile config box.
- Android per-source energy attribution avoids double-counting all-day basal streams; permission re-checked before every read; backfill capped at 30 days, 24h overlap tolerance, 15-min debounce — all as described.
- `WorkoutCompensationCalc` — the Careau et al. 2021 constants and the documented BMI/body-fat approximations are honest, clamped (0.5–1.0), and the fallback multiplier (0.72) is used only when no body data exists.
- `LogUserActivityUsecase` correctly centralizes the activity-write + tracked-day-budget-raise pair, and `UpdateUserActivityUsecase` scales imported workouts proportionally instead of MET-recomputing them — that's the right call for measured data.
- Permissions are read-only (verified in AndroidManifest: no WRITE_* health permissions); disclosure dialog gates the permission request and dismissal = refusal. iOS: HealthKit entitlement + `NSHealthShareUsageDescription` added, no update-usage key (read-only), correctly wired into all build configs in the pbxproj.

### Sentry / privacy (#901, #917, #921, #926)

- Consent-gated init (`kReleaseMode && hasAcceptedAnonymousData`), tracing off, print breadcrumbs off, `_stripUser` in `beforeSend` — all verified.
- Policy notice is recorded on **close**, not open (#921), shown only by `MainScreen`, device-wide via the shared box. Good.
- ⚠️ **Process note, not a code defect**: as the PR body itself states, #900's `user.geo` server-side scrubbing rule is still outstanding — client-side stripping can't remove geo data Sentry attaches server-side. Merging with that rule missing ships the release with the known leak the PR description says blocks it. I'll trust the "do not merge yet" note in the body to govern timing.

### POST search (#913)

- Search terms now travel in the RPC request body (`search_food_summary` / `search_food_translation` / `food_summary_by_ids`); verified nothing user-supplied lands in URLs or query strings anymore. Client-side ranking with the 5× candidate pool is a reasonable workaround for PostgREST rejecting `.order(ts_rank(...))`.
- FDC removal is complete: no `FDC_API_KEY` / `FDCDataSource` references remain anywhere in `lib/`, `android/`, `ios/`, `.github/` (the workflow's stub keys were removed too), or `tool/`. `.env.example` updated.

### Bug-fix batch

Spot-checked the riskier ones:

- #629 (`scalableServingQuantity`) — last-metric-match regex approach correctly prefers "(355 ml)" over "12 oz"; gating the serving unit on scalability kills the 1-gram logging bug at all three call sites.
- #615 (stone rounding) — `poundsDecimals`-aware rollover guard; all three call sites pass `poundsDecimals: 1` consistently.
- #764 — the two halves (cancel OS alarms + flip the shared `notificationsEnabled` pref) are both best-effort before the wipe, with the right reasoning in the comments.
- #766 (calendar heading) — measuring via `TextPainter` instead of scaling 16.0 is the correct fix.
- `namedRouteOrFirst` — good hardening of the popUntil black-screen class of bug.

### Release mechanics

- Version 2.1.0+62, minSdk 24→26 pinned with a comment explaining why (Health Connect). Dependency pins look deliberate (health <13.3.2 win32 collision comment).
- `policy-snapshot.yml` is well thought out — commits to `develop` explicitly, warns-only on unfixable iubenda template drift, redacts the postal address and throws if it can't. Note it only takes effect once on `main` (as its own comments acknowledge).
- Notification service migration to flutter_local_notifications 20 named-parameter API is complete.

### Minor observations (non-blocking)

1. `delete_all_user_data_usecase.dart` still contains the line-wrapped comment "The user can re-enable crash reporting later from\n// Settings" — cosmetic, pre-existing style drift.
2. The `health` pin (<13.3.2) will need revisiting eventually; the comment documents the reason well.
3. The two App Store Connect account blockers (membership expiry 2026-08-29, license agreement) from the PR comment are external to this diff but gate the actual release.

**Verdict**: code is approve-ready. The only true merge blocker is the one the PR body already names — the Sentry `user.geo` Advanced Data Scrubbing rule (#900). Once that org-level rule is confirmed in place, this is good to merge.

simonoppowa:
## Device acceptance pass — Pixel 6, 2026-08-28

Ran against `develop` @ `a34df46c`, byte-identical to this PR's head. Built `--flavor develop --debug` and installed **in place** over the existing `2.0.2-develop` (both debug-signed on key `6a87678b`, so no uninstall). Pixel 6 `oriole`, Android 17 / SDK 37, locale **de-DE**, Health Connect `2025.09.04.00`.

The automated suite was not re-run — `android-integration-tests` already passed the identical single boot test on this commit. This covers what an emulator cannot.

### Verified working

| # | What | Evidence |
| --- | --- | --- |
| 1 | **Hive migration, pre-health → 2.1.0, in place** | 108M of real data preserved; home identical before and after: 65 kg, 2190 kcal übrig, 0/328 g · 0/60 g · 0/82 g. No FATAL from the package. |
| 2 | **#921 policy-change notice fires on upgrade** | Rendered in full German on first launch after update, naming the recipients, the new health section and the 24-hour approximate location. First time seen on hardware. |
| 3 | **#929 rationale intent** | `androidx.health.ACTION_SHOW_PERMISSIONS_RATIONALE` delivered to a **running** app landed on the Health Connect screen, not the diary — the `onNewIntent` path, which is exactly the #927 bug. |
| 4 | **#926 disclosure precedes the system prompt** | Full four-paragraph German text; no platform prompt appeared before it. |
| 5 | **#926 gates rather than merely precedes** | After ABBRECHEN: toggle off, all five permissions still `granted=false`, and `healthconnect.controller` was **never launched**. This is the Play User Data policy requirement demonstrated behaviourally. |
| 6 | **Real grant flow** | Weiter → Health Connect `PermissionsActivity` → all five granted with `USER_SET`. Plugin logged *"5 Health Connect permissions were granted!"*. |
| 7 | **Backfill window arithmetic** | First run read workouts over `2026-07-28T21:51 → 2026-08-28T21:51` — 30-day backfill **plus** the 24 h overlap tolerance, exactly as coded. |
| 8 | **Watermark persists and narrows the window** | Second run read `2026-08-27T21:51 → 2026-08-28T21:56` — stored watermark minus overlap tolerance, not the backfill. Watermark then advanced to 23:56. |
| 9 | **First-opt-in suggestion applies** | Credit moved 100 % → 67 % and persisted across screen re-entry. |
| 10 | **#930 policy tile, locale-routed** | Opened `iubenda.com/privacy-policy/53922100` — the German document, correct for a de-DE device. |
| 11 | **Opt-out** | Toggle off leaves controls inert; watermark and multiplier correctly retained as records rather than reset. |
| 12 | **"danach wird nichts mehr gelesen"** | Cold start after opt-out produced **zero** `FLUTTER_HEALTH` lines — `importIfDue` bailed at the enabled check without touching the platform. |

### Not verified — and this is the honest gap

**Zero workouts were imported, because this device has no workout data to import.** Not a defect, but it means the most important path is still unproven.

Established rather than assumed: Fitbit holds **0** granted Health Connect permissions, and the only packages on the device with any granted health permission are `com.android.shell`, `healthconnect.controller`, `gms`, `gsf` and the app itself. Nothing writes exercise records here.

So these remain untested on hardware:

- workout → `UserActivityEntity` written to the diary
- imported energy raising that day's calorie goal by the credit share
- the external-id dedupe across the overlapping window
- the day-boundary rule that files a 00:30 workout on the previous diary day
- deletion tombstones surviving a re-import

**Also unverified: the body-fat suggestion path.** I initially read the 67 % as proof it had a real body-fat reading. It is not. `WorkoutCompensationCalc` falls back to BMI, and the arithmetic settles it — BMI 22.5 gives t = (22.5−20)/13 = 0.192, compensation = 29.7 + 0.192×16 = 32.8 %, multiplier **0.672 → 67 %**. An exact match for the BMI path, and there is no body-fat data on the device for the other one to have used.

### Incident

One focus drift: a second `KEYCODE_BACK` exited the develop app and surfaced the **production** app. Nothing was tapped there and no state changed; recovered with an explicit `am start`. Every subsequent step used the rationale intent to navigate rather than Back.

### Verdict

Everything reachable on this device passes, including the two compliance-critical behaviours that ship for the first time in 2.1.0 (#926's gate and #929's intent). **The merge decision is not mine and I have not touched this PR.** The gap above is filed separately; it does not block, but it should not be forgotten either.

Screenshots: `scratchpad/device/00-…15-*.png`.

Issue #973, opened 2026-08-30
A model's count becomes grams, and logs an entry out by fiftyfold


simonoppowa:
## Parent

Part of the AI-assisted meal logging feature ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)), staged on `feature/ai-assisted-meal-logging`. Found during a device acceptance pass, [#830](https://github.com/simonoppowa/OpenNutriTracker/issues/830), run against the merge candidate for [#648](https://github.com/simonoppowa/OpenNutriTracker/pull/648).

## What to build

A count the model reports — *two eggs*, *three slices* — keeps its number and loses its unit. The number is then read as grams, and one tap writes it to the diary.

Reproduced on a Pixel 6 (`oriole`, locale `de-DE`), on `feature/ai-assisted-meal-logging` @ `20e9a9b4` (full flavour, debug, `2.1.0+62`), with Anthropic as the active provider:

1. Add meal → **Mehrere Einträge hinzufügen**
2. Type `zwei Eier und drei Scheiben Brot` — no digits, no unit symbols, so the deterministic parser cannot quantify it and the model path is the only thing that can answer
3. **Lebensmittel suchen**

The review screen confirms the model answered — *"Von KI gelesen — prüfe jede Zeile"* — and it read both items and both counts correctly. What arrives is:

| Row | Match | Amount | Energy |
| :-- | :-- | :-- | :-- |
| Eier | Classic, K-Classic, Kaufland | **2 g/ml** | 3 kcal |
| Brot | Backerkronung | **3 g/ml** | 8 kcal |

Both rows carry `Einheit für diese Menge prüfen`. Tapping **2 Einträge speichern** logs them: Home falls 2190 → 2179 kcal, and Breakfast reads **11 / 657 kcal** for two eggs and three slices of bread — out by roughly fiftyfold.

The provenance guarantee is intact throughout. Every number came from Open Food Facts and Supabase (`portions_by_food_ids`, `portion_labels_by_food_ids` in the log); nothing here is a model-emitted macro. That is the point worth drawing out: this is the [#847](https://github.com/simonoppowa/OpenNutriTracker/issues/847) shape again, one field over. Real numbers from a real record, every promise kept, and the entry still wrong — there it was the food that was wrong, here it is the unit the amount is counted in.

[...]

PR #978, opened 2026-08-31
fix(add-meal): only hold back an amount that would actually be wrong

simonoppowa:
Follow-up to [#975](https://github.com/simonoppowa/OpenNutriTracker/pull/975), which fixed [#973](https://github.com/simonoppowa/OpenNutriTracker/issues/973) and over-reached doing it.

## The regression

#975 excluded every row carrying `amountNeedsCheck` from the batch. That flag fires on **any** substituted unit, including substitutions that change nothing — so it refused amounts that were right.

`200ml Milch`, on the merged branch, Pixel 6:

| Match | Amount | Energy | Button |
| :-- | :-- | :-- | :-- |
| Milch, Milbona | 200 g/ml | 94 kcal | **`0 Einträge speichern`** |

94 kcal is correct for 200 ml of milk. The record is measured in grams and its dropdown offers `Portion`, `g`, `oz`, `g/ml` — **no `ml`** — so the typed unit is swapped for `g/ml` and the row is flagged. Before #975 that row logged, correctly. After it, an explicit and unambiguous amount is declined.

This was my error in #975, not a gap in #973: the issue asked that a *model-reported count with no portion* be held back, and I implemented something broader. Every test I wrote used the count branch, so the suite stayed green and a device probe found it.

## The change

`amountWouldBeWrong` narrows the exclusion to the three ways a number actually goes wrong, each asked on its own terms rather than inferred from the label:

- **nothing stated and nothing to count** — `2 Eier` about to be read as two grams
- **a serving the food cannot scale** — `3 serving` converts to a bare `3`, and so does `3 g/ml`. Numerically identical, both meaningless, so comparing conversions is blind to it; caught by asking what the unit *is*
- **a conversion that moves the number** — `4oz` against a record that cannot honour ounces, a 28-fold under-count

Everything #973 was about still stays out of the diary. Only rows where the swap costs nothing are let back in.

**The warning is deliberately unchanged.** A swapped unit is still worth pointing at even when it changes no number — the app is treating a typed `ml` as `g`, which is right for milk and would not be for oil. `amountNeedsCheck` keeps its meaning; only the decision to *withhold* the row is narrowed.

## Verified

- **2055 tests pass**, analyzer clean.
- **Mutation-checked**: restoring `willBeLogged` to exclude on `amountNeedsCheck` fails the named harmless-substitution test.
- New tests both ways — a substitution that preserves the number logs; one that moves it does not.

**Device verification is outstanding.** The APK is built and installed, but the phone locked itself mid-run and I will not unlock it. The two checks to run are `200ml Milch` (must now read `1 Eintrag speichern`) and `zwei Eier und drei Scheiben Brot` (must still read `0`). I would rather say that plainly than let the merge imply a device pass that did not happen.

### DIFF
--- a/lib/features/add_meal/presentation/bloc/bulk_add_bloc.dart
+++ b/lib/features/add_meal/presentation/bloc/bulk_add_bloc.dart
@@ -10,6 +10,7 @@ import 'package:opennutritracker/features/add_meal/util/meal_text_parser.dart';
 import 'package:opennutritracker/features/add_meal/util/portion_match.dart';
 import 'package:opennutritracker/features/add_meal/util/portion_unit.dart';
 import 'package:opennutritracker/features/meal_detail/presentation/bloc/meal_detail_bloc.dart';
+import 'package:opennutritracker/features/meal_detail/util/meal_quantity_converter.dart';
 
 part 'bulk_add_event.dart';
 part 'bulk_add_state.dart';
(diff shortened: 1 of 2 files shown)


simonoppowa:
## Device verification done — both directions hold

Ran on the Pixel 6 against this branch's build (`0208bca0`, installed 07:47), locale `de-DE`, anthropic active. Probes only; nothing written to the diary.

**The regression is gone.**

| Input | Match | Amount | Energy | Warning | Button |
| :-- | :-- | :-- | :-- | :-- | :-- |
| `200ml Milch` | Milch, Milbona | 200 g/ml | 94 kcal | shown | **`1 Einträge speichern`** |

Correct amount, correct energy, row loggable again. On the merged branch this read `0`.

**The #973 guard still holds.**

| Input | Match | Amount | Energy | Warning | Button |
| :-- | :-- | :-- | :-- | :-- | :-- |
| `zwei Eier und drei Scheiben Brot` | Eier / Brot | 2 g/ml, 3 g/ml | 3 kcal, 8 kcal | shown on both | **`0 Einträge speichern`** |

Neither row can reach the diary, which is the case this whole thread started from.

That is the split the change is for: **the warning is shown in both cases, and only the rows whose number would be wrong are withheld.** The milk row is flagged and loggable; the egg and bread rows are flagged and not.

The third branch — a conversion that moves the number, `4oz` against a record that cannot honour ounces — is covered by unit test rather than on device, because it needs a search result with a specific unit classification that cannot be arranged reliably against live Open Food Facts.

Issue #979, opened 2026-08-31
A provider that already has a key cannot be switched to


simonoppowa:
## Parent

Part of the AI-assisted meal logging feature ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)), staged on `feature/ai-assisted-meal-logging`. Found trying to switch providers for case 12 of [#830](https://github.com/simonoppowa/OpenNutriTracker/issues/830).

## What to build

**A hosted provider that already has a key cannot be switched to.** Every way out of the dialog reverts the change, and the button that would confirm it is not rendered.

Reported from the device: *"the switch dialog does not have an OK button to confirm the change"*. It does not, and that is the whole defect.

`_selectProvider` writes the new provider **immediately** — tapping the radio calls `setActiveProvider` before anything else. Then each exit:

| Exit | What happens |
| :-- | :-- |
| **OK** | Not rendered. Gated on `if (!_loading && (!_hasKey \|\| _provider == AiProvider.ownServer))` — `_hasKey` is true for any hosted provider with a stored key |
| **ABBRECHEN** | `_cancel()` → `restoreActiveProviderTag(await _providerOnOpen)` → **reverted** |
| **Back** | `PopScope(canPop: false)` routes to the same `_cancel()` → **reverted** |
| **Tap outside** | `barrierDismissible: false` — nothing happens |

So the user taps OpenRouter, the write lands, and then the only affordances on screen put it back. With keys stored for anthropic, openai and openrouter — the ordinary state for anyone who has configured more than one — **no switch between them is possible at all.**

This is not the [#848](https://github.com/simonoppowa/OpenNutriTracker/issues/848) behaviour misfiring. #848 was right that Cancel should undo a switch. What it did not check is that a *confirm* path exists for the same row, and for a keyed provider there is none.

## How it went unnoticed

Two ways, both worth fixing alongside.

**On device**, it looks like it works if anything kills the app before the dialog is dismissed. That is exactly what happened twice during this pass: the provider was selected by hand, and the harness's next step was `am force-stop` before relaunching to verify. The immediate write survived because the revert never ran. Both "successful" switches in this pass were an artifact of the driver, not the UI.

[...]

Issue #974, opened 2026-08-30
Four Settings rows, AI Assist among them, never reach the accessibility tree


simonoppowa:
## Parent

Part of the AI-assisted meal logging feature ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)), staged on `feature/ai-assisted-meal-logging`. Found while running the device acceptance pass, [#830](https://github.com/simonoppowa/OpenNutriTracker/issues/830), against the merge candidate for [#648](https://github.com/simonoppowa/OpenNutriTracker/pull/648).

## What to build

A contiguous band of rows in the **You** tab's embedded settings — including the AI Assist row this feature is configured from — is never exposed to the accessibility tree, at any scroll position.

The rows render, and they work when tapped by hand. They are simply not *there* as far as anything driving the app by accessibility is concerned.

Observed on `feature/ai-assisted-meal-logging` @ `20e9a9b4` (full flavour, debug, `2.1.0+62`), Pixel 6 (`oriole`), locale `de-DE`.

### What the tree shows

Expand **Einstellungen** on the You tab, then scroll. `uiautomator dump` only ever produces two states:

| State | Rows in the tree |
| :-- | :-- |
| top | `settings-food-units`, `settings-height-units`, `settings-body-weight-unit`, `settings-energy-unit`, `settings-kcal-goal-info`, `settings-kcal-adjustment`, `settings-macro-split`, `settings-per-meal-share`, `settings-nutrient-goals`, `settings-day-boundary` |
| bottom | `settings-delete-all-data`, plus *App-Daten exportieren*, *Cache leeren* and the whole **Über** group as text nodes |

A single swipe moves the tree straight from the first state to the second. Further swipes change nothing. Everything between the two — the **Anzeige** and **Darstellung** groups, the notification rows, and the first four rows of **Daten**: `settings-food-sources`, `settings-health-sync`, `settings-ai-assist`, `settings-import-custom-food` — appears in neither state, and no intermediate resting position exists that contains them.

Two independent probes agree:

- **Semantics identifiers.** `settings-ai-assist` never appears, at any scroll position, across repeated runs with fling swipes, slow drags, and `input motionevent` drags that end stationary. `settings-delete-all-data`, in the *same* `_SettingsGroup`, is exposed normally.
[...]

Issue #848, opened 2026-08-25
Cancel should not leave the provider switched


simonoppowa:
## Parent

Part of the AI-assisted meal logging feature ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)), staged on `feature/ai-assisted-meal-logging`. Found during the device acceptance pass, [#830](https://github.com/simonoppowa/OpenNutriTracker/issues/830).

## What to build

**Cancel does not cancel.** Selecting a provider in the AI settings dialog persists it immediately, so tapping **Cancel** afterwards leaves the newly-selected provider active.

Reproduced on a Pixel 6: active provider OpenRouter → open the dialog → tap **OpenAI** → tap **Cancel** → the Settings tile reads *"On — OpenAI"*.

The dialog is the one place that names each destination and states what it does with a photo, so reading those disclosures is a thing users are actively invited to do — the consent screen exists precisely to make people read them. Tapping through the four providers to compare them is therefore expected behaviour, and doing it silently repoints where the next meal photo is sent.

This is the same ordering [#835](https://github.com/simonoppowa/OpenNutriTracker/issues/835) already caught once from the other side: the consent screen used to claim terms were shown *"before you switch"* when the code had already persisted the switch.

Decide which way to resolve it — both are defensible, and the choice should be made deliberately rather than inherited:

- **Make Cancel revert** — restore the provider that was active when the dialog opened. Keeps the button honest, costs a snapshot of prior state.
- **Make the switch explicit** — keep persisting on selection, but stop calling the dismiss action *Cancel*, since nothing about the dialog is cancellable.

Reverting is the better default: a destination change is exactly the kind of thing a user expects Cancel to undo.

## Acceptance criteria

- [ ] Opening the dialog, selecting a different provider, and dismissing without confirming leaves the originally-active provider active
- [ ] Confirming still switches, and still persists
- [ ] Reading another provider's disclosure never changes where anything is sent
- [ ] A provider switch that *is* confirmed still gates on consent exactly as today
- [ ] Mutation-checked: removing the revert fails a named test

## Blocked by

- None — can start immediately.

simonoppowa:
Fixed in [#856](https://github.com/simonoppowa/OpenNutriTracker/pull/856), merged to `feature/ai-assisted-meal-logging`. CI green.

Cancel — and the system back button, which was the one exit with no handler at all — now restore the provider that was active when the dialog opened.

[...]

Issue #830, opened 2026-08-24
Device acceptance pass for #648 on a Pixel 6


simonoppowa:
Device acceptance pass for [#648](https://github.com/simonoppowa/OpenNutriTracker/pull/648), run on real hardware before it merges into `develop`.

## What is under test

| | |
|---|---|
| Build | `feature/ai-assisted-meal-logging`, full flavour, debug. Ticks below name the head they were taken on: `a7dd7c66` for the first two passes, `20e9a9b4` and later for the third |
| Why the head and not a merge preview | was **0 behind** `develop` when this issue was written; `develop` is now **3 ahead** (#827, #828, #829 — none touch AI code), so the head is no longer the merge result |
| Device | Pixel 6 (`oriole`), Android 17, serial `18101FDF60010M` |
| Install state | `.develop` was **already present** (installed 2026-08-15) and is left untouched; the full flavour is installed fresh for this pass — see the correction comment |
| Endpoints | live: **anthropic**, **openai**, **openrouter** — all three wire implementations shipped in #648 |

Built with real API credentials for the food databases, because every row in the review screen is resolved against Open Food Facts / Supabase — a stubbed build cannot exercise tier 0 at all.

## Known and excluded

[#826](https://github.com/simonoppowa/OpenNutriTracker/pull/826) is still open against this branch, so the meal-row horizontal overflow at 2x text is **present in this build by design**. Invisible at default font size. Not to be filed again from this pass.

### Setup

- [x] **01** Clean install, first launch, onboarding completes and lands on Home

### Tier 0 — the deterministic parser, no key configured

- [x] **02** Multi-item text parses into one row per item, each resolved against the food database
- [x] **03** A row's amount and unit can be edited, and its energy figure follows
- [x] **04** An item that matches nothing is flagged, and offers a working fallback rather than being dropped
- [x] **05** Log all — one intake per **loggable** row lands under the right meal type, and Home and Diary totals move by the right amount. A row whose amount the app cannot vouch for is **held back rather than logged**, and the button's count says how many will be written ([#973](https://github.com/simonoppowa/OpenNutriTracker/issues/973))
- [x] **06** With no key configured the camera action is not offered at all

### Settings, per provider

[...]

PR #980, opened 2026-08-31
fix(settings): offer a way to confirm a switch to a keyed provider

simonoppowa:
Fixes [#979](https://github.com/simonoppowa/OpenNutriTracker/issues/979) — which will not auto-close from here, since `Closes` only fires on the default branch.

## The bug

Tapping a provider row writes it immediately. Then every exit from the dialog undid it:

| Exit | Behaviour |
| :-- | :-- |
| **OK** | Not rendered — gated on `!_hasKey`, true for any provider with a stored key |
| **ABBRECHEN** | `_cancel()` → `restoreActiveProviderTag(...)` |
| **Back** | `PopScope(canPop: false)` → the same `_cancel()` |
| **Outside** | `barrierDismissible: false` |

So with keys stored for two hosted providers — the ordinary state once someone configures a second — **no switch between them was possible.** That is the choice the four-destination design exists to offer.

Reported from a device as *"the switch dialog does not have an OK button to confirm the change"*. It did not.

## The change

OK is offered when something can still be typed into **or when there is already a change to confirm**:

```dart
if (!_loading &&
    (!_hasKey || _provider == AiProvider.ownServer || _changed))
```

`_changed` rather than comparing against the provider on open: a model pick and the enabled toggle are also changes someone may want to keep, and both are already written by the time they set it.

**Cancel and back still revert an unconfirmed switch** — [#848](https://github.com/simonoppowa/OpenNutriTracker/issues/848) is what makes tapping down the list to read each disclosure safe, and it holds. An untouched dialog still shows no OK, which is what kept a no-op confirm off the screen to begin with.

## Why CI never caught it

The `leaving without confirming (#848)` group already had a test named **`OK still switches, and still persists`**, and it passed throughout. Its provider has no key, so `_hasKey` is false and OK renders — the one side of the gate where the button exists. The whole defect lived in the precondition the test happened not to set.

The new test stores keys first, which is the state the report came from. `an untouched dialog offers nothing to confirm` pins the other edge, so the gate is widened rather than removed.

## Verified

- **2058 tests pass**, analyzer clean.
- **Mutation-checked**: restoring the `!_hasKey`-only gate fails `a keyed provider can be switched to and confirmed (#979)`.
- Both #848 directions re-pinned with keys stored — Cancel reverts, and now OK does not.

**Device verification is outstanding, and needs a person.** The build is installed on the Pixel 6 with all three keys present, but the AI Assist row cannot be driven over ADB ([#974](https://github.com/simonoppowa/OpenNutriTracker/issues/974)), so the four taps have to be done by hand: open the dialog, tap another provider, check **OK** is now on screen, tap it, and confirm the Settings row names the new provider. Until then this rests on the widget tests alone.

Worth noting the two provider switches earlier in the #830 pass only stuck because the app was force-stopped before the dialog could revert — an accident of the harness. After this, they should stick because the user confirmed them.

### DIFF
--- a/lib/features/settings/presentation/widgets/ai_assist_dialog.dart
+++ b/lib/features/settings/presentation/widgets/ai_assist_dialog.dart
@@ -1056,14 +1056,30 @@ class _AiAssistDialogState extends State<AiAssistDialog> {
             child: Text(s.dialogCancelLabel),
           ),
         ),
-        // Offered whenever something on screen can still be typed into. For
-        // the hosted three that is exactly "before a key exists", which is
-        // what this used to say. A server the user runs keeps an address and
-        // a model name editable for as long as the dialog is open, and
-        // hanging OK off the credential left both fields rendered with no way
-        // to commit a change — the only exit was Remove, which discards the
-        // address as well.
-        if (!_loading && (!_hasKey || _provider == AiProvider.ownServer))
+        // Offered whenever something on screen can still be typed into, **or
+        // whenever there is already a change to confirm**. For the hosted
+        // three the first half is exactly "before a key exists", which is
+        // what this used to say on its own. A server the user runs keeps an
+        // address and a model name editable for as long as the dialog is
+        // open, and hanging OK off the credential left both fields rendered
+        // with no way to commit a change — the only exit was Remove, which
+        // discards the address as well.
+        //
+        // The second half is #979. Switching to a provider that already has
+        // a key is a change, and it had no way out that kept it: `_hasKey`
+        // is true for that provider, so OK was never rendered, while Cancel
+        // and the back button both run `_cancel` and restore the previous
+        // one. Every exit undid the switch, which made a second configured
+        // destination unreachable — the choice the four-destination design
+        // exists to offer. The test that covered "OK still switches" passed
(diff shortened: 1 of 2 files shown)


simonoppowa:
## Device verification done — by hand, on the Pixel 6

Build from this branch (`f032038f`) installed, all three hosted keys present, active provider **openrouter**.

Opened the dialog, tapped **OpenAI**, and — the point of the change — **OK was on screen**. Tapped it.

| Check | Result |
| :-- | :-- |
| Settings row in the **live app**, no restart | `KI-Unterstützung · Experimentell · An — OpenAI` |
| Settings row after **force-stop and relaunch** | `An — OpenAI` |

Both directions matter and are checked separately. The first shows the dialog closed **with** the switch intact — OK exits without reverting, which is the defect. The second shows it reached storage.

That distinction is not academic here. Every provider switch earlier in the #830 pass survived only because the harness force-stopped the app before `_cancel` could run, so a restart-only check would have passed against the broken build too. This is the first switch in the pass that stuck because a person confirmed it.

`ai-assist-save-key` is now reachable for a provider that already has a key, which is the state every user is in once they configure a second destination.

PR #995, opened 2026-09-01
feat: add scan and create-custom actions to no-results search state (#578)

### LINKED ISSUE #578: Add next-step actions to empty food search results
### Problem / motivation

When food search returns no matches, the screen offers little guidance about the next useful action. Users can interpret this as a dead end even though they can scan a barcode or create a custom food.

This is a focused extraction from #458.

### Proposed solution

Add an actionable empty state to no-result food searches.

Acceptance criteria:

- [ ] Distinguish an untouched/empty query from a completed search with zero results
- [ ] For a zero-result search, explain that no matching foods were found
- [ ] Offer actions to scan a barcode and create a custom food
- [ ] Preserve the current query when navigating away and back where practical
- [ ] Do not show the zero-result state while providers are still loading
- [ ] Handle partial provider failures separately from genuine zero results
- [ ] Add stable `Semantics(identifier: ...)` values to each action
- [ ] Add widget tests for initial, loading, results, partial-failure, and zero-result states
- [ ] Add localized strings to all ARBs and manually maintained generated localization files

Quick calorie/macro entry may be included only if the existing flow exposes a suitable route; creating a new quick-entry feature is out of scope.

### Alternatives considered

Showing only a generic retry button. Retry does not help when the query genuinely has no matching item.

### Feature area

Food search / Open Food Facts

### Target platforms

Both

### Mockups or references

Parent proposal: #458

### Additional context

Use existing scanner and custom-food routes. This issue should not redesign result ranking or provider APIs.

### Checklist

ibrahim-iqbal:
## Summary

Zero-result searches used to hit a wall — just a "No results found" hint, no next step. This PR turns that empty state into a useful one by surfacing the two things the app can actually do next: scan a barcode or open the custom-food form.

## Type of change

- [x] New feature
- [x] Localization

## Related issues

Part of #578.

## Changes

- \`NoResultsWidget\` now takes optional \`onScanBarcode\` and \`onCreateCustomFood\` callbacks and renders "Scan barcode" (\`OutlinedButton.icon\`) and "Create custom food" (\`FilledButton.tonalIcon\`) under a short explanatory subtitle when either callback is supplied. The bare-hint form still works for any caller that leaves both null.
- All four call sites in \`AddMealScreen\` (All / Products / Food / Recent) pass both callbacks. Barcode reuses \`_onBarcodeIconPressed\`; create-custom reads the current imperial-units flag from \`AddMealBloc\` (via a new \`_onCustomAddFromEmptyState\` helper) before running the same confirm-then-open-editor flow as the app-bar Add button.
- \`Semantics(identifier: 'search-no-results-scan-barcode' | 'search-no-results-create-custom-food')\` on both buttons, per the ADB / Appium driver convention in \`AGENTS.md\`.
- Three new ARB keys: \`noResultsSubtitle\`, \`noResultsScanBarcode\`, \`noResultsCreateCustomFood\`. Added to every locale ARB (all nine stay at the same key count) with the English text carried across as a fallback until translations land via Weblate / Crowdin.
- New \`test/widget_test/no_results_widget_test.dart\` covering the three visible shapes: no callbacks (bare hint), both callbacks (buttons render and fire), and scan-only (create button hidden). All 3 pass locally on Flutter 3.44.4.

### Acceptance criteria coverage from the issue

- [x] Distinguish untouched/empty query from completed-with-zero-results — kept as-is; the existing \`_pendingSpinner\` / \`DefaultsResultsWidget\` / \`NoResultsWidget\` split already handles this.
- [x] Explain that no matching foods were found — new subtitle string.
- [x] Offer actions to scan a barcode and create a custom food.
- [x] Preserve the current query — the \`ValueNotifier<String>\` in \`_AddMealScreenState\` already survives navigation to the scanner / editor and back.
- [x] Do not show zero-result state while providers are still loading — unchanged; \`_productsPending\` / \`_foodPending\` gate the empty state on both single-source and merged views.
- [ ] Handle partial provider failures separately from genuine zero results — **not tackled here.** Currently a failed source is treated as answered (comment in \`_buildResults\`) so a merged view can still surface the other source's results. Turning this into a distinct visible state feels like its own PR; happy to tackle in a follow-up if you'd like.
- [x] Stable Semantics identifiers.
- [x] Widget tests for the visible shapes of \`NoResultsWidget\`. I did not add BLoC-level widget tests for initial / loading / results / partial-failure states of the whole search — those cover a lot of surface area beyond the empty-state fix and would be better as a separate test PR.
- [x] Localized strings in every ARB (English text carried into the locale files).

## Screenshots / recordings

Not attached — I don't have a device set up locally to record right now. Layout is a \`Wrap\` under the existing \`EmptyHint\`, so buttons flow to a second row on narrow widths.

## Test plan

- [ ] Manual check on Android (not run — happy to run once someone with a device tries it)
- [ ] Manual check on iOS

### Steps

1. Open a meal type from Home (e.g. Breakfast).
2. Type a query with no matches (e.g. \`asdfqwer\`).
3. Confirm the empty state shows the subtitle plus both action buttons.
4. Tap "Scan barcode" — scanner opens with the current meal + day.
5. Tap "Create custom food" — the existing confirm dialog opens; confirming pushes the edit-meal editor.
6. Repeat with the Recent chip empty (no recent meals for this meal type).

## Checklist

- [x] Code follows project style (\`dart format\`, 120-char line width — analyzer clean on the three changed files)
- [x] New interactive widgets include \`Semantics(identifier: '...')\`
- [x] Localization updated in ARBs (\`lib/generated/\` is gitignored — regenerated locally via \`flutter gen-l10n\` before running tests)
- [x] No codegen-touching changes
- [x] No secrets or \`.env\` values committed


simonoppowa:
@codex review

simonoppowa:
Thanks for this — the empty state is a real gap and the two actions are the right ones. A few things to fix before it can go in.

## The localization is the blocker, and it is mostly copy-paste

The PR adds three keys, and all three carry the English string in every one of the eight non-English ARBs. `CONTRIBUTING.md` asks for a real translation rather than English as a placeholder, and `just check_l10n` cannot catch this — it fails only on *missing* keys, so a key present everywhere with English in it passes CI green.

**The good news: the app already says one of these.** `customMealBarcodeScanButton` is already exactly `"Scan barcode"` in English and translated in all nine locales. Please copy those values verbatim rather than write a second phrasing for the same action:

| locale | value to copy |
|---|---|
| de | `Barcode scannen` |
| cs | `Naskenovat čárový kód` |
| it | `Scansiona codice a barre` |
| pl | `Skanuj kod kreskowy` |
| sk | `Naskenovať kód` |
| tr | `Barkod tara` |
| uk | `Сканувати штрих-код` |
| zh | `扫描条形码` |

For the other two, these follow the terminology already in the ARBs — `newCustomMealLabel` and `customMealsMergeAction` for the noun, `createRecipeTitle` for the verb form:

**`noResultsCreateCustomFood`**

| de | `Eigenes Lebensmittel erstellen` |
|---|---|
| cs | `Vytvořit vlastní potravinu` |
| it | `Crea alimento personalizzato` |
| pl | `Utwórz własny produkt` |
| sk | `Vytvoriť vlastné jedlo` |
| tr | `Özel yiyecek oluştur` |
| uk | `Створити власний продукт` |
| zh | `创建自定义食物` |

**`noResultsSubtitle`**

| de | `Zu dieser Suche passt kein Lebensmittel. Versuche einen anderen Namen, scanne den Barcode oder lege ein eigenes an.` |
|---|---|
| cs | `Tomuto hledání neodpovídá žádná potravina. Zkuste jiný název, naskenujte čárový kód nebo si vytvořte vlastní.` |
| it | `Nessun alimento corrisponde a questa ricerca. Prova un altro nome, scansiona il codice a barre o creane uno personalizzato.` |
| pl | `Żaden produkt nie pasuje do tego wyszukiwania. Spróbuj innej nazwy, zeskanuj kod kreskowy lub utwórz własny.` |
| sk | `Tomuto vyhľadávaniu nezodpovedá žiadne jedlo. Skúste iný názov, naskenujte čiarový kód alebo si vytvorte vlastné.` |

(Machine translation is an accepted starting point here per `CONTRIBUTING.md`; native-speaker review after merge is welcome.)

## Three code points

**The unit lookup returns the wrong bloc.** `AddMealBloc` is registered with `registerFactory`, so `locator<AddMealBloc>()` hands back a *fresh* instance that never received `InitializeAddMealEvent` — its state is always initial, so `usesImperialUnits` always reads `false`. An imperial user creating a custom food from the new empty state gets the editor in metric, unlike the existing app-bar action which reads its initialized instance. Read the screen's bloc, or pass the flag from the state you already have.

**The footer contradicts the results above it.** At the `remoteSourceEmpty` trailing item, passing the callbacks makes this widget render "No foods matched that search" plus the full recovery actions *underneath* real custom, recipe, history or cached matches. That path wants a narrower "no remote results" footer, or the bare widget without the new content.

**Same on the Recent tab.** With an empty query and no history the actioned widget now claims a search returned nothing when no search ran. `noMealsRecentlyAddedLabel` already exists and is the right copy there.

## Two smaller ones

The `Column` isn't scrollable — in landscape or at a large text scale, the 72px icon plus subtitle plus two buttons can overflow the results area and push the buttons off-screen.

And a naming inconsistency worth catching now: the new button says **"Create custom food"** while the dialog it opens (`createCustomDialogTitle`) says **"Create custom meal item?"** — two nouns for one concept, one tap apart. Not introduced by you, but this PR is what puts them next to each other.

The `Semantics(identifier:)` values and the ARB key parity are both correct — thanks for getting those right, they're easy to miss.

ibrahim-iqbal:
Addressed all three points in `555f3ce`:

1. **Imperial fallback**: dropped the AddMealBloc lookup and now read `usesImperialUnits` off the source-specific state (`ProductsLoadedState` / `FoodLoadedState` / `RecentMealLoadedState`) that's already in scope at each call site. The factory registration made the old lookup return metric every time.
2. **remoteSourceEmpty footer**: the products/food ListView footer now renders as `const NoResultsWidget()` with no callbacks, so the appended row shows the bare hint instead of contradicting the visible matches above. Recovery actions stay on the pure empty state.
3. **Overflow in landscape / large text scales**: wrapped the widget's Column in `SingleChildScrollView` so the empty state scrolls when the results area is tighter than icon + subtitle + buttons need.

Existing widget tests still pass.

ibrahim-iqbal:
Thanks for the thorough pass — everything's in as of `60117ba`.

**Localization** — all three keys now carry a real string in every locale. `noResultsScanBarcode` is copied verbatim from `customMealBarcodeScanButton` per locale; `noResultsCreateCustomFood` and the five subtitles use your values. For the three subtitles you didn't supply (tr / uk / zh) I machine-translated against the nouns already in those ARBs (`yiyecek` / `продукт` / `食物`, matching `newCustomMealLabel`) — happy to take a native pass post-merge. `l10n_untranslated.json` is `{}` after `flutter gen-l10n`.

**Recent tab** — empty query + no history now renders `EmptyHint` with `noMealsRecentlyAddedLabel`. The actioned search empty state only shows when a non-empty query filters Recent to nothing.

**Unit lookup / footer / scroll** — these landed in `555f3ce` before this: imperial now comes from the source-specific loaded state at each call site, the `remoteSourceEmpty` footer is the bare widget, and `NoResultsWidget` is wrapped in `SingleChildScrollView`.

**Naming** — I kept "Create custom food" because it matches the ARB majority (`newCustomMealLabel` = "New Custom Food", `customMealsMergeAction` = "Merge with another custom food"); `createCustomDialogTitle` ("custom meal item") is the outlier. Aligning that title means touching nine locales for a string this PR doesn't own, so I left it — glad to open a small follow-up if you want it brought in line.

simonoppowa:
Thanks for this — retargeting the base from `main` to `develop`, and flagging what that changes for you.

`main` only ever receives release merges in this repo (`docs/RELEASING.md`), so a feature PR based on it can't land as-is: merging it would put a change on `main` that `develop` doesn't have, which is the backport gap the release runbook exists to prevent. `develop` is the branch feature work targets.

The side effect is that GitHub now shows this as conflicting. That isn't a problem with your change — the branch was cut from `main`, and `develop` is currently 42 commits ahead of it, so the comparison is against a very different tree.

To get it green, rebase onto current `develop`:

```bash
git remote add upstream https://github.com/simonoppowa/OpenNutriTracker.git
git fetch upstream
git rebase upstream/develop
git push --force-with-lease
```

Your actual change (the scan and create-custom actions on the no-results search state, #578) should survive that cleanly — the conflicts are almost entirely in files you didn't touch, inherited from the old base.

Happy to help if the rebase turns up anything awkward.

ibrahim-iqbal:
Thanks for the retarget and the runbook. Rebased onto `develop` — since the branch was originally cut off `main` a straight `git rebase upstream/develop` tried to replay develop's own commits back on top of mine, so I cherry-picked the four changes onto a fresh `develop` head instead. Diff is now scoped exactly to the empty-state work (add_meal_screen, no_results_widget, the nine ARB entries you reviewed, and the widget test). CI is running.

Force-pushed with `--force-with-lease`.

simonoppowa:
Thanks for the rebase and the translation pass — the ARBs are exactly right now (all nine at parity, Scan barcode copied verbatim per locale), analyze and tests are green on a local merge with develop, and the Recent-tab split is the behaviour I wanted.

One small thing before I merge, on the overflow fix from the last round. The SingleChildScrollView in NoResultsWidget only scrolls when its parent bounds its height. On the All and Recent tabs it sits inside an Expanded, so it works there — but on the Products and Food tabs (add_meal_screen.dart:397 and :450) it is returned as a plain child of the tab's Column, gets unbounded height, and never scrolls; the Column overflows instead. I reproduced it in a widget test at a 260px-tall results area: 32px overflow, maxScrollExtent 0. Wrapping those two returns in Flexible(child: ...), the same way the ListView right below them already is, fixes it.

Could you also change "Fixes #578" to "Part of #578"? You were upfront that partial-failure handling and the state tests are deferred, and I'd like the issue to stay open for those. Happy to take the follow-up whenever you have time.

ibrahim-iqbal:
Both applied.

- `94f70487` wraps the two return sites (`add_meal_screen.dart:397` for Products, `:450` for Food) in `Flexible`, matching the shape the non-empty `ListView.builder` returns just below them, so `SingleChildScrollView` inside `NoResultsWidget` now gets bounded height on those tabs too. Diff is scoped to those two returns only.
- PR title/description updated: `Fixes #578` → `Part of #578` so the issue stays open for the partial-failure handling and state tests. Happy to pick those up as a follow-up.

PR #1005, opened 2026-09-01
fix(net): read the zone id on a link-local address

simonoppowa:
Found while auditing the guard for bypasses the review of #988 might have
missed. It is not one: it is the guard refusing a destination it exists to
permit.

### What happens

A link-local address needs a zone id to be routable on a host with more than
one interface, and a URI spells that zone escaped. `Uri` stores
`http://[fe80::1%wlan0]` with a host of `fe80::1%25wlan0` — and typing the
bare `%` produces the identical string, so this is not a corner you have to
know RFC 6874 to reach.

`InternetAddress.tryParse` answers `null` to that escaped form. The address
therefore fell through to the DNS branch, where the lookup cannot succeed,
and the caller was told the server is **unreachable** — about a link-local
server this check exists to permit, and which an unguarded client reaches
perfectly well. That is the wrong half of the `unreachable` /
`insecureDestination` split those two answers exist to keep apart, and it
sends someone to fix the wrong thing.

Verified against a real `HttpServer` bound to a link-local address: unguarded
client `200`, guarded client `SocketException: Failed host lookup`.

**Fail-closed.** The guarantee was never at risk — this is a refusal, not a
leak. Severity is low and the affected population is narrow (a zoned
link-local address means a direct-link setup), but for anyone who lands on it
the feature is broken with a misleading error.

### The fix

`dart:io` performs this same substitution itself, in
`escapeLinkLocalAddress`, before it resolves or connects — which is exactly
why an unguarded client works. Doing it here is what makes the check agree
with the socket layer it guards.

A host that still fails to parse falls through to the lookup unchanged, so a
name is never touched: the `25` comes off only when it directly follows the
first `%`.

### The tests are fussier than they look

Two earlier versions of them passed with the fix removed. Recorded here
because each one is a trap that produces a green, worthless test:

- **A numeric zone tests nothing.** `%1` escapes to `%251`, which parses
  happily as scope 251, so the guard never stumbled. Only a *named* zone
  reproduces the failure.
- **A named zone cannot be hardcoded.** `tryParse` resolves it against the
  running machine's interfaces, so `%eth0` is `null` on a host with no eth0.
  The tests ask `NetworkInterface.list()` for a name, and skip if there is
  none.
- **There is no zoned-public-address case**, because `tryParse` rejects a
  zone on anything that is not link-local (`2001:db8::1%lo` is `null`), so
  such a URL never reached this branch. An earlier test asserting otherwise
  was asserting a fiction; it is gone, with a comment in its place.

The tests are hermetic on the passing path — the injected lookup calls
`fail()`, so reaching the resolver *is* the failure. With the fix reverted
they report `a zoned literal reached the resolver: fe80::1%25lo`.

Locally: 2118 tests pass, `flutter analyze` clean.

### Not proposed for 2.2.0

This predates the release, so it is not a regression in it. #1004 carries the
#988 review fixes onto `release/2.2.0` and is already in review; adding this
would widen it. Happy to cherry-pick it across if you would rather it shipped.

### DIFF
--- a/lib/core/utils/plaintext_destination_guard.dart
+++ b/lib/core/utils/plaintext_destination_guard.dart
@@ -88,6 +88,32 @@ class PlaintextDestinationGuard {
   static Future<List<InternetAddress>> _resolve(String host) =>
       InternetAddress.lookup(host, type: InternetAddressType.any);
 
+  /// Turns the `%25` a URI must spell a zone id with back into the `%` that
+  /// [InternetAddress] parses.
+  ///
+  /// A link-local address needs a zone to be routable on a host with more
+  /// than one interface, and RFC 6874 says a URI escapes it: `Uri` stores
+  /// `http://[fe80::1%wlan0]` with a host of `fe80::1%25wlan0` — typing the
+  /// bare `%` gets you the same thing, so this is not a corner someone has to
+  /// know the RFC to reach. `InternetAddress.tryParse` returns null on that
+  /// escaped form, which sent the address down the DNS branch below, where
+  /// the lookup fails and the caller is told the server is **unreachable** —
+  /// about a link-local server this check exists to *permit*, and which an
+  /// unguarded client reaches perfectly well.
+  ///
+  /// `dart:io` does exactly this substitution itself, in
+  /// `escapeLinkLocalAddress`, before it resolves or connects. Doing it here
+  /// too is what makes the check agree with the socket layer it is guarding.
+  ///
+  /// Anything that still fails to parse falls through to the lookup unchanged,
+  /// so a name is never touched: the `25` is only dropped when it directly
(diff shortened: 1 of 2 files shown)


simonoppowa:
Closing in favour of #1004, which now carries this commit onto `release/2.2.0` — the two branches hold byte-identical content for `plaintext_destination_guard.dart` and its test file. `develop` gets the fix back when the release merges, so keeping both open would only route the same change to `main` twice under different SHAs.

Nothing here is abandoned: the commit is `cabb3b05` on `fix/plaintext-guard-2.2.0`. The branch `fix/plaintext-guard-link-local-zone` is left in place rather than deleted, in case this needs reopening.

Background on the bug is preserved in #1004's description, including why two earlier versions of its tests passed with the fix removed.

PR #1010, opened 2026-09-01
docs(readme): name every contact in What leaves your device #989

### LINKED ISSUE #989: The README's "what leaves your device" table is wrong in four rows
## Parent

Found in a release-readiness audit before the 2.2.0 merge ([#988](https://github.com/simonoppowa/OpenNutriTracker/pull/988)). Not a merge blocker — a documentation pass.

## What to build

The README's **"What leaves your device"** table is wrong in four rows. That matters more here than in most projects, because the README says the table *is* the product: *"The README does not say 'we respect privacy', it says what leaves your device — these four destinations, nothing else"* ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)).

### 1. The own-server row understates when the app contacts your server

`README.md:145` says the address is contacted *"when you press Load models, and each time a meal line or photo is read"*. There are **four** occasions, and two are missing:

- **Finishing with the address field** — `_endpointCommitted` (`ai_assist_dialog.dart:431`) calls `_loadModels()` on focus loss (`:239`) and on the keyboard's done key (`:889`). Not only when the button is pressed.
- **Saving** — a valid own-server save starts the setup check unawaited (`:685`), sending a fixed example line and a bundled photograph.

The row is immediately followed by *"This is the one row you can falsify yourself: your own server's access log settles it."* **A reader who takes that up will see requests the row does not predict.** That is the row inviting falsification and then failing it.

Same defect class as [#985](https://github.com/simonoppowa/OpenNutriTracker/issues/985), which was the in-app version of this sentence.

### 2. The Open Food Facts row omits photograph downloads

Photographs are fetched whenever a food is shown — search results, meal screens, and the diary card of an already-logged meal — and the URL contains the product barcode. The published privacy policy already says this (`docs/privacy-policy/en.txt:113`); the README does not. The row's User-Agent sentence also does not hold for photograph downloads, which are not identified.

### 3. The Supabase row omits the app language and the enabled-source filter

The policy's own sentence at `docs/privacy-policy/en.txt:159` is correct and can be reused verbatim rather than composing new wording.

### 4. The README contradicts itself about the export bundle

Two sentences disagree about what an export contains, and the weight-log half is a privacy claim in the wrong direction — it understates what leaves the device.

## Acceptance criteria

- [ ] The own-server row names all four occasions, so the access-log dare holds
- [ ] The Open Food Facts row covers photograph downloads and scopes the User-Agent claim to the requests that carry one
- [ ] The Supabase row matches `docs/privacy-policy/en.txt:159`
- [ ] The two export sentences agree with each other and with the code
- [ ] Wording is taken from the published policy where it already exists, rather than newly composed — two documents saying the same thing differently is how this drifted

## Blocked by

- None.

AzazelSensei:
## Summary
Fixes the four wrong rows in the README's "What leaves your device" table so they match the code and the published policy.

## Type of change
- [x] Documentation

## Related issues
Fixes #989

## Changes
- Own-server row names all four contacts: finishing with the address field, Load models, the setup check on save, and a meal line or photo read
- Open Food Facts row covers photograph downloads and scopes the User-Agent to searches and product lookups
- Supabase row matches `docs/privacy-policy/en.txt:159` (app language and the enabled-source filter)
- The Portable blurb and the export bullet now agree with each other and with the JSON/CSV bundles — weight history does leave in the JSON zip

## Screenshots / recordings
n/a — README only

## Test plan
- [ ] Manual check on Android

### Steps
1. Diff README against `ai_assist_dialog.dart` (`_endpointCommitted`, `_loadModels`, unawaited probe on save) and `export_data_usecase.dart` (`weight_log.json` in the JSON bundle).
2. Diff the Open Food Facts and Supabase rows against `docs/privacy-policy/en.txt`.

## Checklist
- [ ] New interactive widgets include `Semantics(identifier: '...')` where needed
- [ ] Localization updated in ARBs **and** `lib/generated/` when user-facing strings change
- [ ] Codegen ran if DBOs/DTOs/env changed (`just build`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ whether that is losing weight, gaining it, managing a condition, or simply knowi
 - **⚖️ Weight history:** Capture weight during onboarding and on demand, see the trend on a chart with a dashed line at your target weight, and optionally taper the calorie goal as you approach it.
 - **🎨 Material You + theme picker:** Adopt the system accent colour on Android 12+, or pick from sixteen built-in presets. The app icon adapts to iOS dark and tinted appearances and to Android themed icons.
 - **🔢 kcal or kJ:** Switch the energy unit globally; every diary entry, target, and chart reflects the choice.
-- **📤 Export and import:** Export your diary entries, activities, tracked days, recipes, weight history and custom activity templates to a JSON zip (recipe photos included, plus photos of custom meals you saved for next time — a one-off custom meal photo that was not saved is not in the zip), or diary, activities and tracked days as a CSV zip — you pick one format, the bundle carries only that one ([bundle format](docs/export-format.md)). Paste a JSON blob to import meals, and share a single meal or activity as a QR code another phone can scan. Your profile, custom-meal catalogue, water and fasting history stay out of the bundle.
+- **📤 Export and import:** Export your diary entries, activities, tracked days, recipes, weight history and custom activity templates to a JSON zip (recipe and custom-meal photos included), or diary, activities and tracked days as a CSV zip — you pick one format, the bundle carries only that one ([bundle format](docs/export-format.md)). Paste a JSON blob to import meals, and share a single meal or activity as a QR code another phone can scan. Your profile, custom-meal catalogue, water and fasting history stay out of the bundle.
 
 </details>
 
@@ -131,7 +131,7 @@ No account, no sign-in, no analytics, no ads. Your profile, diary, activities, w
 | Destination | When | What is sent |
 | :-- | :-- | :-- |
 | [Open Food Facts](https://world.openfoodfacts.org/) | Food search, barcode scan, or whenever a food's photograph is shown — in search results, on a meal screen, and on the diary card of a meal you have already logged | The search term or barcode. A word search also sends your device's **language** code to rank results; the fallback search and the barcode lookup send no locale at all. Your **country is never sent** — the country boost is applied on your device, to the results that come back. A photograph's address contains that product's barcode. Searches and product lookups identify the app by name, platform and version; photograph downloads do not |
-| Supabase reference backend | Food search | The text you typed; the device language too, when it is one this Application has food translations for; and, if any food database is off, the list of the ones left on (onboarding already turns BLS off outside DE/AT/CH, and US branded products off inside them; Settings can change that) |
+| Supabase reference backend | Food search | The text you typed; your **device's** language code, when it is one the backend has food names in; and which food databases you have left switched on — that list travels on every search, because onboarding starts one of the five off |
 | [Anthropic](https://www.anthropic.com/) | **Only if you save your own API key and pick Anthropic** | The meal line you type on the multi-item add screen, or a meal photo you choose to read there, and your app language |
 | [OpenAI](https://openai.com/) | **Only if you save your own API key and pick OpenAI** | As the Anthropic row, plus the model you chose |
 | [OpenRouter](https://openrouter.ai/) | **Only if you save your own API key and pick OpenRouter** | As the Anthropic row, plus the model you chose |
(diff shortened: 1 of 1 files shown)


simonoppowa:
@codex review

simonoppowa:
Thanks for this — it makes the README more accurate on balance, and some of it fixes errors that had been there a while. The export bullet in particular: weight log and custom activity templates were listed as excluded when the code exports both, and custom meals were listed as included when there is no custom-meal dataset in the archive at all. Those were real and you got them right.

There is **one line** I'd like changed, and it happens to cover two separate inaccuracies.

## The Supabase row

Two problems, both in the new text:

**It is the device language, not the app language.** `SpFoodDataSource.fetchSearchWordResults` builds the `loc` parameter from `Platform.localeName` (`sp_food_data_source.dart:43`), which is the *device* locale — the in-app language picker lives in `ConfigRepository` / `LocaleProvider` and nothing on this path reads it. So if someone runs the phone in English and picks German in the app, Supabase still receives English.

This is a genuinely easy trap, because **"your app language" is correct on the AI rows** — those use `Localizations.localeOf(context).languageCode`. Two different sources, one page apart.

**The food-source list travels on every search, not only after a change in Settings.** `defaultFoodSourceToggles` (`locale_units.dart:81`) always leaves one of the five off — German speakers lose `fdc_branded`, everyone else loses `bls` — and onboarding persists that on first run. So the filter list is sent even for someone who has never opened that screen.

Suggested replacement for the whole row:

| Supabase reference backend | Food search | The text you typed; your **device's** language code, when it is one the backend has food names in; and which food databases you have left switched on — that list travels on every search, because onboarding starts one of the five off |

## Not yours to fix

An automated review also flagged the own-server row's bold **"Only if you save a server address in Settings"** as contradicting the trigger list that follows it. The contradiction is real — `_endpointCommitted` fires on focus loss and reaches `_loadModels()` while the dialog is still open, so a request can leave before anything is saved — but **that bold text is pre-existing and unchanged by your PR**. You inherited it. I'll fix it separately; mentioning it only so the flag doesn't look like something you missed.

The same review questioned the export/photo claims. I checked and it was wrong — your text is accurate there, and no change is needed.

Once the Supabase row is updated this is good to merge.

AzazelSensei:
Updated the four review notes. The Supabase row now says device language, and that onboarding already turns some sources off. The own-server row no longer claims the models list waits for Save. Export now says one-off custom-meal photos stay out of the zip.

AzazelSensei:
Swapped the Supabase row for your wording. Put the own-server and export photo lines back.

simonoppowa:
Follow-up on the inherited sentence: fixed in #1060, so it is off your plate.

For the record, the code says requests leave **before** any save — `_endpointCommitted` is wired to the address field's focus listener (`ai_assist_dialog.dart:239`) and to `onSubmitted` (`:889`), and its early returns cover Cancel/OK/back and an unchanged address but not moving focus to another field in the open dialog. So changing the address and tabbing away reaches `_loadModels()` and contacts the machine.

**One heads-up:** #1060 edits the same table cell you did, so once it lands you will see a conflict on that row. The two changes are complementary rather than competing — keep #1060's bold phrase ("From the moment you type a server address in Settings, not from when you save it") and your trigger additions. Sorry for the extra step; better than leaving an inaccurate privacy claim on `develop` while this PR is open.

The Supabase row is still the only thing needed from you.

AzazelSensei:
That's already in the latest commit — device language and the onboarding source filter, in your wording. I'll keep #1060's bold phrase when it lands.

simonoppowa:
Resolved the conflict on your branch rather than leaving it with you — `maintainerCanModify` is on, and #1060 had just landed.

It is a merge commit, not a rebase, so none of your commits were rewritten. One cell, and both halves are taken **verbatim**; there is no new text in it:

| | from |
| :-- | :-- |
| **When** | develop (#1060) — *"Only if you save a server address"* was wrong, and that is the row inviting readers to falsify it against their own access log |
| **What is sent** | **yours** — the only place disclosing the models-list request, and the fixed example line and bundled photograph the setup check sends on save |

Neither side alone makes that row correct, which is why it was worth keeping both.

**One thing I did not touch, and would rather you decide.** Your *What is sent* cell says the models-list request goes out *"when you finish with the address"*, while the *When* cell beside it now says *"when you change the address and leave the field"*. The difference is real: `_endpointCommitted` returns early when the resolved address is unchanged, so re-focusing the field without editing sends nothing. Yours is very slightly the looser of the two, in the row where looseness is most visible.

It is one word and I did not want to put words in your cell. Add "change the" if you agree; if you would rather leave it, that is fine too — it is a nuance, not an error.

Thanks for the four corrections in here. The export bullet and the Open Food Facts photograph row were long-standing and nobody had caught them.

AzazelSensei:
Added it — models-list only when you change the address, same as the When cell.

Issue #1022, opened 2026-09-02
Ship and measure the concurrency group + path gate


simonoppowa:
## Question

Ship the concurrency group and the ubuntu path gate, then measure whether p90 actually moved.

This is the one ticket on the map that **executes** rather than decides — see the execution override in the map's Notes. It carries the two low-risk changes, because their entire value is empirical and cannot be settled on paper.

Scope:

- Add the `concurrency` group decided in the concurrency ticket.
- Add an ubuntu change-detection job and gate `ios-build` / `ios-podfile-lock-guard` on it, using the filter decided in the paths ticket and the reporting shape decided in the skip-propagation ticket.
- **Not** in scope: relocating the iOS boot smoke. That ships as a spec for a follow-up PR.

Done when:

- The change is merged, and
- The baseline method from the measurement ticket has been re-run against post-merge runs, with the before/after numbers recorded here.

A merged PR with no measurement does not resolve this ticket — the map's destination is a number, not a diff.

simonoppowa:
**Progress: change written, draft PR open. Not done — this ticket needs a merge and a measurement.**

Draft PR: #1043 (`ci/gate-macos-jobs-on-changed-paths` → `develop`). Carries the #1018 concurrency group and the #1019 path gate with #1021's job shape. One commit, +72/−1, `default_workflow.yml` only.

**First run is green on the gate.** Run [33752008062](https://github.com/simonoppowa/OpenNutriTracker/actions/runs/33752008062):

- `changes` completed in **4 seconds** (11:51:53 → 11:51:57), ubuntu, no checkout — the paths-filter action uses the REST API on `pull_request` events, which is also why it works for fork PRs.
- `ios-build` and `ios-podfile-lock-guard` **queued rather than skipped**, which is correct for this PR and confirms the gate emitted `true`.

**This PR cannot demonstrate its own main effect.** It edits `.github/workflows/**`, so the filter matches by design and the gate opens. Observing the *skip* path needs a docs-only PR afterwards.

**Deliberately not included:** #1020's relocation of the iOS boot smoke to `push: main`. It edits `ios-package`/`android-package` `needs:` and deserves separate review. Per the map's execution override it ships as a spec, not from this ticket.

## Remaining before this ticket can close

[...]

Issue #1021, opened 2026-09-02
How do path-gated macOS jobs report into the needs chain without re-breaking skip propagation?


simonoppowa:
## Question

How do path-gated macOS jobs report into the `needs` chain without re-breaking skip propagation?

This is the highest-risk ticket on the map. A job skipped by an `if:` gate taints every job that `needs` it — exactly the failure mode fixed twice in the last two days (a8a78997 "stop a skipped retry job taking the whole deploy chain with it", 5c8794c8 "let a build-only bump ship, and gate publishing separately"). Path gating reintroduces skipped jobs on the main path, so it walks straight back into that bug.

Resolve:

- What the downstream chain (`ios-integration-tests-result`, `release-gate`, the package and deploy jobs) sees when `ios-build` is skipped because no iOS files changed, versus when it genuinely failed. These must not look the same.
- Whether the gate belongs as a job-level `if:` (produces a skipped job) or as step-level `if:`s inside a job that always runs (produces a green job that did nothing). The second is uglier but does not propagate a skip.
- Whether `ios-integration-tests-result`'s "empty output counts as not passed" logic stays correct when its `needs` were skipped rather than run.
- Whether any `needs` condition still relies on implicit success and must be made to assert each need explicitly, per the pattern established in #1008.

Blocked by the paths ticket: the exact filter determines which jobs can be skipped and therefore which chain edges are at risk.

simonoppowa:
**Context handed over from [#1019](https://github.com/simonoppowa/OpenNutriTracker/issues/1019)'s verification pass (2026-09-03).** This ticket is now unblocked, and the pass turned up the specific trap it exists to avoid.

**"Make the gate return `true` on non-PR events" is necessary but NOT sufficient.** GitHub ANDs an implicit `success()` onto any job carrying a `needs:` edge, so a gate job that is **skipped or failed** re-skips `ios-build` regardless of what its output says — and `ios-package` / `android-package` both assert `needs.ios-build.result == 'success'`. That is precisely the 2.2.0 outage (run `33547853809`, fixed in a8a78997): every direct need was green, the `if:` evaluated true, and the chain died anyway.

**The obvious fail-open expression is wrong:**

```yaml
# WRONG — `!= 'success'` also fires on `skipped`, flipping everything open by accident
if: ${{ !cancelled() && (needs.changes.result != 'success' || needs.changes.outputs.ios == 'true') }}

# RIGHT
[...]

Issue #1020, opened 2026-09-02
Where does the relocated iOS boot smoke run, and what enforces the release gate?


simonoppowa:
## Question

Where exactly does the relocated iOS boot smoke run, and what enforces "green before a release ships"?

The map settles that the smoke stops being a per-PR check and becomes a release gate. This ticket decides the mechanism.

- Which triggers: `pull_request` against `release/**`, `push: main`, a schedule, or some combination? A `release/**` PR is the last point where a failure is still cheap to act on; `push: main` is after the fact.
- What actually *enforces* it? Today nothing is a required status check (`main` has no `required_status_checks`; `develop` has it enabled with empty `contexts`). Options: turn on a required check scoped to release branches, make `release-gate` depend on the smoke's verdict, or leave it advisory and rely on the human release runbook.
- Does the existing `release-gate` → package → deploy chain need to consume `ios-integration-tests-result`, and if so how does that interact with the skip-propagation work in #1008?
- Does the two-runner retry survive the move? On a release branch the run is rarer and latency matters less, so the serial retry costs little — but confirm rather than assume.
- What does the PR check surface look like afterwards, so a contributor is not confused by an iOS check that simply is not there?

simonoppowa:
## Answer

**The iOS boot smoke moves to `push: main` only, as an advisory signal that blocks nothing.** Concretely:

1. **Trigger:** remove it from `pull_request`; run on `push: main`. PRs get no iOS integration check at all — `android-integration-tests` on ubuntu is the per-PR signal.
2. **Enforcement: none.** Remove `needs.ios-integration-tests-result.result == 'success'` from `ios-package` and `android-package`. A red or hung smoke must no longer be able to strand a release.
3. **Failure handling:** a red smoke on main **auto-opens an issue** (deduplicated, so a repeated failure reopens rather than spams).
4. **Keep the two-runner retry.** Its job changes from "don't block a PR falsely" to "don't file a false issue".

> **This revises the charting-time decision.** The map originally settled on *"must be green before a release ships"*. That was chosen believing the smoke was catching real iOS-only breaks. It is not, and the evidence below is why the bar moved to an advisory post-merge signal.

## Evidence

### The smoke has never caught anything

**Category (b) — a runtime boot break caught by the iOS integration suite and by nothing else — has happened ZERO times** in 732 PR runs across 342 distinct head branches over 120 days (2026-05-05 → 2026-09-02).

[...]

Issue #1019, opened 2026-09-02
Which changed paths must wake the iOS jobs, including the fork-PR case?


simonoppowa:
## Question

Which changed paths must wake the iOS jobs, and how is the filter computed for fork PRs?

The map settles that `ios-build` and `ios-podfile-lock-guard` stay per-PR but only run when iOS-relevant files change. This ticket decides the filter itself.

- What set of paths genuinely implies "iOS could break"? `ios/**`, `pubspec.yaml`, `pubspec.lock` are obvious; `lib/**` is the hard case — a Dart change can break an iOS build through plugin usage, but treating all of `lib/**` as iOS-relevant gates out almost nothing.
- Which side should the default fall on when the filter is ambiguous? Running a Mac unnecessarily costs queue time; skipping one wrongly lets an iOS break reach `develop`.
- `ios-podfile-lock-guard` already uses `dorny/paths-filter` with `base: github.event.pull_request.base.ref` — but it runs that filter *after* macos-26 is allocated (it waited 42 min in queue to decide it had nothing to do). The gate must be computed on an ubuntu job instead. Confirm the filter still works when hoisted.
- Fork PRs check out differently (read-only token, `refs/pull/N/head` mirror). Does the change-detection job have the base ref in scope for a fork PR, and what does it do if not?
- Does `workflow_dispatch` (no PR context, so no diff) need an explicit "run everything" path?

simonoppowa:
## Answer

**Filter: `ios/**`, `pubspec.yaml`, `pubspec.lock`, `.github/workflows/**`, `.github/actions/**`, `.github/scripts/**`.** Computed by a cheap **ubuntu** job whose output gates `ios-build` and `ios-podfile-lock-guard`.

Gates out **92 of 117 recent merged PRs (78.6%)** and saves ~19% of all billed macOS minutes.

**The gate must return `true` for every non-`pull_request` event.** This is not a preference — see the deploy-chain hazard below.

## Evidence

### Safety is not the discriminator

Over 4 months (2026-05-05 → 2026-09-02) there were **43** `ios-build` / `ios-podfile-lock-guard` failures. This filter would have gated out **4** of them, and **not one was a genuine iOS-only regression**. Every genuinely iOS-exclusive failure in the period — 6 `Build iOS (no codesign)` breaks and 3 `Pod install` resolution failures, all with `linux-checks` and `android-build` green — occurred on a PR touching `ios/**` or `pubspec`. **9 for 9.**

[...]

PR #1043, opened 2026-09-03
ci: cancel superseded runs and gate the macOS jobs on changed paths

simonoppowa:
Part of the CI-latency map, #1016. Implements the decisions from #1018, #1019 and #1021.

**Draft on purpose** — see *Verifying this* below. Nothing here should be merged before #1014 lands.

## Why

The pipeline is not slow, it **queues**. Every ubuntu job finishes inside ~10 min; runs stretch to 43–96 min only when several PRs are in flight and the macOS jobs wait for a runner. `ios-integration-tests` does 20–24 min of work whether the run takes 20 minutes or 96.

Nothing is billed — the repo is public, so standard runners are free. The scarce resource is the **hard cap of 5 concurrent macOS jobs**, which is account-wide, plan-invariant across Free/Pro/Team, and was fully binding during the 2026-09-01 burst: 5 running, 25 queued, and every one of the five was this repo's own work.

| | p50 | p90 |
|---|---|---|
| uncontended runs (n=236) | 18.0 min | 24.2 min |
| contended runs (n=56) | 27.8 min | 51.8 min |

19.2% of runs land in the contended stratum.

## What this does

**1. A `concurrency` group.** A force-push stops holding macOS runners for a commit nobody is waiting on. 18.2% of PR runs (150 of 824) are superseded by a newer push to the same branch, and 82% of the macOS time that reclaims is the iOS integration suite.

`main` is excluded deliberately. Those runs carry the deploy chain, and a cancel between a store upload completing and the ledger tag being pushed consumes a TestFlight build number with nothing to show for it. A PR can never share a group with a push to `main` anyway (`refs/pull/N/merge` vs `refs/heads/main`), so the guard is really about a *second push to main* — which has happened once in 26 main pushes.

⚠️ **Do not add a `concurrency:` block to `ios-integration-attempt.yml`.** `${{ github.workflow }}` evaluates to the *caller's* name inside a reusable workflow, so it would land in this same group and the run would cancel itself.

**2. A `changes` gate on the two macOS PR jobs.** Gates out **92 of 117** recent PRs (78.6%) and cuts hours pinned at the 5-slot cap from **5.5 → 0.6 (−89%)**.

Safety is not the trade-off here. Over four months, **all 23 genuine iOS-only failures touched `ios/**` or `pubspec.*` — the filter catches 23 of 23.** Two deliberate choices:

- **`lib/**` excluded.** Including it drops gating to 29.9% while guarding a failure mode with zero recorded instances. The one `Build iOS` break on a non-iOS diff (#410) was caught first by `linux-checks` and `android-build`.
- **`.github/**` included.** Costs ~11 percentage points, and it is honestly insurance rather than a historical catch — the record gives it zero. It is there so the pipeline keeps verifying itself while the rest of #1016 changes it.

Note the filter uses `ios/**` as a glob, **not** the guard's existing three narrow paths. Those miss both genuine iOS-only breaks on record: #368 (`ios/Runner/AppDelegate.swift`) and #448 (Xcode project, a fork PR).

## The part that is easy to get wrong

`changes` carries **no `if:`**. A skip travels the whole dependency chain rather than one hop, and a job that exempts itself with `!cancelled()` does not stop it continuing downstream — that is exactly what shipped an empty 2.2.0 (run 33547853809, fixed in #1008). A gate that skipped on `push: main` would plant a fresh skip source directly upstream of the deploy chain, so it runs on every event and simply reports `true` outside a pull request.

Both gates **fail open**, spelled `== 'failure'` and never `!= 'success'` — the latter is also true for `skipped`, which would flip the gate open by accident. Same trap #1008 documents for `!= 'failure'`.

`ios-package` and `android-package` are **untouched**; they already mirror every dependency explicitly.

## Verifying this

**This PR cannot demonstrate its own main effect.** It edits `.github/workflows/**`, so the filter matches, the gate opens and the macOS jobs run exactly as before. That is the correct behaviour and it exercises the fail-open path — but observing the *skip* path needs a subsequent docs-only PR.

Worth checking on this run:
- [ ] `changes` runs and reports `ios: true`
- [ ] `ios-build` and `ios-podfile-lock-guard` still run (gate open)
- [ ] the deploy chain is unaffected — all five jobs skip on a PR exactly as they do today
- [ ] then, on a docs-only PR: `changes` reports `false`, both macOS jobs skip, and the run still reports success

## Not in this PR

Relocating the iOS boot smoke to `push: main` (#1020) ships separately as a spec — it touches `ios-package`/`android-package` `needs:` and deserves its own review.

## Measurement

#1022 is not done when this merges. The baseline to compare against is recorded in #1017: uncontended p90 **24.2 min** [22.8, 26.0] and contention rate **19.2%** [15.1, 24.1], re-derived with the recipe in that ticket.

---

## Hazards found by verification after this PR opened

**1. Fixed in `98e0d5e0`.** The first cut spelled fail-open as `needs.changes.result == 'failure' || ...`, which is false for `skipped` — so it failed *closed* on a skipped gate, stranding the deploy chain exactly as run 33547853809 did. The gate has no `if:` so a skip is unreachable today, but that made the safety a property of a comment rather than of the expression. Now inverted: skip only when the gate ran cleanly and said no.

| gate result | output | before | after |
|---|---|---|---|
| success | `'true'` | RUN | RUN |
| success | `'false'` | SKIP | SKIP |
| failure | `''` | RUN | RUN |
| **skipped** | `''` | **SKIP** | **RUN** |

**2. Do not make `ios-build` or `ios-podfile-lock-guard` a required status check.** `develop` already has required-status-checks machinery enabled with an empty list (`{"strict":false,"contexts":[],"checks":[]}`), so adding one is a two-click operation. A path-filtered job that skips **never posts a check run**, so a required context on it leaves the PR pending forever — and 72% of recent PRs (42 of 58 sampled) would skip `ios-build`. If a required iOS check is ever wanted, require the *gate* job, which always runs.

**3. Merge-order note.** This branch and #1015 both fork from `a809810e` and both add a gate job upstream of the deploy chain. They touch disjoint line regions so git will merge them cleanly — which is precisely why a weaker guard here would not have been noticed at merge time. With `98e0d5e0` the two guards now have the same shape: `release-gate` asserts `result == 'success' && outputs.deploy == 'true'`, and `changes` skips only on `result == 'success' && outputs.ios == 'false'`.

---

### `main` is serialised, not excluded — please sanity-check this is what you want

`cancel-in-progress: false` does **not** take `main` out of the concurrency group. It puts it in the group without cancelling, which per the docs means:

> *"When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be `pending`."*

and

> *"any existing `pending` job or workflow in the same concurrency group will be canceled and the new queued job or workflow will take its place."*

So a second push to `main` while the first is still deploying now **waits** instead of running alongside — and if a *third* push arrives while the second is still pending, **the second is cancelled and replaced**.

For a deploy chain, serialising is arguably the safer behaviour: two concurrent deploy runs racing on the same build number and ledger tag is worse than one waiting. But the displacement case means a middle commit's deploy run can be dropped, and that is a genuine behaviour change this PR did not previously spell out.

Historically it would have bitten rarely: 26 pushes to `main` in 143 days, one overlapping pair (21 s apart, on 2026-08-27). If displacement is unwanted, `concurrency.queue: max` preserves up to 100 pending runs in FIFO order instead — it is mutually exclusive with `cancel-in-progress`, so it would mean splitting the config by event.

### Where the macOS time actually is — this PR is the smaller half

Re-derived over the same 19-day window (7,767 PR-event macOS slot-minutes):

| Job | share of PR macOS time | gated by this PR? |
|---|---|---|
| `ios-integration-tests` (+ retry, + reusable form) | **70.8%** | ✗ — that is #1020 |
| `ios-build` | 26.9% | ✓ |
| `ios-podfile-lock-guard` | 2.3% | ✓ |

Gating 78.6% of `ios-build` + guard reclaims **23.0%** of PR macOS slot-time. The concurrency group reclaims roughly another 18% by cancelling superseded runs, and the two overlap.

The honest framing: **this PR is the smaller half of the win.** The iOS integration suite is 71% of the budget and is untouched here — relocating it (#1020) is where the bulk of the contention relief comes from.

*(An intermediate analysis put this PR's share at 2.8%; that was wrong by roughly 8x — it mistook the guard's own 2.3% share for the whole gate's benefit. The 23.0% above is re-derived directly from the job dump.)*

### DIFF
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -46,7 +46,19 @@ on:
 # get superseded, and 82% of the macOS time that reclaims is the iOS
 # integration suite.
 #
-# `main` is excluded deliberately: those runs carry the deploy chain, and a
+# `main` is never CANCELLED, but note it is not excluded from the group either:
+# `cancel-in-progress: false` puts it in the group without cancelling, which
+# per the docs means a second push to `main` goes PENDING until the first run
+# finishes — and if a third arrives while the second is still pending, the
+# second is cancelled and replaced. So `main` is SERIALISED, with displacement.
+#
+# That is the intended trade: two concurrent deploy runs racing on the same
+# build number and ledger tag is worse than one waiting. It is a real behaviour
+# change though — before this block, two main pushes ran side by side. If
+# displacement is ever unwanted, `concurrency.queue: max` keeps up to 100
+# pending runs in FIFO instead, and is mutually exclusive with cancel-in-progress.
+#
+# Why cancelling `main` would be worse: those runs carry the deploy chain, and a
 # cancel mid-upload can consume a TestFlight build number with no ledger entry.
 # A pull request can never share a group with a push to `main` anyway —
 # `github.ref` is `refs/pull/N/merge` for a PR and `refs/heads/main` for the
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -41,7 +41,69 @@ on:
       - 'release/**'
   workflow_dispatch:
 
+# Supersede in-flight runs for the same ref, so a force-push stops holding
+# macOS runners for a commit nobody is waiting on any more. 18.2% of PR runs
+# get superseded, and 82% of the macOS time that reclaims is the iOS
+# integration suite.
+#
+# `main` is excluded deliberately: those runs carry the deploy chain, and a
+# cancel mid-upload can consume a TestFlight build number with no ledger entry.
+# A pull request can never share a group with a push to `main` anyway —
+# `github.ref` is `refs/pull/N/merge` for a PR and `refs/heads/main` for the
+# push — so this guard is about a *second push to main*, which has happened
+# once in 26 main pushes. See #1018.
+#
+# Do NOT add a `concurrency:` block to ios-integration-attempt.yml:
+# `${{ github.workflow }}` evaluates to the CALLER's name inside a reusable
+# workflow, so it would land in this same group and the run would cancel itself.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
 jobs:
+  # Decides whether the macOS jobs below need to run at all. Cheap, on ubuntu,
+  # and deliberately carries NO `if:` — it must run on EVERY event.
+  #
+  # A skip travels the whole dependency chain, not one hop (#1008): a job that
+  # exempts itself with `!cancelled()` does not stop the skip continuing
+  # downstream. A gate that skipped on `push: main` would therefore plant a
+  # fresh skip source directly upstream of the deploy chain, six days after
+  # exactly that shipped an empty 2.2.0. Fifteen free ubuntu seconds removes
+  # the question. See #1021.
+  #
+  # No checkout: dorny/paths-filter calls the REST API on `pull_request`
+  # events and never touches git, which is also why it works on fork PRs.
+  changes:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      # Evaluates to the string 'true' or 'false', never empty. Written as a
+      # boolean expression rather than `steps.filter.outputs.ios || 'true'`
+      # on purpose: that form relies on the string 'false' being truthy (it
+      # is — only '' is falsy), and if that were ever wrong the gate would
+      # silently never close and this whole change would be a no-op that
+      # still looked healthy. The non-PR arm short-circuits first, so the
+      # skipped `filter` step's empty output is never consulted.
+      ios: ${{ github.event_name != 'pull_request' || steps.filter.outputs.ios == 'true' }}
+    steps:
+      - name: Check for iOS-relevant file changes
+        id: filter
+        if: github.event_name == 'pull_request'
+        uses: dorny/paths-filter@v4
+        with:
+          filters: |
+            ios:
+              - 'ios/**'
+              - 'pubspec.yaml'
+              - 'pubspec.lock'
+              # The pipeline verifies itself while we change it: an edit to the
+              # workflow or its composite actions can break the iOS jobs.
+              - '.github/workflows/**'
+              - '.github/actions/**'
+              - '.github/scripts/**'
+
   linux-checks:
     runs-on: ubuntu-latest
     permissions:
@@ -111,8 +173,11 @@ jobs:
   # pubspec.yaml, pubspec.lock, or ios/Podfile actually changed —
   # otherwise the macOS minutes aren't worth burning.
   ios-podfile-lock-guard:
+    needs: changes
+    # Zero `needs:` dependents, so gating this one carries no propagation risk
+    # at all. Same fail-open shape as ios-build.
+    if: ${{ !cancelled() && github.event_name == 'pull_request' && (needs.changes.result == 'failure' || needs.changes.outputs.ios == 'true') }}
     runs-on: macos-26
-    if: github.event_name == 'pull_request'
     permissions:
       contents: read
     steps:
@@ -183,6 +248,12 @@ jobs:
           fi
 
   ios-build:
+    needs: changes
+    # Fail OPEN. `== 'failure'`, never `!= 'success'` — the latter is true for
+    # `skipped` too and would flip the gate open by accident (the same trap
+    # #1008 documents for `!= 'failure'`). `!cancelled()` drops the implicit
+    # success() so a broken gate can never withhold the build.
+    if: ${{ !cancelled() && (needs.changes.result == 'failure' || needs.changes.outputs.ios == 'true') }}
     runs-on: macos-26
     permissions:
       contents: write
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -397,13 +397,17 @@ jobs:
   # a verdict, and ios-integration-tests-result below turns the pair into the
   # single required check.
   ios-integration-tests:
+    needs: changes
+    if: ${{ !cancelled() && !(needs.changes.result == 'success' && needs.changes.outputs.ios == 'false') }}
     uses: ./.github/workflows/ios-integration-attempt.yml
 
   # Only when the first runner did not pass. An empty output counts as "did not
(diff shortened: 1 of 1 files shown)


simonoppowa:
### Verified against the new `develop` (after #1014 landed)

`release-gate` is now on `develop`, and it rewrites the same `ios-package` / `android-package` blocks this branch touches. Git auto-merges the two cleanly — **which is exactly the situation where a wrong result would go unnoticed** — so I ran a real trial merge of `develop` + #1044 + #1043 and audited the outcome rather than trusting the absence of conflict markers.

**Result: sound.** 16 jobs, and every deploy-chain job keeps both guards:

| Job | `needs:` | smoke still referenced | unmirrored needs |
|---|---|---|---|
| `ios-package` | release-gate, linux-checks, ios-build, android-build, android-integration-tests | no | none |
| `android-package` | *(identical)* | no | none |
| `ios-deploy` | release-gate, ios-package, android-package | no | none |
| `android-deploy` | release-gate, android-package, ios-package | no | none |
| `github-release` | release-gate, ios-deploy, android-deploy | no | none |

Every entry is mirrored in its `if:`, per this file's own rule.

**On the two gate jobs now sitting upstream of the deploy chain.** An earlier analysis warned that landing both would give the chain "TWO skippable gates with opposite failure polarity". The polarities *are* opposite, and that is correct — they gate opposite things:

- **`release-gate`** carries `if: github.ref == 'refs/heads/main' && (push || workflow_dispatch)`, so it **is** skipped on pull requests. Its consumers assert `needs.release-gate.result == 'success'`, so a skip closes the chain — right, because a PR must never deploy.
- **`changes`** carries no `if:` at all and cannot be skipped. Its consumers fail **open**, so a dead gate can never withhold a build — right, because withholding `ios-build` is what strands a release.

Each gate fails in the direction that is safe for what it protects. The danger would be either one adopting the other's polarity.

simonoppowa:
### Copilot triage — 1 fixed, 1 declined

**✅ Fixed (`default_workflow.yml:55`) — "the comment claims `main` is *excluded*".** Correct, and a fair catch. `cancel-in-progress: false` does not remove `main` from the group; it puts it in without cancelling, so a second push to `main` goes **pending**, and a third **displaces** the pending second. `main` is *serialised, with displacement* — not excluded.

I had already worked this out and written it into the PR description, but never corrected the comment in the file, which is the copy a future editor actually reads. That is the same defect class flagged on #1014's `outputs:` block, and this file has already shown that a stale comment here becomes load-bearing. Corrected, with the `concurrency.queue: max` alternative noted in case displacement is ever unwanted.

**❌ Declined (`default_workflow.yml:194`) — fork checkout in `ios-podfile-lock-guard`.** Two reasons:

1. **This PR does not touch that block.** The diff adds `needs:` and `if:` to the job; the `actions/checkout` step is unchanged from `develop`. Nothing here alters fork behaviour.
2. **The evidence contradicts the concern.** Across the last 60 workflow runs the guard is **55 success / 3 skipped / 0 failures** — it has never failed at checkout, on a fork PR or otherwise. Reading a *public* fork needs no authentication, which is why this works; `ios-build` uses the `refs/pull/N/head` mirror for a different reason, stated in its own comment — it carries `contents: write` and pushes a lockfile commit back, which genuinely cannot authenticate against a fork.

If the guard's checkout is worth hardening it should be its own change, judged against its own evidence, rather than folded into a contention fix.

simonoppowa:
Pushed `b16f929f`.

`ios-integration-tests` was ungated — the two-line reusable-workflow call had neither `needs: changes` nor an `if:`, while `ios-build` and `ios-podfile-lock-guard` both had the full treatment. That left the **longest** macOS job running on every PR, which is the one #1016 measured at 42 minutes queued for 22 minutes of work.

Gating it alone would have been worse than leaving it. Two knock-ons had to move with it:

- **The retry would have run anyway.** It fires on `needs.ios-integration-tests.outputs.passed != 'true'`, and a *skipped* job leaves that output empty — empty is not `'true'`, so it would have claimed a macOS runner on exactly the runs the gate had just spared. It now carries the same path gate, ANDed with the verdict test.
- **`ios-integration-tests-result` would have gone red on every non-iOS PR.** It exits 1 when neither output says `'true'`; with the suite skipped, both are empty. That is the job your own comment names as *the check to require in branch protection*. It now takes `changes` as a dependency and returns early with a log line, rather than passing silently.

Deploys are untouched, and deliberately so: the package jobs depend on `ios-integration-tests-result`, never on the suite directly, and on a push to `main` the `changes` job short-circuits `ios` to `'true'` before the filter is consulted — so the release chain runs exactly as before. Checked each of the five deploy gates still mirrors every `needs:` entry with a `result == 'success'` line, per the rules in that block's comment.

Found this while looking at why three macOS jobs queued on #1045, a PR touching only `AGENTS.md`, `justfile`, two new docs files and three `.github` files.

simonoppowa:
### Copilot triage, round 2 — both accepted, one change resolves both

**✅ `default_workflow.yml:118` — the guard boots a Mac for iOS-code-only PRs.** Correct, and it is the exact waste this PR exists to remove, so merging it would have been poor. An `ios/Runner/*.swift`-only PR opened the broad `ios` gate, allocated a macos-26 runner, ran the guard's own narrow filter, learned the lockfile could not have moved, and skipped every remaining step — roughly **11 seconds of a scarce macOS slot** to answer a question ubuntu had already been asked.

Taken as suggested: `changes` now exports a second, narrower **`pod_deps`** output (`pubspec.yaml`, `pubspec.lock`, `ios/Podfile`), and the guard gates on that instead. Its in-job `paths-filter` step and the seven step-level `if:`s it drove are deleted — the job now only starts when the answer is already yes.

Worth noting this was independently flagged by the verification pass on #1021 as "NEW WASTE (small)", so two separate reviews landed on it.

**✅ `default_workflow.yml` — the guard uses `paths-filter` without `pull-requests: read`.** Resolved by the same change rather than by adding the permission: the guard **no longer runs `paths-filter` at all**. The `changes` job that does call `pulls.listFiles` already declares `pull-requests: read`.

For the record, the concern was theoretical here — the guard is 55 success / 3 skipped / **0 failures** across the last 60 runs, because reading a public repo's PR files needs no special scope. But removing the call is strictly better than granting a permission for a call that did not need to exist.

**Behaviour is unchanged.** The set of PRs on which `pod install` runs is identical to before. What changed is that deciding it no longer costs a macOS runner.

Issue #1016, opened 2026-09-02
Shorten the CI pipeline: cut p90 wall clock under runner contention


simonoppowa:
## Destination

> **✅ REACHED 2026-09-04.** Both changes are on `develop`; macOS queue p90 fell 35.4 -> 1.7 min and demand fell 3.06 -> 2.17 jobs/run. See [#1022](https://github.com/simonoppowa/OpenNutriTracker/issues/1022) for the measurement and its caveats. Everything in **Not yet specified** below is unstarted and would be a fresh effort.

Cut the Default Workflow's **p90 wall clock under runner contention** by removing macOS demand from ordinary PRs.

Non-iOS PRs touch zero Macs. `ios-build` and `ios-podfile-lock-guard` stay per-PR but gated on changed paths. The iOS boot smoke moves from a per-PR check to a **release gate** on `release/**` and `main`, with `android-integration-tests` carrying the per-PR integration signal.

Ends with the concurrency + path-gating change **merged and measured**, and the iOS relocation written up as a **spec** for a follow-up PR.

## Notes

**Domain.** GitHub Actions CI for a Flutter app — `.github/workflows/default_workflow.yml` (77 KB, 14 jobs) plus the reusable `.github/workflows/ios-integration-attempt.yml`.

**Measured baseline** (last ~40 `Default Workflow` runs, 2026-08-31 → 09-02): 15 min best case, 43–96 min when several PRs are in flight. Every ubuntu job finishes by ~10 min (`android-integration-tests` ~10 min, `linux-checks` ~4 min). The entire tail is macOS **queueing**, not macOS work. Run `33529672680` is the clearest case: created 16:05:01, `ios-integration-tests` started 16:47:18 — **42 min queued**, then 22 min of work; `ios-build` queued 44 min.

**What that macOS time buys.** `integration_test/` is a single file, `app_boot_test.dart`, with one `testWidgets` — a first-run boot smoke. `android-integration-tests` runs the *identical* suite (`flutter test integration_test/ --flavor full`) on an ubuntu emulator in ~10 min. Every Dart-level assertion (routing, bloc state, l10n, `FlutterError.onError`) is covered there; iOS uniquely covers the iOS plugin-init / platform-channel surface at boot.

[...]

Issue #541, opened 2026-07-18
[Bug] Nothing is being saved


jluims:
I have a OnePlus 12R running LineageOS and the app seems to work well; I had it setup and logged some food but upon waking up, the app was back to its initial state.

I found that any time the app closes on my device/is no longer running in the background, it goes back to factory settings leading me to believe that nothing is being saved. 

The app is version 2.0.0.

Version 1.5.0 seems to save just fine.

jluims:
Seems that upon reinstalling the app, it now saves my data just fine.

simonoppowa:
That sounds very strange. I will definitely investigate that before rolling out 2.0 to the stores. Maybe a problem with the new profiles setup. Thanks for reporting that @jluims

PR #1044, opened 2026-09-04
ci: make the iOS boot smoke an advisory signal on main

simonoppowa:
Implements #1020, from the CI-latency map #1016. **Draft** — this touches the deploy chain's `needs:`, which is the riskiest code in the file.

## What changes

| | before | after |
|---|---|---|
| iOS boot smoke | every PR + push | **`push: main` / dispatch only** |
| Gates the deploy chain | yes (`ios-package`, `android-package`) | **no — removed from `if:` and `needs:`** |
| A red smoke | red X, blocks the release | **files/updates an issue** |
| Two-runner retry | kept | kept |
| Per-PR integration signal | iOS + Android | **Android only** (identical test, free ubuntu) |

## Why

In **732 PR runs across 342 branches over 120 days**, this suite was never the unique detector of anything. Its 25 PR-run failures:

| Class | n | Also caught by |
|---|---|---|
| Compile / build breaks | 12 | `ios-build`, every one |
| Podfile.lock drift | 1 | the guard **and** `ios-build` |
| Cross-platform dependency breaks | 3 | `linux-checks` **and** `android-build` |
| Infrastructure (flake) | 8 | — |

Counting cancellations, **60% of its non-green events were infrastructure**, 21% genuine breaks. On `push: main` it has run **29 times and never gone red**.

Meanwhile it is the largest single consumer of the account-wide **5-slot macOS cap** — 71% of PR macOS time, and 4 of the 5 slots held during the 2026-09-01 contention incident that produced 96-minute runs.

The sharpest single data point: **#541** — *"nothing is being saved"* across restarts, exactly the Hive-init class this smoke was written to catch. It shipped. On Android, where the identical suite runs on every PR and did not catch it.

## The event test is on all three jobs, and load-bearing on each

This is the part that is easy to get wrong, and getting it wrong turns **every PR red**:

- **`ios-integration-tests-retry`** — its condition is `passed != 'true'`, and a skipped job's outputs are the **empty string**. `'' != 'true'` is TRUE, so scoping only the first attempt would fire the retry on every pull request.
- **`ios-integration-tests-result`** — carries `!cancelled()`, which exempts it from the skip its needs inherit. Without its own event test it would **run** on PRs with both outputs empty, read that as "did not pass", and `exit 1`.

## The veto is removed from `needs:` as well as `if:`

Leaving it in `needs:` would violate the rule this file states on `ios-package` — *"every entry in `needs:` needs a matching line in `if:`. The implicit check is gone; an unmirrored dependency is an ungated dependency."* It would also keep packaging waiting ~18–24 min for an answer nobody consults, so **`push: main` runs get faster** as a side effect.

A flaky simulator can no longer strand a release. Its only organic two-runner verdict on record — run [33561715901](https://github.com/simonoppowa/OpenNutriTracker/actions/runs/33561715901), a CI-YAML-only diff with `ios-build`, `linux-checks`, `android-build` and Android integration all green — hung on both runners and announced *"treat this as a real break, not a flake."* It was a false alarm. One for one, wrong.

## `ios-smoke-alarm` replaces the veto

An advisory check nobody reads is worse than no check, because it reads as coverage — this repo already has one of those in `android-deploy`, which reports green while Play received nothing. So a red smoke on `main` files an issue, or comments on the open one.

Deduplicated by **exact title** via `gh issue list`, not `--search`: the search index lags by seconds to minutes and would file duplicates across consecutive red runs. The issue body tells the reader how to distinguish a VM-service hang from a real break before they act on it.

## Verifying this

Unlike #1043, **this PR demonstrates its own main effect** — it is a pull request, so the smoke should skip:

- [ ] `ios-integration-tests`, `-retry`, `-result` all **skipped**
- [ ] no red X, and the run reports success
- [ ] `android-integration-tests` still runs and is the integration signal
- [ ] the deploy chain still skips on a PR, as it always did
- [ ] after merge, on the next `push: main`: the smoke runs, and packaging no longer waits for it

## Also in here

One sentence in the deploy-guard rules block claimed a bare `!cancelled()` would push a failed integration suite straight to TestFlight. It no longer can — the suite is absent from `needs:`. Re-anchored to the consequence that is still true (deploying off a failed `ios-build`), with a note on why it changed. The 2.2.0 history and the four rules are untouched and still correct.

## Not in here

The `ios-integration-tests-result` job's own `needs:` are unmirrored in its `if:` — it reads their **outputs** in shell rather than asserting their **results**, which is the job's entire purpose. Pre-existing on `develop`, deliberate, and left alone.

## Relationship to the other open PRs

Zero merge conflicts with #1043 or #1014, re-checked after each commit. Independent of both; can land in any order.

### DIFF
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -294,13 +294,30 @@ jobs:
   # a verdict, and ios-integration-tests-result below turns the pair into the
   # single required check.
   ios-integration-tests:
+    # Not on pull requests. In 732 PR runs over 120 days this suite was never
+    # the unique detector of anything: every genuine break it caught was also
+    # red on ios-build, linux-checks or android-build in the same run, while
+    # 60% of its non-green events were infrastructure. It is also the single
+    # largest consumer of the 5-slot macOS cap — 71% of PR macOS time — so it
+    # is what makes PRs queue. android-integration-tests runs the identical
+    # `flutter test integration_test/` on a free ubuntu emulator and remains
+    # the per-PR signal. See #1020.
+    if: ${{ github.event_name != 'pull_request' }}
     uses: ./.github/workflows/ios-integration-attempt.yml
 
   # Only when the first runner did not pass. An empty output counts as "did not
   # pass": that is what a job killed by its own timeout-minutes leaves behind.
   ios-integration-tests-retry:
     needs: ios-integration-tests
-    if: ${{ !cancelled() && needs.ios-integration-tests.outputs.passed != 'true' }}
+    #
+    # The event test is NOT redundant with the one above. This job's condition
+    # is `passed != 'true'`, and a skipped job's outputs are the empty string —
+    # `'' != 'true'` is TRUE — so without it, skipping the first attempt on a
+    # pull request would fire the retry on every pull request instead.
+    if: >-
+      ${{ !cancelled()
+      && github.event_name != 'pull_request'
+      && needs.ios-integration-tests.outputs.passed != 'true' }}
     uses: ./.github/workflows/ios-integration-attempt.yml
 
   # The check to require in branch protection — the two jobs above are green
@@ -310,7 +327,11 @@ jobs:
     needs:
       - ios-integration-tests
       - ios-integration-tests-retry
-    if: ${{ !cancelled() }}
+    # Same event test again, and load-bearing for the same reason: `!cancelled()`
+    # exempts this job from the skip its needs inherit, so on a pull request it
+    # would RUN with both outputs empty, read that as "did not pass", and exit 1
+    # — a red X on every PR for a suite that deliberately did not run.
+    if: ${{ !cancelled() && github.event_name != 'pull_request' }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


simonoppowa:
### Verified against the new `develop` (after #1014 landed)

`release-gate` is now on `develop`, and it rewrites the same `ios-package` / `android-package` blocks this branch touches. Git auto-merges the two cleanly — **which is exactly the situation where a wrong result would go unnoticed** — so I ran a real trial merge of `develop` + #1044 + #1043 and audited the outcome rather than trusting the absence of conflict markers.

**Result: sound.** 16 jobs, and every deploy-chain job keeps both guards:

| Job | `needs:` | smoke still referenced | unmirrored needs |
|---|---|---|---|
| `ios-package` | release-gate, linux-checks, ios-build, android-build, android-integration-tests | no | none |
| `android-package` | *(identical)* | no | none |
| `ios-deploy` | release-gate, ios-package, android-package | no | none |
| `android-deploy` | release-gate, android-package, ios-package | no | none |
| `github-release` | release-gate, ios-deploy, android-deploy | no | none |

Every entry is mirrored in its `if:`, per this file's own rule.

**On the two gate jobs now sitting upstream of the deploy chain.** An earlier analysis warned that landing both would give the chain "TWO skippable gates with opposite failure polarity". The polarities *are* opposite, and that is correct — they gate opposite things:

- **`release-gate`** carries `if: github.ref == 'refs/heads/main' && (push || workflow_dispatch)`, so it **is** skipped on pull requests. Its consumers assert `needs.release-gate.result == 'success'`, so a skip closes the chain — right, because a PR must never deploy.
- **`changes`** carries no `if:` at all and cannot be skipped. Its consumers fail **open**, so a dead gate can never withhold a build — right, because withholding `ios-build` is what strands a release.

Each gate fails in the direction that is safe for what it protects. The danger would be either one adopting the other's polarity.

simonoppowa:
### Copilot triage — fixed

**✅ Fixed (`default_workflow.yml:377`) — the alarm's title says "on main" but it could fire elsewhere.** Correct, and a genuine defect I introduced.

`ios-smoke-alarm` was conditioned on `github.event_name != 'pull_request'`, which also matches a `workflow_dispatch` against **any** ref. Dispatching the suite on a feature branch — to check a simulator fix, say — would have filed an issue asserting `main` was broken. Worse, because the dedup keys on the **exact title**, a branch run would have commented on the open `main` thread and folded unrelated refs together, which is precisely the failure the dedup exists to prevent.

Now gated on `github.ref == 'refs/heads/main'` as well.

Note what deliberately did **not** change: `ios-integration-tests` still runs on a dispatch against any ref. Running the suite manually on a branch is useful; it is only *raising an alarm about main* from such a run that is wrong. The suite's own condition and the alarm's are intentionally different scopes now, and both carry comments saying why.

Issue #1032, opened 2026-09-02
Protect the Actions secret store before installing any write-capable app


simonoppowa:
Part of #1023

## Question

**What has to be true about the Actions secret store before any write-capable GitHub App is installed here?**

The map recorded a decision resting on the premise that *a GitHub App cannot read Actions secrets*. That premise is true in the narrow sense and **false in the sense that matters**. All of the following are verified first-hand, not inferred:

**The Jules App's actual permission record** (`GET /apps/google-labs-jules`, record updated 2026-04-15):

```
contents: write   workflows: write   actions: write
pull_requests: write   issues: write
administration: read   members: read   emails: read
security_events: read   checks: read
```

`checks: read` confirms a check run is impossible. `workflows: write` + `actions: write` + `contents: write` is the problem: GitHub's own hardening guidance states that any actor with write access to a repository has read access to all of its secrets — because that actor can author and run a workflow that prints them.

**This repository has nothing standing in the way of that:**

| Control | State |
|---|---|
| `main` required approving reviews | **0** |
| `main` required status checks | **none** |
| `enforce_admins` | **false** |
| Environments `.env`, `copilot`, `github-pages`, `tofu-apply` | `protection_rules: []` — all four |
| Secret scope | **repo-level**, not environment-scoped |

**And the secret store is far larger than this map assumed.** It is not only the store signing material. All 26 are repo-level:

- Release signing — `ANDROID_KEYSTORE_BASE64`, `ANDROID_KEY_PASSWORD`, `APP_STORE_CONNECT_API_KEY_P8`, `GOOGLE_PLAY_SERVICE_ACCOUNT_JSON`, `MATCH_*`
- **Infrastructure** — `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`, `CLOUDFLARE_API_TOKEN`, `CLOUDFLARE_PURGE_TOKEN`, `R2_ACCESS_KEY_ID` / `R2_SECRET_ACCESS_KEY`
- **Terraform state and DNS control** — `TF_GITHUB_TOKEN`, `TF_VAR_STATE_PASSPHRASE`, `TF_VAR_ACCOUNT_ID`, `TF_VAR_ZONE_ID_OPENNUTRITRACKER_ORG`
- Other tokens — `PROJECTS_TOKEN`, `CATALOG_ACCESS_TOKEN`

[...]

Issue #1031, opened 2026-09-02
Ship it: merge the configuration, review a real PR, judge the result


simonoppowa:
Part of #1023

## Question

**Ship it, review a real PR, and judge the result.**

The one ticket on this map that carries execution rather than a decision. Everything upstream is settled by the time this runs; this is where it meets a real contributor's code.

Do:

- Land whatever configuration the map decided on — steering file, trigger, scope rule — as a PR, conventional-commit titled, following the repo's own conventions. No AI co-author trailer, no assistant named in the subject or body.
- Point it at a real, currently-open outside PR: [#1010](https://github.com/simonoppowa/OpenNutriTracker/pull/1010) (`AzazelSensei`, a README change) or [#995](https://github.com/simonoppowa/OpenNutriTracker/pull/995) (`ibrahim-iqbal`, a real Flutter change touching search-state UI — the better test, since it should trip the `Semantics(identifier:)` convention if anything does).
- Read the review as the contributor would read it, not as its author.

Then judge, and write the verdict down:

- **Did it catch the conventions?** Specifically the ones no generic reviewer would know. If it missed them, the steering is wrong and that is a finding, not a failure.
- **How much of it was noise?** Count the findings worth acting on against the total.
- **Would this have helped or discouraged the contributor?**
- **What did it cost** in quota, against what #1026 established.

Verdict is **keep / tune / drop**, stated plainly. "Tune" must name what to tune. Reaching a verdict completes the map — including a verdict of "drop", which is a real outcome and not a wasted effort.

simonoppowa:
**Two prerequisites surfaced while resolving #1029** — both are execution work that belongs with the ship, not separate efforts.

**1. `.github/PULL_REQUEST_TEMPLATE.md:35` contradicts what Codex will say.** It reads *"Localization updated in ARBs **and** `lib/generated/` when user-facing strings change"* — but `lib/generated/` is gitignored (`.gitignore:42`) and `CONTRIBUTING.md` forbids hand-editing it. The new review rules explicitly tell the reviewer that the absence of generated Dart is *correct*. Ship both together or a contributor gets a checklist and a bot review that disagree on the same PR. Issue #578's acceptance criteria carry the same stale instruction.

[...]

Issue #1029, opened 2026-09-02
What the review checks, how loud it is, and what fits in 3,087 bytes


simonoppowa:
Part of #1023

## Question

**What does the review check, how loud is it — and what fits in 3,087 bytes?**

Largely tool-independent as originally written, and #1035 proved the substance is right. What #1034 added is a **hard byte budget**, which turns this from an editorial question into an engineering one.

### The budget

Codex reads `AGENTS.md` on the review path, wants a `## Code Review Rules` heading, and **truncates at 32 KiB silently** (openai/codex#13386, open).

```
AGENTS.md   29,681 bytes
cap         32,768 bytes  (32 KiB)
headroom     3,087 bytes  — 90.6% consumed
```

There is no `## Code Review Rules` heading today, and adding one costs from that 3,087. The file has grown steadily — docs-correction PRs land most weeks — so this will breach on its own, quietly, with no error and no signal beyond reviews mysteriously getting worse.

Decide:

- **Does the review section fit in the headroom, or does `AGENTS.md` get trimmed / split?** Codex merges instruction files root→leaf, so a nested `AGENTS.md` may carry per-directory rules — worth establishing whether that relieves the root file.
- **Ordering as insurance.** If truncation cuts the tail, then position is protection: the non-negotiables belong early in the file, not wherever they landed historically. Cheap, and it survives the file growing.
- **A guard against silent breach.** A CI check failing when `AGENTS.md` exceeds ~31 KiB would turn a silent degradation into a visible one. This repo already runs `just check_l10n` for an analogous quiet failure, so the pattern exists.

### What the rules should say

From #1035, the conventions a review must enforce — with the evidence that they matter:

- **Real translations, not English placeholders.** The single most valuable catch in the trial: three of four new strings left verbatim English across eight locales, against `CONTRIBUTING.md:54`, and structurally invisible to CI because `just check_l10n` only fails on *missing* keys.
- **ARB key parity** — nine locales at equal count, **discounting template-only `@key` metadata**. State the exception explicitly or a reviewer will report a false violation (raw counts legitimately differ: en 1011, de 986, cs 982, all at 946 real keys).
- **`Semantics(identifier:)`** on new interactive widgets, kebab-case, with the `container: true` rule stated as *conditional on a layout-greedy parent* — the trial showed a good reviewer reasons about this rather than pattern-matching it.
[...]

Issue #1028, opened 2026-09-02
Codex reviews every PR including mine — is that acceptable, and what is the trigger surface?


simonoppowa:
Part of #1023

## Question

**Codex reviews every PR, including mine. Is that acceptable — and what is the actual trigger surface?**

Rewritten after #1034 chose Codex. The original question was how to express `author != simonoppowa`. **That is not expressible**: Codex has no author filter, and the only author-scoped setting is the inverse ("Review my pull requests"). So the scope decision this map opened with is reversed, and the real question is what to do about the overlap.

Decide:

- **Is reviewing my own PRs tolerable, or actively bad?** `/code-review` already runs locally on my branches, so this is duplicated work on the *majority* of traffic — I author most PRs here. It costs quota (shared 5-hour rolling windows with all other Codex usage) and adds a second opinion I did not ask for on work I already reviewed. Against that: a second pass is not worthless, and there is no way to switch it off per-author anyway.
- **The trigger surface.** Establish precisely what the connected-repo settings expose: the "Automatic reviews" toggle, and the per-repo `Auto code review` / `Review trigger` overrides. Is automatic-on-open all-or-nothing, or is there a middle setting?
- **Mention-only as a deliberate choice.** If automatic proves too noisy or too costly, `@codex review` on demand is the fallback — but it surrenders the unattended property that justified this whole map. Decide what would make that trade worth it.
- **Draft PRs** — reviewed or not.
- **Dependabot.** 11 open, all targeting `develop`, all created 2026-08-21 to 08-30. A bump diff has little to read and Dependabot rebases often. Does Codex review them, can it be stopped from doing so, and does each rebase burn another review?
- **Backfill.** #1010 and #995 are open and unreviewed. Does enabling automatic reviews reach existing open PRs, or only new ones?

Quota context from #1026's sibling research: reviews draw the same budget as all other Codex usage on 5-hour rolling windows; at the limit reviews simply stop, and Plus can buy credits.

simonoppowa:
**A case this ticket should cover, surfaced while resolving #1029.**

Since Codex has no author filter, it will also review **your own `release/* → main` PRs** — the most sensitive ones in the repo. The review rules written in #1029 are deliberately contributor-shaped: a five-finding ceiling, "write for that person", observations rather than demands. That is the wrong posture for a release PR, where you want completeness over gentleness and there is no stranger to protect.

[...]

PR #1045, opened 2026-09-04
feat(ci): give Codex review rules to read, and stop AGENTS.md outgrowing them

simonoppowa:
## Summary

Prepares the repo for Codex PR review, which reads `AGENTS.md` to know this project's conventions — and stops that file quietly outgrowing the limit Codex reads it under.

Refs #1031, #1029, #1028.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] CI / tooling
- [ ] Localization
- [ ] Other

## Changes

**A `## Code Review Rules` section in `AGENTS.md`.** Its content is not guesswork — it is shaped by a hand-run review of #995 that established what a good review of this repo has to do:

- **The one check CI cannot do.** `just check_l10n` fails only on *missing* keys, so a key present in every locale with the English text still sitting in it ships green. #995 did exactly that with three of four new strings, against `CONTRIBUTING.md:54`, and passed CI.
- **Two rules that exist to prevent confident false findings.** ARB key counts legitimately differ between *any* two locales because every file carries its own `@key` metadata (en 79, de/sk/tr 42, zh 39, cs/it/pl/uk 38) — parity means non-`@` keys, and those are equal. And `container: true` is *conditional* on a layout-greedy parent, so its absence under a `Wrap` is correct.

**Room made for it.** Codex truncates instruction files at 32 KiB, keeping the head and dropping the tail, with no warning anywhere a human looks. `AGENTS.md` was at 90.6% of that. Two blocks no reviewer can act on move out, each leaving a pointer:

| moved to | what | why |
|---|---|---|
| `tools/adb/README.md` | ADB driver scripts, the `content-desc` gotcha, form-field helpers | device operating guidance — a reviewer has no device |
| `docs/demo-data.md` | the demo seeder and "try it" onboarding flow | authoring orientation, not diff-checkable |

**`AGENTS.md`: 29,681 → 27,776 bytes**, rules at byte 653.

**`just check_agents_md`** fails the build *before* the cap is reached (31,000, leaving 1,768 of margin) and asserts the rules stay near the head where truncation cannot take them. Pure bash, so it runs first in `linux-checks` and needs no toolchain — no cost against the macOS queueing tail.

**Dependabot updates grouped.** Four ecosystems ran weekly with `open-pull-requests-limit: 10` and no grouping, fanning out together — 11 PRs were open at once. Each ecosystem now collapses to one PR.

**PR template corrected.** It asked for localization updates "in ARBs **and** `lib/generated/`", which is gitignored and which `CONTRIBUTING.md` forbids editing by hand — a contributor following the checklist could not comply.

## Screenshots / recordings
n/a — no UI changes.

## Test plan
- [ ] Manual check on Android

### Steps
1. `wc -c AGENTS.md` → 27,776; heading offset 653 (`grep -b -m1 '^## Code Review Rules$'`).
2. `just check_agents_md` passes.
3. Guard's **failure** paths tested against copies, not just the happy path: an oversized file fails on size (31,776 > 31,000), and a renamed heading fails on the missing-heading branch.
4. `.github/dependabot.yml` and `.github/workflows/default_workflow.yml` both re-parse as YAML.
5. `## Commands` still carries `flutter test test/unit_test/tdee_calc_test.dart` and `flutter analyze` — the trim range originally planned would have deleted them.

## Checklist
- [x] New interactive widgets include `Semantics(identifier: '...')` where needed — n/a
- [x] Localization updated in ARBs — n/a, no user-facing strings

## Note for review

This does **not** install anything. The Codex GitHub App requests `contents`, `workflows` and `actions: write`, and #1032 established that scoping secrets without a required reviewer buys nothing against that — an actor holding `workflows: write` can add `environment:` to a job it writes. The `release-build` / `release-ship` environments must exist first.

The `environment:` keys on `android-package`, `ios-package`, `ios-deploy` and `android-deploy` are deliberately **not** in this PR: adding them before the environments exist would create them implicitly with no protection, so the guard would look present and do nothing.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -32,7 +32,7 @@
 ## Checklist
 - [ ] Code follows project style (`just format` / 120-char line width)
 - [ ] New interactive widgets include `Semantics(identifier: '...')` where needed
-- [ ] Localization updated in ARBs **and** `lib/generated/` when user-facing strings change
+- [ ] Localization updated in **every** `lib/l10n/intl_*.arb` with a real translation, not the English string (`lib/generated/` is gitignored — do not edit or commit it)
 - [ ] Codegen ran if DBOs/DTOs/env changed (`just build`)
 - [ ] No secrets or `.env` values committed
 - [ ] PR title follows conventional commit style (e.g. `feat:`, `fix:`, `chore:`)
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -30,7 +30,7 @@
 2.
 
 ## Checklist
-- [ ] Code follows project style (`just format` / 120-char line width)
+- [ ] Code follows project style — run `just format` (`dart format`'s default 80-column width; nothing configures a wider one)
 - [ ] New interactive widgets include `Semantics(identifier: '...')` where needed
 - [ ] Localization updated in **every** `lib/l10n/intl_*.arb` with a real translation, not the English string (`lib/generated/` is gitignored — do not edit or commit it)
 - [ ] Codegen ran if DBOs/DTOs/env changed (`just build`)
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,10 @@ updates:
       interval: "weekly"
     target-branch: "develop"
     open-pull-requests-limit: 10
+    groups:
+      ios-bundler:
+        patterns:
+          - "*"
 
   - package-ecosystem: "bundler"
     directory: "/android"
@@ -14,6 +18,10 @@ updates:
       interval: "weekly"
     target-branch: "develop"
     open-pull-requests-limit: 10
+    groups:
+      android-bundler:
+        patterns:
+          - "*"
 
   - package-ecosystem: "pub"
     directory: "/"
@@ -21,6 +29,10 @@ updates:
       interval: "weekly"
     target-branch: "develop"
     open-pull-requests-limit: 10
+    groups:
+      pub:
+        patterns:
+          - "*"
 
   - package-ecosystem: "github-actions"
     directory: "/"
@@ -28,3 +40,7 @@ updates:
       interval: "weekly"
     target-branch: "develop"
     open-pull-requests-limit: 10
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -73,6 +73,12 @@ jobs:
       # period when CI was disabled. We run the rest of `just ci` (l10n
       # generation, build_runner, analyze, test) and leave the format pass
       # for a dedicated follow-up PR.
+      # AGENTS.md is read by Codex on the PR review path and truncated at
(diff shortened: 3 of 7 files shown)


simonoppowa:
Added `500f2faa` — the last copy of the 120-column claim.

#1047 corrected `AGENTS.md` and `CONTRIBUTING.md`, but `.github/PULL_REQUEST_TEMPLATE.md:33` still said `just format` / 120-char line width. That is the worst of the three places to leave it wrong: it is the checklist a contributor reads *while deciding how to format*, whereas the other two are reference docs they may never open.

Kept out of #1047 deliberately because this PR already edits that file, and two open PRs touching one line would have conflicted.

Verified the merge of this branch into current `develop` is semantically correct, not just conflict-free — `git merge-tree` produces an `AGENTS.md` with **0** occurrences of `120-character`, **1** of the new 80-column text, and all three of this PR's structural changes intact (the `## Code Review Rules` heading plus both extraction pointers). The guard still passes on the merged result: **28,116 bytes**, 2,884 under the 31,000 limit, rules at byte 653.

simonoppowa:
Fixed in `5990e7b5` — good catch, and a real defect rather than a nit.

The demo-data section came out of `AGENTS.md` at the repo root, where `[`lib/core/utils/demo/`](lib/core/utils/demo/)` resolved correctly. From `docs/demo-data.md` that same relative target points at `docs/lib/core/utils/demo/`, which does not exist. Now `../lib/core/utils/demo/`.

Since this is a class of error rather than a one-off — moving a file without re-basing its relative links — I audited both extracted files instead of just patching the flagged line:

| file | relative links | status |
|---|---|---|
| `docs/demo-data.md` | 1 | was broken, now fixed |
| `tools/adb/README.md` | 0 | nothing to break |

And confirmed the pointers left behind in `AGENTS.md` still resolve (`docs/demo-data.md`, `tools/adb/README.md`, `docs/export-format.md`) — they stayed at the repo root, so their targets are unchanged.

simonoppowa:
Both of the latest review's points addressed — `78e5e15f` and merge `81fec971`.

**`AGENTS.md:15` — stale command list.** Correct, and it was a self-inconsistency: the Code Review Rules tell the reviewer *not to restate what CI already failed on*, then enumerate what `linux-checks` runs — and this PR adds `check_agents_md` to that very job. Left as-is, the reviewer would have been instructed to comment on a size breach CI had already gone red on. The list now reads `just check_agents_md`, `just check_l10n`, `just build`, `flutter analyze`, `just test`, matching the job's actual step order.

**`.github/PULL_REQUEST_TEMPLATE.md:33` — internal inconsistency.** Also correct, and the diagnosis is exact: this branch merged `develop` at `3d742070`, *before* #1047 landed the 80-column correction, so within the branch the template said 80 while `AGENTS.md` still said 120. Merged current `develop` (`81fec971`), which brings the fix in. `grep -rn "120-char"` across `AGENTS.md`, `CONTRIBUTING.md` and the template now returns nothing.

Guard re-verified on the merged result: **28,140 bytes**, 2,860 under the 31,000 limit, `## Code Review Rules` at byte 653 — passes both conditions.

Worth noting the second finding was only visible because the reviewer read the branch as a whole rather than the diff in isolation. The diff alone looks consistent; the contradiction only exists between a changed file and an unchanged one.

Issue #942, opened 2026-08-28
Play's API rejects health-permission bundles regardless of the declaration


simonoppowa:
2.1.0 is the first build that declares Health Connect permissions. `git` puts all five in #651 (`af28c39f`, 2026-08-21), which is not an ancestor of `v2.0.2`:

```
android.permission.health.READ_EXERCISE
android.permission.health.READ_TOTAL_CALORIES_BURNED
android.permission.health.READ_DISTANCE
android.permission.health.READ_STEPS
android.permission.health.READ_BODY_FAT
```

The Play Console's **Health apps** declaration (App content, last edited 28 Apr 2024) asks only about **feature categories** — `Aktivität und Fitness` and `Ernährungs- und Gewichtskontrolle` are ticked, which is correct. Google's own page describes something more than that: declaring **uses of the data types**, with a justification per permission.

> This process must be completed for all publishing requests, both for a new app that has not been published yet, or when updating an existing, already published app that now uses a different set of data types.

— [Request access to Health Connect data types](https://developer.android.com/health-and-fitness/guides/health-connect/publish/request-access)

No such per-data-type section renders on the form today. The likely reason is that it only appears once a bundle carrying `android.permission.health.*` reaches a track, and none ever has. That cannot be confirmed before the 2.1.0 bundle is uploaded, which is why this is a separate ticket rather than part of [#941](https://github.com/simonoppowa/OpenNutriTracker/issues/941).

## Why this blocks the release

If the section exists and is left unfilled, the build still publishes — and the health import fails silently for every user:

> If your health app is published in the Play store and released to the public, but you didn't request for data type accesses, your end users receive the following dialog when attempting to link with Health Connect

That is #651's entire feature dead on arrival with correct code behind it, and nothing in CI or in the app can detect it.

## What to do

After uploading the 2.1.0 bundle to a track and **before promoting it to production**:

1. Reopen **App content → Gesundheits-Apps**. Check whether a per-data-type section has appeared under step 1.
2. If it has, declare the five permissions above with a justification each. They are all read-only, on-device, and feed the workout/activity import only — the app holds `HealthDataAccess.READ` and there is not one WRITE call in the repo.
3. If it has not, install the track build on a device, open Settings → health import, and confirm the Health Connect link actually succeeds rather than showing the "can't access Health Connect" dialog. That is the only real proof.

[...]

PR #1053, opened 2026-09-04
ci: assert the release happened, instead of reporting the colour

### LINKED ISSUE #1012: CI: a release that ships nothing reports success
Merging #988 to `main` produced **no release**, and the pipeline reported `success`.

Run [33547853809](https://github.com/simonoppowa/OpenNutriTracker/actions/runs/33547853809): every check green, conclusion `success`, and `ios-package`, `ios-deploy`, `android-package`, `android-deploy` and `github-release` all skipped. No `v2.2.0` tag, no GitHub release, nothing to TestFlight, no AAB. Nothing anywhere said so — the run looked exactly like a successful release.

The cause is fixed (#1008, `a8a78997`), but the *silence* is a separate defect and it will outlive that fix. Any future condition that skips the chain — a new `needs:` entry, another conditional job, a guard edit — fails the same way: green, quiet, and empty.

There is a second flavour of the same problem already shipping: `android-deploy` deliberately exits 0 when Play rejects the upload (#942). It emits a `::warning::`, which is easy to miss on a green run, and Play genuinely does not have the bundle afterwards. Confirmed again on the v2.2.0 run [33554300243](https://github.com/simonoppowa/OpenNutriTracker/actions/runs/33554300243).

## What would have caught it

A final assertion step that checks the *artifact*, not the job colour. Options, roughly in order of effort:

1. **`github-release` asserts the tag now exists** at the expected version, and fails loudly if not. Cheap, catches the exact 2.2.0 failure.
2. **A `release-summary` job with `if: !cancelled()`** that runs regardless, reads the outcome of each deploy job, and fails when a push to `main` produced no release. Catches skips too, because it does not sit downstream of them in a way that inherits the skip.
3. **Assert the version was actually published** — query the Play and App Store Connect APIs for the build number and fail if absent. Most thorough, most moving parts, and would also surface the #942 gap as a real failure rather than a warning.

(1) plus (2) would cover both flavours without touching the deploy jobs.

## Why it is worth doing

Two consecutive releases were lost. v2.1.0 to a genuine `android-deploy` failure, v2.2.0 to a silent skip. In both cases the gap between "pipeline finished" and "users can install it" was invisible from the run page — and in the 2.2.0 case, invisible for as long as nobody thought to check `gh release list`.

simonoppowa:
Closes #1012.

Two consecutive releases were lost behind a green run. v2.1.0 died on a real `android-deploy` failure; v2.2.0 died because a skipped job propagated a skip down the whole `needs:` chain — every packaging, deploy and release job skipped, nothing reached a store, no tag was cut, and the run reported `success`. **Nothing downstream of a skipped job can report on it, because it inherits the skip.**

## `release-summary`

`needs:` all five release jobs plus `release-gate`, guarded by `!cancelled()` so it inherits nothing, and judges every case explicitly from `needs.*.result`.

A bare `!cancelled()` is *wrong* in the five deploy gates — the comment there explains it would push an integration suite that failed on two independent runners straight to TestFlight — and it is exactly right here, where running after failures and skips is the entire point. It asserts that what `release-gate` promised was actually produced, and stays quiet on a push that was never meant to ship.

## `github-release` asserts the release exists

The action reporting success is not the same as a release existing: on an already-published tag it logs, returns, and stays green over assets from a different build. The new step asks the remote, checks the tag **points at this run's commit** — names alone cannot tell the two cases apart, because the names are identical — and checks all three binaries are attached.

## `android-deploy` publishes whether Play took the bundle

This is the one place I departed from the issue, which hoped to cover both flavours "without touching the deploy jobs". That is not possible honestly: `android-deploy` tolerates one Play error (#942), exits 0 with the bundle rejected, and pushes the `deployed/<n>` ledger tag either way — so nothing downstream could distinguish *Play has it* from *Play refused it*. A `play_uploaded` output is the smallest thing that can.

The result is that this lands on the run page instead of in a log nobody opens on a green run:

> **The Android bundle is not on Play.** … `android-deploy`'s result says only that the bundle was built and the error was the expected one — never that Play accepted it.

## What the review changed

Five reviewers attacked the first draft; eight findings survived judging, all fixed. The two that mattered most were both cases of my breaking discipline this file already established:

- **Both new remote lookups used `if ! git ls-remote --exit-code`**, collapsing "no such tag" (exit 2) into "could not reach the remote" (128). `release-gate` spends a comment block on exactly this, and its `ref_exists()` cases the three answers separately. A transient failure would have reddened a fully published release with the words *"Nothing was published"*. Both call sites now use the gate's form.
- **`release-summary` said "There is no release" without asking the remote** — the check sat in an `elif` that only ran when `github-release` had already succeeded. But `action-gh-release` creates the release *before* uploading assets, so a mid-upload failure leaves a real, partial release. The old wording would have sent the operator to re-run the job, which renames the published release and appends a second copy of the notes.

Also fixed: a red deploy job no longer claims *"No store received this build"* (the ledger step runs after the upload, so it may well have); the #942 warning no longer suppresses itself when `android-deploy` goes red, which is the run where it matters most; the closing line no longer says the run "reported success" on runs that are loudly red; `release-gate`'s summary no longer claims nothing depends on `github-release`; and the job declares `permissions: contents: read` rather than being the only steps-bearing job without a block.

## Verification

Both `run:` blocks were extracted from the YAML and executed against **fourteen scenarios**: full release; the exact v2.2.0 skip shape (fails); an ordinary docs merge (passes); gate failed (passes, asserts nothing); gate green with empty outputs (fails); build-only bump (passes); one deploy job failed (fails); a transient `ls-remote` exit 128 (fails, and says so correctly); a release job red with the tag present, and with it absent (two different messages); #942 with `android-deploy` green and with it red; and for the assertion step, a tag at this run's commit, a tag at a different commit, an absent tag, and an unreachable origin.

The tag-identity check was confirmed against the real `v2.2.0`, whose tag points at its run's `head_sha` as a lightweight ref.

`docs/RELEASING.md` gains the post-run step and a note that `release-summary` repeats the Play warning whatever colour `android-deploy` ended up.

### DIFF
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -1823,8 +1823,21 @@ jobs:
           # something else entirely when the question was never answered, and
           # a transport failure must never be reported as "nothing was
           # published". An earlier draft of this step collapsed them.
+          # Both refs, on purpose. An ANNOTATED tag's own ref points at the
+          # tag object, not at the commit — the commit sits behind the peeled
+          # `^{}` ref, and an exact pattern does not return that unless it is
+          # asked for by name (a bare `--tags` would list both; this does not).
+          # So asking only for `refs/tags/$TAG_NAME` and comparing to
+          # $GITHUB_SHA would compare a tag-object sha to a commit sha and
+          # always differ.
+          #
+          # `action-gh-release` creates a lightweight ref through the Releases
+          # API, so today only the plain line comes back and the peeled one
+          # does not exist. This does not rely on that staying true: correctness
+          # here should not depend on an implementation detail of a third-party
+          # action, because the failure mode is every release going red.
           set +e
-          ref=$(git ls-remote --exit-code --tags origin "refs/tags/$TAG_NAME" 2>/dev/null)
+          ref=$(git ls-remote --exit-code --tags origin "refs/tags/$TAG_NAME" "refs/tags/$TAG_NAME^{}" 2>/dev/null)
           ls_status=$?
           set -e
 
@@ -1846,7 +1859,10 @@ jobs:
           # different build. Names alone cannot tell the two apart, because the
           # names are identical. The commit can: a release this run published
           # is tagged at this run's commit.
-          tag_sha=${ref%%[[:space:]]*}
+          tag_sha=$(printf '%s\n' "$ref" | awk -v t="refs/tags/$TAG_NAME" '
+            $2 == t "^{}" { peeled = $1 }
+            $2 == t       { plain  = $1 }
+            END           { print (peeled != "" ? peeled : plain) }')
           if [ "$tag_sha" != "$GITHUB_SHA" ]; then
             echo "::error::$TAG_NAME points at $tag_sha, not this run's commit ($GITHUB_SHA). The tag existed before this job ran, so the assets on it belong to a different build and this run published nothing. Do not re-run this job — see the overwrite_files note above."
             exit 1
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -1961,19 +1961,43 @@ jobs:
           fi
 
           if [ "$DEPLOY" = true ]; then
+            BUILD_NUMBER=${APP_VERSION##*+}
+            ledger_advice=''
+
             for pair in "ios-package:$IOS_PACKAGE" "android-package:$ANDROID_PACKAGE" \
                         "ios-deploy:$IOS_DEPLOY" "android-deploy:$ANDROID_DEPLOY"; do
               job=${pair%%:*}
               res=${pair##*:}
-              if [ "$res" != success ]; then
-                # NOT "no store received this build". Both deploy jobs end with
-                # the ledger push, which runs AFTER the upload — so a red
-                # deploy job may well mean the store has the build and only the
-                # bookkeeping failed. Saying otherwise would invite a re-upload
-                # that Apple rejects and Play cannot accept twice.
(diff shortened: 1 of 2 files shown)


simonoppowa:
Fixed in `79410952`. The conclusion was right and worth acting on; the mechanism described is not what happens, and the difference changes the fix.

**The suggested fix would not have worked.** The comment says `git ls-remote --tags` "can return two lines for an annotated tag" and to "prefer the peeled `refs/tags/<tag>^{}` SHA when present". A bare `--tags` does return both, but this call passes an **exact ref pattern**, and `git ls-remote` matches a pattern against the tail of a ref from the start or a slash — so `refs/tags/vANN` does not match `refs/tags/vANN^{}`. Reproduced against a real annotated tag:

```
$ git ls-remote --exit-code --tags origin refs/tags/vANN
ec35f19c731b57ae3ba302da35e9df48cdfb0813	refs/tags/vANN     <- the tag OBJECT
$ git rev-parse HEAD
5ce96ddb6652f57f8102b8652932bb4d5ef3d920                       <- the commit
```

One line, not two, and it is the tag object. There is no peeled line to prefer, so the defect is real but a line-selection fix cannot reach it — both refs have to be requested by name. They now are, with the peeled one preferred when present.

Verified across four cases: an annotated tag now resolves to its commit (previously a false failure), a lightweight tag still does, a missing tag still exits 2, and the real `v2.2.0` still resolves to its run's `head_sha` so the happy path is unchanged.

Worth noting why this was worth fixing at all, since it is unreachable today: `action-gh-release` creates a lightweight ref through the Releases API, so only the plain line ever comes back. The check should not depend on that — it is an implementation detail of a third-party action, and the failure mode if it ever changes is every release going red.

Issue #1054, opened 2026-09-04
Finish the Codex PR-review setup: environments, install, settings, backfill, verdict


simonoppowa:
Everything in the repository is ready for Codex PR review — `## Code Review Rules` is in `AGENTS.md` at byte 653, the `check_agents_md` guard is in CI, Dependabot is grouped. **Nothing is installed and no review has run.**

This is the remaining checklist. Every step needs a Google/OpenAI account or a secret value, so none of it is delegable — hence no `ready-for-agent`.

Background and reasoning: closed map #1023, decisions on #1032 (secrets), #1034 (tool choice), #1028 (trigger policy), #1029 (the rules).

---

## 1. Environments first — this is a hard gate

**Do not install the Codex App before this.** It requests `contents: write`, `workflows: write` and `actions: write`, and #1032's finding is that scoping secrets without a **required reviewer** buys nothing: an actor holding `workflows: write` can add `environment:` to a job it writes. The reviewer is the control, because **a GitHub App cannot approve a deployment gated on a human.**

Current state (verified 2026-09-04):

```
main            required_approving_review_count: 0
environments    .env, copilot, github-pages, tofu-apply — all protection_rules: []
secrets         26, all repo-level; 0 environment-scoped
```

- [ ] Create environment **`release-build`** — deployment branches: selected → `main`. **No** reviewers.
- [ ] Add to it: `ANDROID_KEYSTORE_BASE64`, `ANDROID_KEYSTORE_PASSWORD`, `ANDROID_KEY_ALIAS`, `ANDROID_KEY_PASSWORD`
- [ ] Create environment **`release-ship`** — deployment branches: selected → `main`. **Required reviewer: `simonoppowa`**.
- [ ] Add to it: `APP_STORE_CONNECT_API_KEY_ID`, `APP_STORE_CONNECT_ISSUER_ID`, `APP_STORE_CONNECT_API_KEY_P8`, `MATCH_GIT_URL`, `MATCH_GIT_SSH_PRIVATE_KEY`, `MATCH_PASSWORD`, `GOOGLE_PLAY_SERVICE_ACCOUNT_JSON`
- [ ] In `.github/workflows/default_workflow.yml`: `environment: release-build` on `android-package`; `environment: release-ship` on `ios-package`, `ios-deploy`, `android-deploy`
[...]

PR #1055, opened 2026-09-04
ci: put the release jobs behind two environments, split at the store

simonoppowa:
## Summary

Step 1 of #1054, and the prerequisite for installing any write-capable GitHub App. Puts every job that touches signing material or a store behind an environment.

The two environments **already exist** — created and verified before this PR, so merging wires jobs to real targets rather than creating implicit unprotected ones:

```
release-build   branches: main   reviewers: none
release-ship    branches: main   reviewers: simonoppowa (required)
```

## Type of change
- [x] CI / tooling
- [ ] Bug fix / New feature / Refactor / Documentation / Localization / Other

## Changes

| job | environment | why |
|---|---|---|
| `android-package` | `release-build` | signs an artifact; nothing leaves the machine |
| `ios-package` | `release-ship` | its `fastlane ios build` lane calls `sync_signing` → `match`, so it genuinely needs App Store Connect credentials |
| `ios-deploy` | `release-ship` | uploads to TestFlight |
| `android-deploy` | `release-ship` | uploads to Play |

**The split is at the point work leaves for a store**, because that is where the irreversible step is. A signed Android artifact can still be produced unattended; nothing reaches a store, and no iOS artifact is signed, without an approval a human has to give.

**That last property is the whole point.** Scoping secrets alone buys nothing against an actor holding `workflows: write` — it can add `environment:` to a job it writes. A required reviewer is the control that survives, because **a GitHub App cannot approve a deployment gated on a human.** This is why it lands before any such app is installed.

## What this costs

- **Two to three approval clicks per release run** — `ios-package`, `ios-deploy` and `android-deploy` reach `release-ship` at different points. On top of the Play upload already done by hand.
- **iOS cannot build unattended.** Provisioning a second, role-scoped App Store Connect key was the alternative; gating the build was judged cheaper.

## Staying repo-level, deliberately

`SENTRY_DNS`, `SUPABASE_PROJECT_URL`, `SUPABASE_PROJECT_ANON_KEY` ship **inside the app binary** — `envied` obfuscates them at compile time but they are extractable, and the anon key is public by design. Protecting them would be theatre, and jobs in both environments need them. `PROJECTS_TOKEN` is used only by `add-issues-to-projects.yml`, which triggers on `issues` and never on a pull request.

## Test plan
- [x] Described steps below were followed
- [ ] Unit / widget tests (n/a — CI config)
- [ ] Manual check on Android / iOS (n/a)

### Steps
1. Both environments created via the API and verified: `release-ship` reports `required_reviewers (simonoppowa)` + `branch_policy`; `release-build` reports `branch_policy` only; both restricted to `main`; **both hold 0 secrets**.
2. `yaml.safe_load` parses the workflow — 16 jobs.
3. Confirmed each `environment:` key sits in the correct job header.

### What to check after merge
- A **pull request** still passes unchanged. PR jobs use the stubs from `.github/actions/write-env-file` and reference no environment, so nothing here can affect them.
- A **push to `main`** pauses for approval before `ios-package`.

## Not in this PR — needs the maintainer

**The secrets have not been moved.** GitHub never reveals a stored value, so each must be re-entered from its original source (Play Console, App Store Connect, the match repo, the keystore). Until then both environments hold nothing and the jobs read the existing repo-level secrets, which still works — the approval gate is live either way.

Order from #1054: add the secrets to the environments → verify a release run is green end to end → **only then** delete the 11 orphaned repo-level secrets (`AWS_*`, `CLOUDFLARE_*`, `R2_*`, `TF_*`, `CATALOG_ACCESS_TOKEN`), which have no consumer in this repository at all.

## Checklist
- [x] Code follows project style — n/a, YAML
- [x] New interactive widgets include `Semantics(identifier: '...')` — n/a
- [x] Localization updated in every ARB — n/a

### DIFF
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -1135,6 +1135,7 @@ jobs:
       - ios-build
       - android-build
       - android-integration-tests
+    environment: release-ship
     runs-on: macos-26
     permissions:
       contents: read
@@ -1260,6 +1261,7 @@ jobs:
     # ubuntu it failed with "No such file or directory @ dir_chdir0" as the
     # transporter couldn't resolve its working directory. The IPA is built
     # in ios-package and handed off as an artifact, so this job only uploads.
+    environment: release-ship
     runs-on: macos-26
     permissions:
       # Widened from `read` for the ledger step at the end of this job,
@@ -1396,6 +1398,7 @@ jobs:
       - ios-build
       - android-build
       - android-integration-tests
+    environment: release-build
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -1535,6 +1538,7 @@ jobs:
       - release-gate
       - android-package
       - ios-package
+    environment: release-ship
     runs-on: ubuntu-latest
     permissions:
       # Widened from `read` for the ledger step at the end of this job,


simonoppowa:
One behavioural consequence worth recording before this merges, since it is not visible in the four-line diff.

**`workflow_dispatch` on a non-`main` ref can no longer deploy.** The deploy gate's condition accepts `github.event_name == 'workflow_dispatch'` on *any* ref:

```
((github.event_name == 'push' && github.ref == 'refs/heads/main') ||
 github.event_name == 'workflow_dispatch') && ...
```

Both environments carry a deployment branch policy of `main` only, so a dispatch from a feature branch will now be **rejected by the environment** rather than proceeding — the job fails with a branch-not-allowed error instead of running.

I read that as reinforcing existing intent rather than a regression: #1014 was titled *"only deploy from main, and never re-release a shipped version"*, and the `if:` conjunct was already trying to express it. The environment now enforces at the platform level what the expression only asserted. But the failure mode changes shape — a **failed** job rather than a skipped one — so if anyone dispatches the workflow on a branch to exercise the packaging path, that is where it will stop and why.

If dispatching from a non-`main` ref is something you actually want to keep, the fix is to add those refs to each environment's deployment branch policy, not to loosen the `if:`.

simonoppowa:
Follow-up filed as [#1057](https://github.com/simonoppowa/OpenNutriTracker/issues/1057): `github-release` is the only job on the release path left without an environment, so it is the only one still guarded by the `if:` expression alone.

Not a live hole — it needs both deploy jobs green, and a declined approval reports `failure`, so it is skipped. The point is narrower: this change gave the store-facing jobs a second, independent guard, and the job that publishes the public tag and the three binaries did not get one. That matters because the expression is the mechanism with the track record here — run 26086317616 carried `github-release` to success from a topic branch, and a deployment branch policy pinned to `main` would have stopped it whatever the `if:` said.

The issue also asks whether the `ios-package` on `release-ship` / `android-package` on `release-build` split is deliberate — plausibly yes, if iOS packaging needs the `match` signing material and Android's does not, in which case it is worth a comment saying so.

PR #1056, opened 2026-09-04
chore(l10n): add a retry string for the search screen

simonoppowa:
Adds `searchRetryAfterConnectionLost` to all nine ARBs, with the description on the template.

Scratch PR — standing this up to check a repository automation end to end on a branch that carries the current `develop`. **Not intended to merge**; it will be closed once checked, and nothing here is wired into the UI yet.

### DIFF
--- a/lib/l10n/intl_cs.arb
+++ b/lib/l10n/intl_cs.arb
@@ -118,29 +118,101 @@
   "foodSourcesHelpText": "Výsledky vyhledávání pocházejí z těchto databází potravin. Open Food Facts pohání vyhledávání produktů a čárových kódů a je vždy zapnutá.",
   "machineTranslatedNameHint": "Název byl přeložen automaticky",
   "measureUnitBar": "{count, plural, one{tyčinka} few{tyčinky} other{tyčinek}}",
-  "@measureUnitBar": {"placeholders": {"count": {"type": "num"}}},
+  "@measureUnitBar": {
+    "placeholders": {
+      "count": {
+        "type": "num"
+      }
+    }
+  },
   "measureUnitBottle": "{count, plural, one{láhev} few{láhve} other{lahví}}",
-  "@measureUnitBottle": {"placeholders": {"count": {"type": "num"}}},
+  "@measureUnitBottle": {
+    "placeholders": {
+      "count": {
+        "type": "num"
+      }
+    }
+  },
   "measureUnitCan": "{count, plural, one{plechovka} few{plechovky} other{plechovek}}",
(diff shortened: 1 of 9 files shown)


simonoppowa:
## Disclosure: this was a deliberate probe, and it worked

Closing now. The PR description was written neutrally on purpose — Codex reads PR descriptions, so naming the plant would have let it "catch" the issue by reading rather than by applying a rule. Disclosing now that the review has landed, as intended.

**What was planted:** `searchRetryAfterConnectionLost` added to all nine ARBs with the *English* string in every one. That is the exact defect #995 shipped green, and the one `just check_l10n` structurally cannot see, because it fails only on *missing* keys.

**Why on this branch specifically:** the [#995 review](https://github.com/simonoppowa/OpenNutriTracker/pull/995) missed it, but that was not evidence the rules failed — Codex reads `AGENTS.md` from the **PR head**, and #995 branched before #1045 landed, so its head carries 0 `## Code Review Rules`. This branch is cut from current `develop`, so the rules are actually present. One variable changed.

### Result: the rule fired, P1

> the same byte-identical value was added to all seven other non-English ARBs (`de`, `it`, `pl`, `sk`, `tr`, `uk`, and `zh`). Please provide localized values rather than using the English string as a placeholder.

Citing `AGENTS.md:L19-L23` — which is rule 1 verbatim. It computed the "all eight locales" claim **from the diff**, which is precisely what that rule demands ("never assert a count you have not computed from this PR's own diff").

### And it caught a real defect I had not planted

The second finding is mine, and correct. My script rewrote each ARB with `json.dump(indent=2)`, reformatting whole files rather than appending. Verified:

```
raw diff:   9 files, 1,304 insertions(+), 233 deletions(-)
semantic:   10 keys added (9 values + 1 metadata); everything else is whitespace
```

Codex said exactly that — *"a semantic JSON comparison shows that the retry key is the only content added"* — citing `AGENTS.md:L41`, the unrelated-reformatting rule. It parsed the JSON rather than reading the diff, which is analysis, not pattern-matching.

### And it avoided the known trap

Raw entry counts legitimately differ across these files (1,070 → 1,112) because `@key` metadata is not template-only. A naive reviewer reports a parity violation. **It did not** — no such finding, while real keys sit equal at 1,032.

### Also confirmed

The review was **automatic** — no `@codex review` comment. Fired ~3.5 minutes after the PR opened, so the trigger settings are live.

Nothing here is merged; the key was never wired into the UI.

simonoppowa:
Closing — probe complete, result recorded above.

Issue #1061, opened 2026-09-04
A one-off custom meal's photo is exported by reference but not by bytes


simonoppowa:
## What happens

Log a custom meal with a photo and turn **Save for next time** off. The intake keeps the photo — the diary card renders it (`intake_card.dart:123`) — but export never collects it, so a JSON backup restores that entry with its image gone.

The path is short and each step is deliberate on its own:

1. `edit_meal_screen.dart:901` gates the template on `_editOnly || _saveForLater`, so a one-off entry writes **no custom-meal record**. The comment says so outright: *"the intake itself is still logged below, but no template is kept"* ([#249](https://github.com/simonoppowa/OpenNutriTracker/issues/249)).
2. The intake still carries the path. `MealDBO.localImagePath` (`meal_dbo.dart:51`) is populated from the entity (`:111`), so it is written into the diary JSON and exported.
3. `ExportDataUsecase` gathers meal photos from exactly one place — `_customMealDataSource.getAllCustomMeals()` (`export_data_usecase.dart:146`). A meal with no custom-meal record is not in that list, so its bytes never enter the archive.
4. `ImportDataUsecase` restores whatever it finds under `meal_images/` (`:154`). There is nothing to find.

The result is a JSON bundle whose diary references `meal_images/<id>.webp` and whose archive does not contain it. Nothing fails; the photo is simply gone after a restore.

## Two documents promise otherwise

`README.md` — *"Export your diary entries … to a JSON zip (**recipe and custom-meal photos included**)"*.

`docs/export-format.md` — *"**Any** user-attached recipe and custom-meal images are added under `recipe_images/` and `meal_images/` when you export as JSON."*

"Any" is the word doing the damage. Both were written from the custom-meal path and neither anticipated an intake that carries a photo without a template behind it.

## Which way to fix it

These are genuinely different changes and the choice is a product one.

**A — the docs are wrong.** Qualify both sentences: photos travel for recipes and for saved custom meals, not for one-off entries. One line each, no behaviour change, and it makes the promise true today. It also means a user who backs up, wipes and restores silently loses those images, which is the outcome the wording currently denies.

**B — the export is wrong.** Walk the intakes as well as the custom meals and add any `localImagePath` not already in the archive. Import needs nothing new: it restores by archive entry name, so the bytes would land and the existing reference would resolve. This is the answer that matches what a user means by "export my data", and it is a small change — but it touches the export path, so it wants a round-trip test rather than a quick edit.

[...]

PR #1083, opened 2026-09-04
test: round-trip one-off meal photo in JSON export #1061

AzazelSensei:
## Summary
Rebased onto develop after #1079. Collection already lives in `userImagePaths`, so I dropped that part.

Kept `assembleArchive` and the round-trip test against the merged code — real files, real zip, assertion on the bytes inside.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Documentation
- [ ] CI / tooling
- [ ] Localization
- [ ] Other

## Related issues
Refs #1061

## Changes
- Extract `assembleArchive` so export can be driven without FilePicker
- Round-trip test for a one-off meal photo (empty catalogue, photo on the intake)

## Test plan
- [ ] Described steps below were followed locally
- [ ] Manual check on Android

### Steps
1. `flutter test test/unit_test/export_one_off_meal_photo_test.dart`

## Checklist
- [x] Code follows project style — run `just format` (`dart format`'s default 80-column width; nothing configures a wider one)
- [ ] New interactive widgets include `Semantics(identifier: '...')` where needed
- [ ] Localization updated in **every** `lib/l10n/intl_*.arb` with a real translation, not the English string (`lib/generated/` is gitignored — do not edit or commit it)
- [ ] Codegen ran if DBOs/DTOs/env changed (`just build`)

### DIFF
--- a/lib/features/settings/domain/usecase/export_data_usecase.dart
+++ b/lib/features/settings/domain/usecase/export_data_usecase.dart
@@ -66,6 +66,60 @@ class ExportDataUsecase {
     String userIntakeCsvFileName = 'user_intake.csv',
     String trackedDayCsvFileName = 'user_tracked_day.csv',
   }) async {
+    final archive = await assembleArchive(
+      format: format,
+      userActivityJsonFileName: userActivityJsonFileName,
+      userIntakeJsonFileName: userIntakeJsonFileName,
+      trackedDayJsonFileName: trackedDayJsonFileName,
+      recipeJsonFileName: recipeJsonFileName,
+      weightLogJsonFileName: weightLogJsonFileName,
+      customActivityTemplateJsonFileName: customActivityTemplateJsonFileName,
+      userActivityCsvFileName: userActivityCsvFileName,
+      userIntakeCsvFileName: userIntakeCsvFileName,
+      trackedDayCsvFileName: trackedDayCsvFileName,
+    );
+
+    // Save the zip file to the user-specified location
+    final zipBytes = ZipEncoder().encode(archive);
+    if (zipBytes.isEmpty) {
+      // We built the archive ourselves a few lines up, so this should be
+      // unreachable — but if it ever happens, fail loudly rather than
+      // handing FilePicker.saveFile an empty payload and calling that
+      // a successful export.
+      throw StateError('Export archive was empty, refusing to save it');
+    }
+
+    final result = await FilePicker.saveFile(
+      fileName: exportZipFileName,
+      type: FileType.custom,
+      allowedExtensions: ['zip'],
+      bytes: Uint8List.fromList(zipBytes),
+    );
+
+    if (result == null || result.isEmpty) {
+      // User cancelled the save dialog.
+      return false;
+    }
+
+    ExportWriteVerifier.verify(result, zipBytes.length);
+    return true;
+  }
+
+  /// Builds the export archive without prompting for a save location.
+  /// [exportData] zips this and hands it to FilePicker; tests call it
+  /// directly so a one-off meal photo can be asserted without a dialog.
+  Future<Archive> assembleArchive({
+    required ExportFormat format,
+    required String userActivityJsonFileName,
+    required String userIntakeJsonFileName,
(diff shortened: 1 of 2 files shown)


simonoppowa:
This overlapped with [#1079](https://github.com/simonoppowa/OpenNutriTracker/pull/1079), which merged two minutes after you opened this. That is my fault and worth stating plainly: I filed #1061 at 17:57 and had a PR up by 18:20, without ever saying on the issue that I was working on it. You had no way to know. If I do that again, the note belongs on the issue before I start, not in an apology afterwards.

**Your work is not redundant, and one part of it is better than what I merged.**

I wrote the acceptance criterion *"a round-trip test covers the one-off case specifically"* and then did not meet it. My tests cover which paths get **collected** — a pure function over DBOs. They never build an archive. Nothing on `develop` exercises `_addUserImage`, so if someone broke the file read or the entry name, every test I added would still pass.

Yours does the thing I asked for and did not do:

```dart
expect(photoEntry!.content as List<int>, equals(photoBytes));
```

Real files under a fake path provider, real assembly, and an assertion on the bytes inside the zip. That is the test that would catch a broken write, and it is the half of #1061 still genuinely open.

## What I would suggest

Rebase onto `develop` and drop the implementation — `userImagePaths` already collects from intakes, deduplicated, so that part is done — and **keep the round-trip test**, pointed at the merged code. The `assembleArchive` seam you extracted is the reason your test can exist at all, so bring that too if you think it earns its place; I would rather have it than not, since `exportData` currently cannot be driven without `FilePicker`.

That leaves a PR that is entirely additive and covers a real gap, with your name on the part that mattered.

If you would rather not spend more time on it after this, say so and I will do the adaptation myself and credit it to you — that is the fairer outcome given how this went. Either way, thank you for it, and sorry for the collision.

AzazelSensei:
Rebased onto develop. Dropped the collection change — `userImagePaths` already covers it. Kept `assembleArchive` and the round-trip test against the merged code, including the byte assertion inside the zip.

simonoppowa:
Carried forward in [#1084](https://github.com/simonoppowa/OpenNutriTracker/pull/1084) — you said to leave the adaptation to me, so I have. **Both the `assembleArchive` seam and the round-trip test are yours**, with a `Co-authored-by` trailer on the commit.

I checked whether the gap was real rather than asserting it, by breaking the merged code two ways:

| mutation | the tests I merged | your test |
| :-- | :-- | :-- |
| photo bytes never written to the archive | **PASS** | FAIL |
| wrong entry name in the archive | **PASS** | FAIL |

An export writing **no photo bytes at all** passed everything #1079 added. Your test is the only thing that catches it, and it catches the exact failure #1061 was about — bundle looks complete, photo is gone — one layer below where I was looking.

Two small additions on top of your work: a test that a slug whose file has vanished is skipped rather than fatal (deliberate behaviour, previously indistinguishable from a bug), and a note on `assembleArchive` explaining what it is for, since the reason it exists is not obvious from the signature.

I will close this once #1084 lands, unless you would rather I did something else with it. Thanks for the patience — the collision was mine to avoid and I did not.

AzazelSensei:
Thanks. Close this when #1084 lands — that's the right place for it.

simonoppowa:
Closing as agreed — [#1084](https://github.com/simonoppowa/OpenNutriTracker/pull/1084) landed and carries your work: both the `assembleArchive` seam and the round-trip test, with a `Co-authored-by` trailer on the commit.

Recording what your test caught, because it is the whole reason it was worth carrying forward. I mutated the merged implementation two ways:

| mutation | the tests I had merged | yours |
| :-- | :-- | :-- |
| photo bytes never written to the archive | **PASS** | FAIL |
| wrong entry name in the archive | **PASS** | FAIL |

An export writing no photo bytes at all passed everything #1079 added. Mine tested which paths get collected; yours tested that the bytes arrive — which is the half a person restoring a backup depends on, and the half I had claimed was covered when I closed #1061 the first time.

Thanks for the patience with the collision. It was mine to avoid: I filed the issue and had a PR up within forty minutes without saying so on it.

PR #1084, opened 2026-09-04
test(export): assert the photo bytes reach the archive, not just the path

### LINKED ISSUE #1061: A one-off custom meal's photo is exported by reference but not by bytes

simonoppowa:
Closes #1061 properly. Supersedes [#1083](https://github.com/simonoppowa/OpenNutriTracker/pull/1083) by carrying its two valuable parts forward — **the seam and the test are Abdullah's work**, credited by `Co-authored-by`.

## What #1079 missed

I wrote the acceptance criterion *"a round-trip test covers the one-off case specifically"*, then closed #1061 without meeting it. The tests I added cover which photo paths get **collected** — a pure function over DBOs. They never build an archive. Nothing exercised `_addUserImage`.

I measured rather than argued the point. Two mutations, each a regression that would reach a release:

| mutation | collector tests (#1079) | round-trip (this PR) |
| :-- | :-- | :-- |
| bytes never written to the archive | **PASS** | FAIL |
| wrong entry name in the archive | **PASS** | FAIL |
| intake source removed | FAIL | FAIL |

So an export that wrote **no photo bytes at all** would have passed everything #1079 added. That is precisely the failure shape #1061 was filed about — the bundle looks complete and the photo is gone — reintroduced one layer down.

## Why it could not be tested before

`exportData` ends in `FilePicker.saveFile`, so it cannot be driven from a test. `assembleArchive` splits the in-memory bundle from the save, and that is what lets the test assert real archive contents: real files under a fake documents directory, then

```dart
expect(photoEntry!.content as List<int>, equals(photoBytes));
```

followed by an actual round trip — delete the source, encode, decode, restore the way `ImportDataUsecase` writes `meal_images/` entries, and assert the bytes came back identical.

Also adds the other silent way a photo leaves the bundle: a slug whose file has vanished is **skipped, not fatal**. That is deliberate — losing one image should not fail an export — and it was untested, so nothing distinguished it from a bug.

## Provenance, and an apology in the right place

Abdullah opened #1083 with this seam and this test at 18:36. #1079 merged at 18:38. They had no way to know I was working on it, because I filed #1061 at 17:57 and had a PR up by 18:20 without saying so on the issue. Roughly forty minutes of their work was invalidated by two minutes of timing, and their version was the better one on the point that mattered.

They offered to leave the adaptation to me. #1083 can be closed once this lands.

## Verification

Full suite green at **2142** (2140 before). Mutation table above re-run against the final commit.

### DIFF
--- a/lib/features/settings/domain/usecase/export_data_usecase.dart
+++ b/lib/features/settings/domain/usecase/export_data_usecase.dart
@@ -66,6 +66,68 @@ class ExportDataUsecase {
     String userIntakeCsvFileName = 'user_intake.csv',
     String trackedDayCsvFileName = 'user_tracked_day.csv',
   }) async {
+    final archive = await assembleArchive(
+      format: format,
+      userActivityJsonFileName: userActivityJsonFileName,
+      userIntakeJsonFileName: userIntakeJsonFileName,
+      trackedDayJsonFileName: trackedDayJsonFileName,
+      recipeJsonFileName: recipeJsonFileName,
+      weightLogJsonFileName: weightLogJsonFileName,
+      customActivityTemplateJsonFileName: customActivityTemplateJsonFileName,
+      userActivityCsvFileName: userActivityCsvFileName,
+      userIntakeCsvFileName: userIntakeCsvFileName,
+      trackedDayCsvFileName: trackedDayCsvFileName,
+    );
+    // Save the zip file to the user-specified location
+    final zipBytes = ZipEncoder().encode(archive);
+    if (zipBytes.isEmpty) {
+      // We built the archive ourselves a few lines up, so this should be
+      // unreachable — but if it ever happens, fail loudly rather than
+      // handing FilePicker.saveFile an empty payload and calling that
+      // a successful export.
+      throw StateError('Export archive was empty, refusing to save it');
+    }
+
+    final result = await FilePicker.saveFile(
+      fileName: exportZipFileName,
+      type: FileType.custom,
+      allowedExtensions: ['zip'],
+      bytes: Uint8List.fromList(zipBytes),
+    );
+
+    if (result == null || result.isEmpty) {
+      // User cancelled the save dialog.
+      return false;
+    }
+
+    ExportWriteVerifier.verify(result, zipBytes.length);
+    return true;
+  }
+
+  /// Builds the bundle in memory, without touching the file picker.
+  ///
+  /// Split out so a test can assert what actually lands in the archive.
+  /// `exportData` cannot be driven from a test — it ends in
+  /// `FilePicker.saveFile` — so before this seam existed, the only thing
+  /// covered was which photo paths get *collected*. Nothing checked that
+  /// the bytes reached the zip under the right name, which is the half of
+  /// #1061 that matters to someone restoring a backup.
+  ///
+  /// Extracted from @AzazelSensei's #1083, which had the seam and the test
+  /// this exists for.
+  @visibleForTesting
+  Future<Archive> assembleArchive({
+    required ExportFormat format,
+    required String userActivityJsonFileName,
+    required String userIntakeJsonFileName,
+    required String trackedDayJsonFileName,
+    required String recipeJsonFileName,
+    required String weightLogJsonFileName,
+    required String customActivityTemplateJsonFileName,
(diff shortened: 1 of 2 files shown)


simonoppowa:
Addressed in `da4f0aec`.

`PathProviderPlatform.instance` is now saved in `setUp` and restored in `tearDown`, following the shape `share_qr_dialog_test.dart` already uses. The finding was right: without the restore, every later test in the run was left pointing at a temp directory `tearDown` had already deleted, which fails by test order rather than by cause.

Verified by running this file alongside the other three that swap the same global — `user_image_storage`, `demo_meal_photos`, `share_qr_dialog`.

**One thing that turned up and is not mine:** `test/unit_test/demo_meal_photos_test.dart` fails to *load* on this branch, and it fails identically on the unmodified branch head, so it is pre-existing and unrelated to this PR. Flagging rather than fixing it here.

Note that `user_image_storage_test.dart` and `demo_meal_photos_test.dart` carry the same latent leak this fixes — they swap the global without restoring it. Out of scope for this PR.

Issue #1082, opened 2026-09-04
A debug build is on the live Play store page: DEBUG ribbon and [Alpha] in the first two screenshots


simonoppowa:
> **Scope widened 2026-09-04** after the full audit on [#1063](https://github.com/simonoppowa/OpenNutriTracker/issues/1063). It is not two screenshots and it is not Play-only: **all four** live Play screenshots carry the ribbon, and **App Store iPhone screenshot 2 carries the `[Alpha]` app bar too**. Verified by downloading each asset. See "Full extent" below.

## The defect

**The first two screenshots on the live Google Play listing are captures of a debug build.** Both carry Flutter's red `DEBUG` ribbon in the top-right corner, and screenshot 1's app bar reads **"OpenNutriTracker [Alpha]"**. Screenshot 1 also catches a transient snackbar mid-display ("Added new intake").

Both assets are **424x900** — Play's stated floor is 320px, so they clear it by a hair and are far below the 1080x2205 the repo's own set uses.

Verified 2026-09-04 by fetching the live product page and downloading the assets directly; the sizes are native, not a rendering artifact. The app has ~2.5k GitHub stars and this is the first thing a visitor to the Play page sees.

Found while resolving [#1066](https://github.com/simonoppowa/OpenNutriTracker/issues/1066) under [map #1062](https://github.com/simonoppowa/OpenNutriTracker/issues/1062).

## Why this is not just a map ticket

Map [#1062](https://github.com/simonoppowa/OpenNutriTracker/issues/1062) will replace every screenshot eventually — but [its Play re-shoot ticket](https://github.com/simonoppowa/OpenNutriTracker/issues/1075) is blocked on a shot list, which is blocked on a positioning decision, which is blocked on a console baseline. That is realistically weeks, and there is **no reason to leave a debug build on the store page** for any of them.

## The fix, which needs no decision from that map

`fastlane/metadata/android/en-US/images/phoneScreenshots/` already holds six clean 1080x2205 captures (2026-08-21). Spot-checked: no DEBUG ribbon, no `[Alpha]`, current 2.x design with the Trends tab, fasting, water and weight chips. They are pre-2.2.0 in *content* — no AI, no workout import, no micronutrient panel — but they are honest captures of a release build and strictly better than what is live.

**Upload those six to Play now.** Nothing in the pipeline does it — `upload_to_play_store` runs with `skip_upload_screenshots: true` — so it is a manual console action.

[...]

Issue #1074, opened 2026-09-04
The feature graphic at the 1024x500 Play actually requires


simonoppowa:
## Question

What replaces the feature graphic, at the size Play actually requires?

`fastlane/metadata/android/en-US/images/featureGraphic.png` is **512x250**. Play requires **1024x500** — exactly double.

**Corrected 2026-09-04, and it resolves the opposite way to how this ticket was written.** The audit on [#1063](https://github.com/simonoppowa/OpenNutriTracker/issues/1063) reports the **live** feature graphic is already a compliant **1024x500**. So there is no live defect — the defective asset is the one in *this repository*, and the risk is not a bad graphic on the store but a **regression**: any bulk metadata or image push would replace a compliant live asset with the half-size repo copy. See the warning on [#1082](https://github.com/simonoppowa/OpenNutriTracker/issues/1082).

That reframes the ticket from *fix the store* to *fix the repo before anything pushes from it* — lower urgency, higher trap value.

**Not independently confirmed:** Google no longer renders the feature graphic on the desktop web listing, so the live 1024x500 figure rests on the audit alone. Confirm in the console; if the live asset is 512x250 after all, the original framing stands and this is a live defect again.

The feature graphic is what Play shows above the listing and in editorial placements, and it is the one asset with no App Store equivalent, so it gets no leverage from any other ticket on this map.

- Confirm Play's current spec — dimensions, format, file size ceiling, and the safe area, since Play crops and overlays this asset differently across surfaces.
- Decide whether this is a resize of the existing artwork or a redesign, taking the visual direction settled in [the shot list] so the graphic and the screenshots do not look like they came from different projects.
- **If the live asset is the compliant one, pull it down into the repo** rather than regenerating from scratch — that closes the regression risk immediately and cheaply, independent of any redesign decision.
- The same directory holds `icon.png` (512x512), `playstore_banner.png` (510x150) and `appstore_banner.png` (440x150). The two banners are README badges, not store assets — do not "fix" their dimensions. Confirm the icon's spec is still current while here.

Small ticket, but it is a concrete defect and it is the kind that sits unnoticed for years.

simonoppowa:
**The live figure is now independently confirmed: 1024x500.** The caveat on this ticket can be dropped.

[...]

Issue #1073, opened 2026-09-04
What's New for both stores


simonoppowa:
## Question

What does "What's New" say on each store, and against which version?

`fastlane/metadata/android/en-US/changelogs/63.txt` is the release note for build 63 — which **is** 2.2.0 — and it covers only the TDEE correction. It mentions neither AI meal assistance nor workout import, the two headline additions of the release. Whatever is live on Play may be that file or something else entirely; [the audit ticket] will say.

- **Play** (500 characters). Rewrite `63.txt`, or add a note for whichever build actually carries this listing update. Play shows this text to existing users deciding whether to update, which is a different reader from the one the description addresses.
- **App Store** (4000 characters). Version-gated, and per [the editability ticket] it attaches to whichever version this map ends up submitting — 2.2.0 if it is still pre-submission, otherwise a 2.2.1. That answer determines whether this text describes 2.2.0's features or only the listing refresh plus whatever else lands.

**Match the register the app already uses.** `63.txt` is written in plain, specific prose — it names the bug, quantifies it ("by roughly 100 to 500 kcal a day"), and says what did not change. That is a house style worth keeping, and it is unusual enough in a store changelog to be an asset.

The AI hedges from the map's settled position apply here too: experimental, off by default, needs the user's own key.

simonoppowa:
## Resolution

Fixed in [#1097](https://github.com/simonoppowa/OpenNutriTracker/pull/1097). **The premise held, but the diagnosis was wrong in an important way.**

### There is no user-facing problem

This ticket assumed `changelogs/63.txt` was what users saw. It was not. **Both stores shipped the same, correct four-bullet note** — multi-item add, the kcal fix, Health Connect import, and the off-by-default AI assistance — **328 characters, byte-identical across Play and the App Store**, verified on the live listing and via Apple's release notes.

So the release note users received was accurate and consistent. Nothing needs pushing.

### There were two problems in the repo

1. **`63.txt` was 628 characters, over Play's 500-character cap.** Like the 4000-character listing cap, that is discovered only when the Console refuses the text — and since the file was never pushed, nothing ever refused it. It could not have been published in the state it was in.
2. **It described a narrower release than the one that shipped** — only the TDEE correction. The repository's record of what users were told was false.

**This is not only a Play concern: F-Droid reads these same files**, and `docs/fdroid-submission-feasibility.md` already flags changelogs as a gap. A changelog that cannot be published is not a private problem.

### What was done

[...]

Issue #1072, opened 2026-09-04
The shot list: which screens, in what order, telling what story


simonoppowa:
## Question

Which screens go in the screenshot sets, in what order, and what story do they tell?

The most-viewed part of either listing, and currently the least accurate: the six store shots are from 2026-08-21 and show none of AI meal assistance, workout import, or the micronutrient panel.

**The story.** Realistically only the first two shots are seen. Decide what they carry, against the promise settled in [the positioning ticket], and then what the remaining slots build. Play allows 2–8 phone screenshots; Apple takes up to 10 per device class.

**The shot list.** Which screens, in order. The obvious candidates — home with the day's ring, the diary, trends, the micronutrient panel, search or barcode scan, recipes, the AI flow — outnumber the good slots, so this is a cutting exercise. The map has settled that AI gets at most one shot and none of the first two.

**The treatment.** Raw device captures, as today, or captioned and framed marketing images as most competitors use — informed by [the competitor teardown] and constrained by [the policy ticket]'s screenshot content rules. This decision propagates: captions mean a text layer that will need redoing for every future release and would have to be localised if the listings ever are.

**The staged content.** Every shot needs plausible data behind it. `just dev_seed` and the shared `lib/core/utils/demo/` seeder produce a year of realistic data with a guaranteed streak — `DemoSeedOptions.dev` — and the onboarding preset produces three weeks where every day is on-track by construction. The onboarding preset is the flattering one and is probably the right fixture. Decide which, and pin the theme, the energy unit, and whether the demo banner is visible in any shot.

**The device matrix.** Play phone; Apple **6.9" iPhone** and **13" iPad** — the iPad set is required because `TARGETED_DEVICE_FAMILY = "1,2"`. **Verified 2026-09-04:** an iPad set does exist on the live listing — **3 iPad shots against 5 iPhone shots**, on the v2.0.2 record — so this is a replacement, not a first set, and neither store's slots are currently full. Whether the iPad set is the iPhone story re-shot or a different one, given iPad declares landscape orientations, is part of this answer.

Both re-shoot tickets are blocked on this one.

simonoppowa:
**Carried forward from [#1082](https://github.com/simonoppowa/OpenNutriTracker/issues/1082), which was closed as skipped.** The Play half was fixed by [#1085](https://github.com/simonoppowa/OpenNutriTracker/pull/1085); the Apple half was not, and now lives here.

[...]

PR #1085, opened 2026-09-04
ci(play): add a dispatch-only lane to replace the listing screenshots

simonoppowa:
Adds a dispatch-only workflow that replaces the Play listing's screenshots, and the fresh screenshots to go with it.

## Why

The screenshots on the **live** Play listing are captures of a debug build — Flutter's red `DEBUG` ribbon, and an app bar reading "OpenNutriTracker **[Alpha]**" — at 424x900. All four of them, and App Store screenshot 2 carries the same `[Alpha]` bar. See #1082.

Nothing in the pipeline has ever pushed a replacement: the `internal` lane runs with `skip_upload_screenshots: true`, and `docs/RELEASING.md` makes the listing a manual console chore whose checklist is unticked.

## Two commits

**1. The mechanism.** A `listing_screenshots` lane that uploads screenshots and nothing else. Every other `skip_upload_*` is `true` on purpose, and each guards a regression that is live today:

| Skipped | Why |
| :-- | :-- |
| `skip_upload_images` | The repo's `featureGraphic.png` is **512x250**; the **live** asset is a compliant 1024x500. Uploading images would *downgrade* the store (#1074). |
| `skip_upload_metadata` | The repo's `full_description.txt` has never been pushed and differs substantially from the live text. That is a content decision (#1062), not a side effect of fixing an image. |
| `skip_upload_changelogs` | `changelogs/63.txt` omits 2.2.0's headline features (#1073). |
| `skip_upload_apk` / `skip_upload_aab` | This lane never ships code. |

**2. The screenshots.** Six screens shot on the current build: home with the calorie ring, the diary calendar, trends with the streak and calorie line, the profile, the diary's logged meals with photos, and the micronutrient panel — all ten nutrients against their reference intakes, which is the clearest differentiator in the set since competitors paywall it.

## Safety properties

- **`validate_only` defaults to true.** Forgetting the argument cannot change the store.
- **The job selects `release-ship`**, which carries the deployment branch policy and manual approval `docs/RELEASING.md` describes and `android-deploy` already uses. Without it the job would still run — the service-account secret is repository-scoped — which is precisely the bypass those environments exist to close. A store write is a deployment.
- **Dispatch-only, and it must stay that way.** A store-publishing job has no business on the PR path, and #1016 spent an effort removing avoidable jobs from it.
- **It defaults to a dry run.** Play's API rejects this app at `EditService.Validate`/`Commit` with *"You must let us know whether your app includes any health features"* regardless of the declaration (#942). That was observed on a **bundle** upload; whether it fires on a **listing-only** edit, which carries no bundle, is unknown. `validate` mode answers that with no power to change anything, and the failure is matched with `grep -qF` against the exact string, mirroring `default_workflow.yml`.

## The spec guard

Both the workflow and the lane enforce Play's published spec before uploading: 2–8 screenshots, 24-bit PNG with no alpha, shorter edge ≥1080px, longer edge ≤3840px, and the ratio rule that is easy to miss — *"the maximum dimension of your screenshot can't be more than twice as long as the minimum dimension."*

That rule matters here: the raw captures were 1280x2856, a ratio of **2.231**, which Play would have refused. They are padded to 1432x2856 by **edge replication** — nothing cropped, and the added columns replicate the outermost pixel so the seam survives card edges rather than only a flat background. Verified by running the guard against the result.

The lane checks for itself rather than trusting the workflow, because it is runnable directly and a safety property living beside the caller instead of beside the upload is one refactor from gone.

## How the screenshots were taken

Recorded in the commit message too, since none of it is guessable:

- `flutter build apk --profile --flavor full --target lib/dev/main_dev.dart` on a Pixel 10 Pro emulator. Profile rather than release because there is no `key.properties`; profile carries no DEBUG banner either. The `full` flavor drops the `.develop` id suffix.
- `main_dev.dart` seeds a year of data with a guaranteed streak, so trends and the diary look like a real account.
- **`seedDemoData` sets `ConfigEntity.isDemoData`**, which pins the demo banner to every tab. `setIsDemoData(false)` was called locally after seeding purely for the capture — **that edit is not committed**. The onboarding "try it with sample data" path shares the flag and is equally unusable for screenshots.
- Accent set to preset 07, the brand green. Material You was picking up the emulator wallpaper and rendering the app blue, which would have clashed with the green icon and feature graphic.
- The capture loop asserts the app owns the focused window before each screencap — without it, a stray back-press had already produced a full set of the Android launcher.

## Reviewer notes

- **The lane has never executed.** No Ruby, fastlane, or Play credentials locally. The YAML and the guard were tested for real (padded set passes; 1080x2205 rejected on ratio; RGBA rejected on alpha; 540x1102 rejected on size; single file rejected on count) — but `upload_to_play_store` is unexecuted. The first `validate` run is its first real test.
- **`track: 'production'` is a judgement call.** Screenshots are not track-scoped, but `supply` needs a track for the edit. First thing to change if validate rejects.
- **The button appears only once this reaches `main`**, since GitHub surfaces `workflow_dispatch` from the default branch. This targets `develop` per the repo's CI convention.
- **These are honest raw captures, not a designed set.** The shot list, ordering and any captions belong to #1072.
- This fixes the **Play** half of #1082 only. The App Store `[Alpha]` screenshot needs iOS captures.

Refs #1082, #1062, #942, #1074, #1072.


simonoppowa:
All four findings addressed in `6496965e`. One of them meant this workflow could never have worked, so thank you for it.

## The ratio finding was correct, and fatal

Play's spec: *"The maximum dimension of your screenshot can't be more than twice as long as the minimum dimension."* The committed set was **1080x2205** — 2205 is **2.042x** 1080 — so my guard passed assets the Play API would have refused. Verified against Google's own requirements page.

Both the workflow guard and the lane now check the ratio, plus the 24-bit-no-alpha rule and the 3840px ceiling. Tested: the padded set passes; the pre-padding 1080x2205 geometry is rejected with the ratio named; an RGBA PNG is rejected on alpha; `docs/screenshots` at 540x1102 and a single-file directory are still rejected.

**On the suggested remedy** — the review proposed cropping/resizing to at most 1080x2160. I padded to **1104x2205** instead (ratio 1.997). Cropping 45px off a 2205px-tall screenshot clips the navigation bar; padding keeps every pixel. The added 12px columns are **edge-replicated** rather than filled flat, so the seam is invisible against card edges and shadows, not just against a uniform background — I checked a zoomed crop of both edges before committing.

## `release-ship` — correct, and the reasoning was sharper than mine

Worth recording why it matters here, since it is not the secret-availability argument: `GOOGLE_PLAY_SERVICE_ACCOUNT_JSON` **is** repository-scoped, so the job would have run fine without the environment. That is exactly the problem — it would have reached the store without the approval pause and without the branch policy that `docs/RELEASING.md` describes and that `android-deploy` already selects. A store write is a deployment. Added.

## Lane-side validation

Agreed, and for the stated reason: the lane is runnable directly, and a safety property living beside the caller instead of beside the upload is one refactor from being gone. The lane now enforces the same count, geometry and format rules itself.

## Exact error string

Now `grep -qF` against the full string, mirroring `default_workflow.yml`, so an unrelated failure is no longer relabelled as the known #942 defect and pointed at the wrong issue.

Also updated `docs/fdroid-submission-feasibility.md`, which recorded the old 1080x2205 geometry.

simonoppowa:
All three addressed in `0a29bfa4`.

## The 24-bit check was a real hole

The guards claimed to enforce Play's *"JPEG or 24-bit PNG (no alpha)"* and only enforced the *no alpha* half. Rejecting colour types 4 and 6 left grayscale, palette and 16-bit PNGs passing the preflight and being refused by Play afterwards — precisely the outcome the guard exists to prevent, and the claim in my own comment was wider than the code behind it.

Both the workflow and the lane now require **bit depth 8 and colour type 2**, and say what they actually found. Tested against generated files:

| Input | Result |
| :-- | :-- |
| 8-bit grayscale | rejected — `8-bit grayscale PNG; Play requires 24-bit (8-bit truecolour) PNG` |
| 8-bit palette | rejected — `8-bit palette PNG; ...` |
| 16-bit grayscale | rejected — `16-bit grayscale PNG; ...` |
| RGBA | rejected on alpha |
| the committed set | passes |

## The feature-graphic comment was over-certain, and I contradicted myself

[#1074](https://github.com/simonoppowa/OpenNutriTracker/issues/1074) records that the live 1024x500 figure is **not independently confirmed** — Google no longer renders the feature graphic on the desktop web listing, so it rests on the [#1063](https://github.com/simonoppowa/OpenNutriTracker/issues/1063) audit alone. I wrote that caveat on the ticket and then dropped it here and stated it as fact. Now attributed with the caveat attached. That is the difference between a note and folklore a later reader inherits as settled.

## The stale geometry

Both flagged comments hard-coded 1080x2205 — which **this branch itself replaces** with 1432x2856 fresh captures. The ratio rule is now explained generically, since it is a property of Play's spec rather than of whichever assets happen to be committed. I applied the same fix to the file header, which described the old set too.

---

Note the branch was force-pushed earlier to squash out redundant churn: the screenshots were being padded and then replaced, so the PNGs changed twice. It is now three commits — mechanism, assets, this fix — and the PNGs are touched once. Earlier inline threads may render as outdated against the rewritten history; the tree was byte-identical across the squash.

simonoppowa:
@codex review

simonoppowa:
Fixed in `1fab5095` — and this is the third round the same claim has survived, which is the interesting part.

Copilot found the live-feature-graphic assertion still stated as fact in two more places: the lane's rationale comment and the **job step summary**. I corrected the workflow header last round and left those behind, so the claim appeared three times in three registers — and the step summary is the one most readers actually see, which makes it the worst of the three to have missed.

Both now say *reported by the [#1063](https://github.com/simonoppowa/OpenNutriTracker/issues/1063) audit, not independently confirmed*, matching [#1074](https://github.com/simonoppowa/OpenNutriTracker/issues/1074). Grepped for the phrasing afterwards rather than fixing the two that were pointed at: there are no remaining assertions of compliance anywhere in either file.

No behaviour change — `skip_upload_images` stays true regardless. The value is that a reader does not inherit an unverified figure as settled, and whoever opens the console knows there is something to verify.

Codex reviewed `0a29bfa48f` and found no major issues.

simonoppowa:
@codex review

Issue #992, opened 2026-08-31
Three first-run gaps in the AI feature: hidden camera, hint lands wrong, photo failure offers no route


simonoppowa:
## Parent

Part of the AI-assisted meal logging feature ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)). Found in a release-readiness audit before [#988](https://github.com/simonoppowa/OpenNutriTracker/pull/988). **None blocks the merge**, and the reasoning for holding them is below rather than assumed.

## What to build

Three gaps in the paths a first-time user actually walks. Each is small; together they are the difference between the feature being discoverable and being stumbled into.

### 1. Configuring from inside the bulk-add screen leaves the camera hidden until you leave and come back

The screen resolves camera availability once, into a `late final` future. A user who follows the hint into Settings, configures a provider and returns has a screen that still believes nothing is configured — the camera stays hidden until the screen is re-entered.

This is the **first-run funnel of the flagship feature**: the most likely path is exactly the one that ends with the camera missing.

Held out of this release deliberately: converting two `late final` futures into `setState`-refreshed state touches the screen's initialisation, and this is not the week to do it — three changes have already landed in the amount-withholding logic nearby ([#975](https://github.com/simonoppowa/OpenNutriTracker/pull/975) over-reached, [#978](https://github.com/simonoppowa/OpenNutriTracker/pull/978) narrowed it back, [#981](https://github.com/simonoppowa/OpenNutriTracker/pull/981) followed).

### 2. The discovery hint drops a first-run user at the top of Settings

The failure notices learned to land on the AI dialog directly ([#852](https://github.com/simonoppowa/OpenNutriTracker/issues/852)). The **discovery** hint — the one a user who has configured nothing sees — still lands at the top of Settings, where the AI row is far down a long screen.

A one-argument change to a single `onTap`, with the precedent two hundred lines below it in the same file. The screen currently sends its two audiences to different places, and the one that knows less gets the worse route.

### 3. Photo failures name Settings but offer no way there

The text path's failure notice carries an action button to Settings. The photo path's identical text does not, so a user told to check Settings has to find it themselves.

Mechanically similar to the text branch, but `_centeredMessage` does not currently take an action, so it is a slightly wider change than it looks — worth doing carefully rather than quickly.

## Why none of these blocks 2.2.0

[...]

Issue #852, opened 2026-08-25
Send the failure notice to the AI settings, not the top of Settings


simonoppowa:
## Parent

Part of the AI-assisted meal logging feature ([#599](https://github.com/simonoppowa/OpenNutriTracker/issues/599)), staged on `feature/ai-assisted-meal-logging`. Found during the device acceptance pass, [#830](https://github.com/simonoppowa/OpenNutriTracker/issues/830).

## What to build

When a model read fails, the notice on the multi-item add screen offers a **Settings** action. It navigates — verified on a Pixel 6 — but it lands at the *top* of Settings, on Units & Energy.

The AI tile is in the Data group, near the bottom, below Food sources and Health Connect. So the recovery path from "your server did not answer in time" is: tap Settings, then scroll past four category groups to reach the only control that could fix it.

That group is far enough down that finding it takes a deliberate scroll to the end of the list — during this acceptance pass it took several attempts to reach even while looking for it.

The notice already knows the failure is about the AI configuration; that is why it offers the action at all. It should land the user on the thing it is telling them to change.

Route the action to the AI settings directly — opening the AI dialog, or at minimum scrolling the AI tile into view. This applies to every failure that offers the action, which per the failure enum means a rejected key, an unsupported model, a billing failure, a timeout, and a refused insecure destination — each of which is fixed in that one dialog.

## Acceptance criteria

- [ ] The notice's action lands the user on the AI settings, not the top of Settings
- [ ] It behaves the same for every failure kind that offers the action
- [ ] Back from there returns to the add screen with the rows still on it, since the parser rows survive the failure and are the reason the user is still there
- [ ] Mutation-checked: routing it back to plain Settings fails a named test

## Blocked by

- None — can start immediately.

simonoppowa:
Fixed in [#858](https://github.com/simonoppowa/OpenNutriTracker/pull/858), merged to `feature/ai-assisted-meal-logging`. CI green.

The notice now asks for the AI settings by name — `SettingsScreenArguments(openAiAssist: true)` — and the settings screen opens the dialog itself, instead of dropping the user at Units & Energy four category groups above the row they need.

Two things that turned out to matter more than the routing:

- **The dialog opens after the frame, not during it.** `didChangeDependencies` runs inside a build and `showDialog` pushes a route.
[...]

PR #1086, opened 2026-09-05
fix(bulk-add): re-read the AI setup on return, and send the hint to the AI dialog

simonoppowa:
Closes items **1 and 2** of #992. Two commits, because they are two defects that happen to share a line.

## The camera stays hidden after configuring from inside the screen

The screen resolved its three AI answers — `_photoAvailable`, `_modelUnknown`, `_photoDestination` — into `late final` futures, computed once. A first-time user who followed the in-screen hint into Settings, configured a provider and came back still saw no camera until they left the screen entirely and re-entered.

**That is the first-run funnel of the flagship feature**, and the most likely path through it ended with the feature apparently absent.

Both routes out of the screen now go through one `_openSettings({arguments})` helper that awaits the push, checks `mounted`, and re-resolves the three futures in `setState`. The IIFE initialisers became named methods so they can be called again; the bodies are unchanged.

## The hint sent the wrong audience the long way round

The failure notices have deep-linked to the AI dialog since #852. The discovery hint pushed plain Settings, which opens several category groups above the AI row — so of this screen's two audiences, **the one that knows least about where it is going got the worse route**. It now asks for the dialog by name.

## Verification

58 tests pass. Both changes were mutation-checked: reverting the re-resolve fails the camera test on the post-return `bulk-add-photo` expectation, and reverting the deep-link fails the new route-arguments assertion. The test pins **both** call sites' arguments, so moving one without the other fails here rather than quietly splitting the audiences again.

A doc comment on `_photoAvailable` claimed the user visiting settings "closes this screen anyway" — the false premise behind the bug. Corrected rather than left documenting the defect as intent.

## One judgement call

The fix holds while `_openSettings` stays the only route out of this screen that can change the keystore. That invariant is in its doc comment but nothing enforces it: a fourth entry point added with a bare `Navigator.pushNamed` reintroduces the bug silently. A `RouteObserver` would enforce it structurally, but needs one registered on `MaterialApp` — wider than this fix. Flagging it as the trade rather than hiding it.


simonoppowa:
Both review findings addressed.

**Copilot — the test's `reason:` contradicts its expectation.** Correct: it asserts the camera *is* shown and then explains the old behaviour. Fixing.

**Codex — an own-server probe that lands after the user returns.** Verified and real, but narrower than this PR. `_resolvePhotoDestination` routes own-server through `_ownServerDestination`, which reads the probe verdict; a user who saves an address and returns before the probe finishes gets `unknown` and the camera stays hidden, and this PR's single trigger — a route popping — has already fired by then.

The waiting itself is deliberate: the camera is offered only once a photograph has actually been read. What is missing is a second trigger for the verdict arriving, which is a different mechanism from the one here. Filed as [#1089](https://github.com/simonoppowa/OpenNutriTracker/issues/1089) rather than widened into this PR, which already carries two defects.

PR #1087, opened 2026-09-05
fix(bulk-add): give the photo failure the route its text already names

simonoppowa:
Closes item **3** of #992.

The text path's failure notice carries an action to the AI settings. The photo path shows the same words through `_centeredMessage`, which took no action — so a user told to check Settings had to go and find it.

`_centeredMessage` now takes an **optional** action, leaving its other call sites untouched, and the photo failure passes the same label and destination the text path already uses.

**No new string.** The label the text path shows is the one this needed — worth saying because the alternative was nine ARB files for a button that already had a name.

66 tests pass; analyze clean.

### DIFF
--- a/lib/features/add_meal/presentation/screens/bulk_add_screen.dart
+++ b/lib/features/add_meal/presentation/screens/bulk_add_screen.dart
@@ -401,6 +401,33 @@ class _BulkAddScreenState extends State<BulkAddScreen> {
         BulkAddPhotoError.insecureDestination =>
           S.of(context).bulkAddModelInsecureServerLabel,
         BulkAddPhotoError.billing => S.of(context).bulkAddPhotoNoCreditLabel,
+      },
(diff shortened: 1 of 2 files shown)


simonoppowa:
Reviewed and **not reproduced**, with a test added so the claim can be checked rather than argued.

The concern was that the new `SingleChildScrollView` breaks vertical centring for short messages, because it "expands to the available height". It does not: `Center` passes loose constraints, so the scroll view shrink-wraps its child and `Center` centres the result.

Nothing pinned that, though, so the claim could have become true later without a test noticing. Added one asserting the scroll view's centre against the region it is given.

It took two attempts to make honest. The first version compared against the whole screen with a generous tolerance and **passed under an `Align(topCenter)` mutation** — a test that could not fail. The committed version compares against the enclosing region, where top-aligning moves the box by half the region, and that mutation is caught.

PR #1088, opened 2026-09-05
fix(bulk-add): report every bad row, and say what is wrong with each

### LINKED ISSUE #1013: Bulk add: one invalid quantity blocks the whole batch, with a message that does not say why
`_onSubmitPressed` in `lib/features/add_meal/presentation/screens/bulk_add_screen.dart` validates every row before writing any of them, and `return`s on the first one that fails:

```dart
for (final row in rows) {
  final text = row.amountText.trim();
  final quantity = double.tryParse(text.replaceAll(',', '.'));
  if (!_quantityPattern.hasMatch(text) || quantity == null ||
      quantity <= 0 || quantity > _maxQuantity) {
    ScaffoldMessenger.of(context).showSnackBar(SnackBar(
      content: Text('${row.meal?.name ?? row.resolved.parsed.query}: '
                    '${S.of(context).quantityLabel}'),
    ));
    return;
  }
```

Validating the batch before writing is correct and deliberate — the comment above it explains that a mid-loop throw would leave earlier rows written with no rollback. Two things around it are not:

1. **One bad row refuses every good row beside it.** The user fixes one field and re-submits the whole batch.
2. **The message names the field and nothing else** — `mince: Quantity`. It does not say what is wrong with it, and the constraint it violates (at most two decimals) is not visible anywhere in the UI.

This was reachable until `fcb2323a`: a converted imperial quantity was prefilled at full precision, so `1 lb mince` prefilled `453.59237`, failed the pattern, and blocked the batch. That fix rounds the prefill, so the *known* route in is closed — but the batch-fatal behaviour and the uninformative message are still there for anything else that produces a non-conforming amount, including a hand-typed one.

## Suggested shape

- Collect every failing row rather than returning on the first, and mark them inline in the list so the user can see all of them at once.
- Say what is wrong: "at most two decimals", "must be greater than zero", "too large" are three different failures currently rendered identically.
- Keep the validate-then-write ordering exactly as it is.

Low priority — no data loss, and the path that made it easy to hit is closed.

simonoppowa:
Closes #1013.

`_onSubmitPressed` refused the batch on the **first** row that failed, with a snackbar naming the field and nothing else: `mince: Quantity`. A user with three bad amounts fixed one, resubmitted, found the next, and repeated — never learning which rule any of them broke. The decimal limit in particular is stated nowhere in the UI.

The check now collects every failure and marks the rows inline so they are visible together, and separates the three reasons an amount is refused — malformed, too small, too large. The snackbar names each row and its reason, because a marked row can be off screen.

## What did not change, on purpose

**The validate-then-write ordering.** `addIntake` parses with `double.parse` and no guard, so a bad row throwing mid-loop would leave the rows before it written with no rollback. Checking every row up front is what makes the batch all-or-nothing, and the issue asked for it to stay that way.

`quantityError` is deliberately **not** consulted by `willBeLogged`: a row the user mistyped is handed back to them, not silently dropped from the batch and logged around.

The pattern and bound moved to the bloc because three things have to agree about them and only one is a widget — the field's input formatter, the prefill rounding, and the error the row reports.

## Verification

Three new l10n keys across all nine locales, **translated rather than copied** — parity exact at 1034 each, and no locale carries the English string.

61 tests pass. Mutation-checked: restoring "return on the first bad row" fails 3 of the 5 new tests.

One of those tests was failing when I picked this up, and the cause was in the **test**, not the change: the refusal's snackbar sits over the submit bar, so the second `tap` landed on the snackbar instead of the button. Traced by instrumenting the bloc handler, the rebuild and the submit — the state was correct at every step and the second submit simply never ran. The test now clears the snackbar first, which is what a real user does by waiting.


simonoppowa:
Two findings, one real.

**Codex — the snackbar summary is unbounded.** Correct. Nothing limits how many rows a paste produces and a snackbar does not scroll, so a large batch would cover the very list the marks are on. Bounded to four lines with an ellipsis; the marks carry every failure, which is the point of the change.

**Copilot — a marked row that is then skipped keeps a stale mark.** I could not reproduce it, and I found that out the embarrassing way. I took the finding, added a `willBeLogged` guard at the render site and a test, and pushed both — then mutation-checked and found the test passes with the guard removed. Confirmed the mutation was live by grepping the guard out of the file before re-running. Something on the render path already declines to show the mark for a skipped row, so the guard changed nothing and the test asserted behaviour that already held.

Both reverted. Shipping a fix for a case I could not reproduce, with a test that passes either way, is worse than leaving it alone.

One caveat I have **not** verified: a unit change that makes `amountWouldBeWrong` true does not clear the mark either, and that path has no test. That is a different claim from the one made, and I am not asserting it — just noting where a reproducible version of this finding might live.

Issue #1062, opened 2026-09-04
Refresh both store listings for 2.2.0: accurate, evidence-led, and measured


simonoppowa:
## Destination

Both store listings — Google Play and the App Store — describe OpenNutriTracker **2.2.0** accurately in **en-US**, with copy and screenshots chosen against **ASO evidence rather than taste**, **applied live in both consoles**, and **measured** against a baseline taken first.

Reaching the end means the listings are live and a post-change read exists — not that a spec exists.

## Notes

**Domain.** Store listings for a Flutter app, not app code. This map touches `fastlane/metadata/`, two web consoles, and one CI workflow. It does not change app behaviour.

**Execution and measurement are both in scope — this map overrides wayfinder's plan-don't-do default twice, deliberately.** The stated problem is that the *live* pages are wrong, and a decided description nobody pasted leaves them wrong. Same precedent as [map #935](https://github.com/simonoppowa/OpenNutriTracker/issues/935), which took the store declarations all the way into the console. Both the applying and the measuring are their own gated tickets, not smuggled into the decisions.

### The starting position, established while charting so no ticket re-derives it

- **Play copy lives in the repo, and has never reached the store.** `full_description.txt` was rewritten for 2.2.0 on 2026-09-04 in [#1049](https://github.com/simonoppowa/OpenNutriTracker/issues/1049) — 3952 of Play's 4000 characters, and it lives on `develop`, not `main`. **Verified 2026-09-04:** the live Play short description is **51 characters** ("Simple and Open-Source Calorie and Nutrient Tracker") against the repo's 77, and the live full description is roughly **1,179 characters** against the repo's 3,952. The repo metadata is not a newer version of the live text — it is a *different* text that was never pushed.
- **Nothing in CI ever pushes it.** `android/fastlane/Fastfile` runs `upload_to_play_store` with `skip_upload_metadata`, `skip_upload_images` and `skip_upload_screenshots` all `true`. [`docs/RELEASING.md`](docs/RELEASING.md) makes the paste a manual checklist item, and that checklist is currently unticked.
- **The App Store listing exists nowhere in this repo.** No `fastlane/metadata/ios/`. `ios/fastlane/Fastfile`'s `release` lane uploads the IPA and nothing else. Subtitle, keywords and promotional text are console-only and unversioned — neither the Lookup API nor the public product page exposes them, so they need a console visit.
[...]

PR #1090, opened 2026-09-05
ci(play): put the dispatch-only screenshot lane on main

simonoppowa:
Main half of a hotfix pair. The develop half landed as [#1085](https://github.com/simonoppowa/OpenNutriTracker/pull/1085) (`53b1ede7`); this is the same change cherry-picked onto `main`.

## Why it needs to be on main at all

GitHub only surfaces `workflow_dispatch` for workflows present on the **default branch**. The lane merged to develop cannot be run until it is here, so the live Play listing keeps showing debug-build screenshots until this lands or until develop next reaches main on its own.

## Why it carries the screenshots too

Not optional. The job checks out whichever ref it runs from, and `main`'s committed set is **1080x2205** — a ratio of 2.042 against Play's rule that the longest edge may be at most twice the shortest. Shipping the workflow alone would put a lane on main that **its own guard rejects on every run**. Verified: the guard passes against this branch's tree.

## What was dropped from the pick

`docs/fdroid-submission-feasibility.md` — it does not exist on `main`, and the change to it was a one-line dimension record. Nothing else conflicted.

## ⚠️ Merging this starts the deploy chain

Worth stating plainly before anyone hits merge, because it is a property of `main` rather than of this change:

- `pubspec.yaml` on main is **2.2.0+63**, and the ledger tag **`deployed/63` does not exist**, so `release-gate` computes **`deploy=true`**. Any push to main right now begins the packaging and deploy jobs.
- `v2.2.0` **does** exist, so `publish=false` — `github-release`, the only destructive path in the workflow, will not run.
- `ios-deploy` and `android-deploy` sit in `release-ship`, which requires a manual approval, so nothing reaches a store without someone approving it.

So the realistic cost of merging is a packaging run and a pending deployment nobody should approve — not a release. But note the App Store is still on **v2.0.2**, so build 63 has genuinely never shipped to iOS; do not create a `deployed/63` ledger tag to silence the gate, because that would suppress a deploy that is still owed.

If that cost is unwanted, the alternative is to let this ride to main with the next release instead, and accept that the debug screenshots stay live until then.

## Unchanged from #1085

The lane uploads screenshots and nothing else — every other `skip_upload_*` is true, so it cannot touch the feature graphic, the descriptions or the changelogs. `validate_only` defaults to true. The job selects `release-ship`. Both the workflow and the lane enforce Play's count, geometry and format rules, including the 24-bit PNG requirement, before uploading.

Reviewed clean by Codex on the develop side at `1fab509550`.

Refs #1085, #1062, #942, #1074.


simonoppowa:
Fixed in `37824c87`, and the finding was right.

Both validators read 26 bytes and unpacked the IHDR without checking they got 26. Reproduced with a 20-byte file carrying the PNG signature:

```
Traceback (most recent call last):
  File "/tmp/gt2.py", line 19, in <module>
    w, h, depth, colour = struct.unpack('>IIBB', head[16:26])
struct.error: unpack requires a buffer of 10 bytes
```

It was already fail-safe — non-zero exit either way, so nothing unsafe could reach Play — but the whole point of validating locally is that the failure explains itself, and a traceback explains nothing. Now:

```
Rejected:
  a.png: truncated PNG — 20 bytes, too short to contain an IHDR header
```

Verified against a 20-byte truncated PNG, a zero-byte file (caught earlier by the signature check, still `not a PNG`), and the real set, which passes.

**Paired to develop as [#1091](https://github.com/simonoppowa/OpenNutriTracker/pull/1091)** — identical commit. The guard code landed on develop in #1085, so fixing only this branch would have left the two copies drifting apart, which is the failure mode the hotfix-pair convention exists to prevent. Both need to land.

Issue #1080, opened 2026-09-04
Does the App Store change ride a new build, or reach the live page without one?


simonoppowa:
## Question

Does the App Store change ride a new build, or reach the live page without one?

A fork the map did not have when it was charted, opened by the research on [#1067](https://github.com/simonoppowa/OpenNutriTracker/issues/1067) (branch `research/aso-metadata-editability`, `docs/aso-store-metadata-editability.md`). The two routes cost wildly different amounts and deliver different subsets of the change, so this is settled before anyone opens a console.

**Route A — cut a 2.2.1.** Everything changes: name, subtitle, keywords, description, screenshots, What's New. The cost is a full build, tag, pipeline run and App Review pass. Forced if 2.2.0 has already shipped, because Apple's *Choose a build to submit* rule means "the list only includes builds you have uploaded since that version was released on the App Store" — **there is no metadata-only version submission**; 2.2.0's binary cannot be re-attached to a 2.2.1 metadata-only record.

**Route B — product page optimization, no build.** Apple states that test-treatment metadata "can be submitted without submitting a new version of your app", and that a treatment may be applied to the original product page "at any time" — not conditional on the test reaching significance. That delivers **screenshots and app previews** to the live default page with no build. Promotional text (170 chars) is separately editable whenever. Everything else — description, subtitle, keywords, app name — stays stale until the next natural release.

**Corrected 2026-09-04 — build-free is not review-free.** New screenshot metadata still goes through App Review, and Apple bundles it: "Items in the submission will be reviewed together with your latest iOS version." With Apple's ceiling of two concurrent submissions per platform, that is a real scheduling constraint if a 2.2.x record is in flight. Route B is cheaper than Route A, not free.

**A precondition this ticket originally missed:** "Your app must be in the Ready for Distribution state to test its product page." Met here — the live record is 2.0.2 — but it is a gate, and it would not be met by an app that had never shipped.

**A detail in Route B's favour:** applying a treatment targets the live page *and* a version in Prepare for Submission in the same action, so screenshots delivered this way need not be re-uploaded into a later 2.2.1. And reordering screenshots already on the store, or changing only the icon, needs no resubmission at all.

**Still a console question:** whether a treatment can be applied without the test ever being *started*. Apple's wording is "at any time", and the 90%-confidence line is explicitly a recommendation rather than a gate — but nobody has confirmed the console permits apply-without-start.

**What decides it:**

[...]

Issue #1076, opened 2026-09-04
The iOS screenshot lane on macOS CI, dispatch-only


simonoppowa:
## Question

How does a macOS CI job produce the 6.9" iPhone and 13" iPad screenshot sets?

There is no Mac here — the dev machine is Linux — and Apple requires both device classes because `TARGETED_DEVICE_FAMILY = "1,2"`. The map settled that CI simulators do this rather than a borrowed Mac, so the next release gets the re-shoot for free.

**Hard constraint, and it is not negotiable.** The lane is **`workflow_dispatch`-only**. [Map #1016](https://github.com/simonoppowa/OpenNutriTracker/issues/1016) spent an entire effort removing macOS demand from ordinary PRs — macOS queue p90 fell 35.4 to 1.7 min, mean macOS jobs per run 3.06 to 2.17 — against a hard cap of 5 concurrent macOS runners. Adding this lane to `pull_request`, `release/**` or `main` triggers regresses that work. There is prior art for the simulator boot itself in `.github/workflows/ios-integration-attempt.yml`, including the `macos-15` pin, which exists because Flutter never discovers the VM service on the iOS 26 simulator.

Open questions this ticket resolves:

- **The driver.** `fastlane snapshot` expects a UI-test target; this is a Flutter app whose only integration test is `integration_test/app_boot_test.dart`, a single boot smoke. Whether to extend `integration_test/` with a screenshot driver run under `flutter drive`, or to add the XCUITest target `snapshot` wants, is the central decision.
- **Staging the content.** The shots need the demo fixture from [the shot list]. `lib/core/utils/demo/` ships in every build, but `lib/dev/main_dev.dart` throws in release mode — so establish which entry point and which flavor the simulator runs.
- **Getting the files out.** Upload as a workflow artifact for a human to commit, or have the job open a PR. Note that an auto-commit from CI has caused trouble here before: the iOS Podfile.lock auto-commit makes dependabot refuse to rebase.
- **Where they land** — `fastlane/metadata/ios/en-US/screenshots/`, per `deliver`'s conventions, alongside the copy from [the App Store copy ticket].
- Whether the same lane can also produce the Android set, or whether [the Play re-shoot] stays on the physical Pixel.

simonoppowa:
## Resolution

Shipped in #1100, against `develop`. **Not closed** — most of the diff has never run, and the first honest verification is the first dispatch after it merges.

[...]

Issue #1075, opened 2026-09-04
Re-shoot the Play screenshots on the Pixel via adb-driver.sh


simonoppowa:
## Question

How are the Play screenshots re-shot, and does the result hold up?

Android is the cheap half: there is a physical Pixel and a full ADB toolkit — `adb-driver.sh` with `tap_id`, `wait_for_id`, `enter_text_at`, `_tap_text`, `screenshot` and the form-field helpers, documented in `AGENTS.md`. Nothing equivalent exists for iOS, which is why [the iOS lane] is a separate and harder ticket.

Build the shoot as a **script**, not a session of manual taps, so the next release re-shoots for free:

- Seed the device from the demo fixture chosen in [the shot list], then walk to each screen in the list and capture.
- Native capture resolution, not the half-scale used for `docs/screenshots/` — those are 540x1102 and too small for the store. The current store set is 1080x2205.
- Apply whatever treatment [the shot list] settled: raw captures, or a compositing step for frames and captions.
- Write into `fastlane/metadata/android/en-US/images/phoneScreenshots/`, keeping the `N_en-US.png` naming that `deliver`/`supply` expects, even though nothing in CI uploads them.

**A device-safety note that has bitten before:** verify the app still has focus after every tap on the physical Pixel — a stray HOME keypress has previously opened an unrelated app mid-run, and a screenshot taken in that state is worse than no screenshot.

Consider whether `docs/screenshots/` — re-shot 2026-09-01 for the README, from a different session and at a different scale — should be regenerated from the same script, so the README and the store stop drifting apart.

simonoppowa:
**Stopgap filed separately: [#1082](https://github.com/simonoppowa/OpenNutriTracker/issues/1082).**

The live Play screenshots 1 and 2 are debug-build captures — Flutter's red `DEBUG` ribbon and an app bar reading "OpenNutriTracker [Alpha]", at 424x900. This ticket is three blockers deep (shot list ← positioning ← baseline), which is realistically weeks, and there is no reason to leave a debug build on the store page for that long.

#1082 covers uploading the repo's existing clean 1080x2205 set as an immediate replacement. It needs no decision this ticket owns, and it does **not** substitute for the real re-shoot — this ticket still owns that.

simonoppowa:
## Resolution

**Done, but by a different route than this ticket assumed, and out of the order the map intended.** Both deviations are stated here rather than smoothed over.

### What was shot, and how

[...]

Issue #1067, opened 2026-09-04
What Apple lets us change without a new version, and where 2.2.0 sits now


simonoppowa:
## Question

Which App Store listing fields can change without submitting a new version, and where does 2.2.0 currently sit in each console?

This decides the *shape* of the whole map's endgame, and it is a fact, not a preference.

**Apple's editability rules.** Establish from Apple's own documentation which App Store Connect fields are editable at any time and which are locked to a version submission. The working understanding — to be confirmed, not assumed — is that **promotional text** is editable whenever, while **description, keywords, screenshots and app previews** are version-gated, and **app name and subtitle** are too. If that holds, every substantive change this map makes to the App Store requires a version submission, which means either riding 2.2.0's submission or cutting a 2.2.1.

**Where 2.2.0 actually is.** `v2.2.0` was tagged 2026-09-01 and both store uploads are manual, so the state is genuinely unknown from here:

- App Store Connect — is 2.2.0 submitted, in review, approved-pending-release, or already live? If it is still editable pre-submission, the listing changes ride along for free. If it has shipped, this map needs a version to attach to.
- Play Console — which track holds build 63, and at what rollout percentage? Play's listing is not version-gated the way Apple's is, but a staged rollout interacts with any measurement design.

**Also settle Play's changelog gap while here.** `fastlane/metadata/android/en-US/changelogs/63.txt` covers only the TDEE correction and mentions neither AI meal assistance nor workout import — build 63 *is* 2.2.0. Whether the live Play "What's new" matches that file or something else is part of the answer.

Console access is required for the second half.

simonoppowa:
@/tmp/claude-1000/-home-simon-Documents-OpenNutriTracker/c06eb3af-b919-42b0-b842-370716bc684e/scratchpad/comment.md

simonoppowa:
## Play half answered; Apple half still blocked, and I could not unblock it

### Play: 2.2.0 is fully live

From the Production track, 2026-09-05:

> **Aktiv · Neuester Release: 63 (2.2.0) · 177 Länder/Regionen · 10.174 Installationen**

Build **63 is 2.2.0**, it is on the **production** track, rolled out to **100%** across 177 countries. There is no staged rollout in flight, so nothing about a listing change has to be sequenced around one — which removes one of the two sequencing worries this ticket raised.

[...]

PR #1100, opened 2026-09-05
ci(ios): add a dispatch-only lane for the App Store screenshot sets

simonoppowa:
Closes nothing on its own — resolves the design questions on #1076, which stays open for a human to close.

Adds a `workflow_dispatch`-only macOS lane that produces both Apple screenshot sets, plus the caption compositing step #1072 created and neither re-shoot path had.

## What was unexecuted, and what was not

Stating this first, because most of the diff is code that has never run.

**Never executed.** Everything macOS: `.github/workflows/ios-screenshots.yml`, `.github/scripts/ios_capture_screenshots.sh`, the simulator selection, `simctl status_bar override`, the app-container read-out, and `binding.takeScreenshot` on iOS. There is no Mac here and this workflow cannot be run from a fork or a branch — `workflow_dispatch` only offers branches once the workflow file is on the default branch, so **the first real run of this lane happens after it merges.** Expect to fix something on the first dispatch.

**Actually executed, on this machine:**

- `flutter analyze integration_test/` — clean. Every widget class, finder and signature the capture test uses resolves against the pinned 3.44.6 SDK, including `find.bySemanticsIdentifier`, `runAppWithChangeNotifiers`'s six positional arguments, and `DemoSeedOptions.dev`/`.onboarding`.
- **The capture test was launched on a real Android emulator** (`flutter test integration_test/store_screenshots_test.dart -d emulator-5554 --flavor full --dart-define=STORE_SCREENSHOTS=true`). It built, installed, launched, the binding came up, the `STORE_SCREENSHOTS` gate let the body run, and it executed as far as `initLocator()`. It then died inside `flutter_secure_storage` on an emulator Keystore fault (`Finish failed for AppUid(...) Error::Km(UNKNOWN_ERROR)`) — stale keystore state in that AVD, nothing to do with this change — and the emulator's CPU threads hung on the retry, so **the navigation and the six captures themselves are unverified.** What is verified is that the file is a loadable, runnable integration test that gets the app off the ground.
- **The skip-gate is verified on CI, not just reasoned about.** `android-integration-tests` on this PR runs `flutter test integration_test/ --flavor full` on an emulator and reported `00:00 +0 ~1` for `store_screenshots_test.dart` before finishing `+1 ~1: All tests passed!` — the `~1` is the skipped test. The new file loads into the combined suite, adds nothing to its runtime, and does not disturb `app_boot_test.dart`.
- `tools/screenshots/compose.py` — fully exercised, at 1290x2796, 2064x2752 and 1432x2856, captioned and `--no-captions`, over the six real PNGs already committed in `fastlane/metadata/android/en-US/images/phoneScreenshots/`. Output inspected by eye.
- The Apple spec guard — run against that output. Passes on 12 good files; correctly rejects an RGBA file and a 2048x2732 iPad file, naming the alternate size in the message.
- The workflow parses as YAML; the shell script passes `bash -n`.

## The driver: `integration_test/`, not `fastlane snapshot`

`snapshot` drives an XCUITest target and there is no XCUITest target in `ios/Runner.xcodeproj`. Adding one means editing `project.pbxproj` across six build configurations — blind, on a machine that cannot open Xcode — and then writing Swift that navigates a Flutter app, which from XCUITest is one opaque `FlutterView` whose accessibility tree is not what the repo's own tests address. What `snapshot` buys (status bar, `deliver` layout) is two `simctl` commands and a directory name.

The integration test buys three things `snapshot` cannot:

1. **It can fix the demo-banner trap in committed code.** More below.
2. **It navigates with the finders the repo already has.** `MainScreen`'s bottom-nav items already carry `Semantics(identifier: 'nav-home')` etc. — added for `tools/adb/adb-driver.sh`. Both drivers now agree on what a nav item is called.
3. **It runs unchanged on Android**, which is what makes the "can one lane do both stores" answer a real yes rather than a hopeful one.

**Refinement on the ticket's framing: `flutter drive` is not needed.** `binding.takeScreenshot()` returns the PNG bytes to the test on both platforms (`integration_test/lib/src/_callback_io.dart` invokes the platform channel directly; `flutter drive`'s `onScreenshot` is only how the *host* receives them). So the test writes the files into the app's own documents directory and the workflow reads them out with `xcrun simctl get_app_container`. That keeps the run a plain `flutter test -d <udid> --flavor full` — the exact command shape `ios-integration-attempt.yml` already runs green on `macos-15` — instead of adding `flutter_driver` to `dev_dependencies` and betting on a second tool path that has never run in this repo. Two lines of container lookup, with a `find` fallback, against a dependency-graph change and an unproven `flutter drive` iOS path.

## The demo-banner trap, solved without an uncommitted edit

`seedDemoData` ends with `setConfigIsDemoData(true)` (`lib/core/utils/demo/demo_seeder.dart:297`), which pins `DemoModeBanner` to the top of every tab in `MainScreen`. Both seeding routes do it, so there is no seeded state in the app that does not carry the banner. #1075 got around it with a local `setIsDemoData(false)` that was never committed.

The test does it in committed code, and then **asserts it**: `expect(find.byType(DemoModeBanner), findsNothing)` runs once after boot and again before every single capture. If a future change reintroduces the banner the run fails instead of quietly producing six banner-topped assets. The same `_shoot` helper also refuses to capture when its subject widget is not on screen or a `CircularProgressIndicator` is still up — the iOS counterpart of the focus assertion #1075 needed, where a stray back-press produced a full set of the Android launcher.

No production code changed. Theme, locale, energy unit, Material You and the brand accent (preset 07, `0xFF43A047`) are pinned by calling `runAppWithChangeNotifiers` directly rather than through `main()`, so a set never depends on what a runner's simulator happened to persist.

## Captions

`tools/screenshots/compose.py`, deliberately outside the app and outside either platform's lane. It takes raw captures and a target size, paints a caption band, scales the capture into what is left, and flattens to RGB with no alpha — which is what *both* stores require and what the raw `UIGraphicsImageRenderer` output is not.

Copy lives in `tools/screenshots/captions.en-US.json`. The strings there are **#1072's drafts verbatim**, marked as drafts; #1072 hands final wording to the copy tickets, and they are claims under Play's Misleading Claims policy, so they belong in a reviewable data file rather than inside a script. `band_fraction` defaults to 0.18 and the script refuses anything over 0.20, because Play caps a tagline at 20% of the image; Apple 2.3.3 sets no limit, so one value serves both.

**A useful consequence: the Play set's outstanding second pass is this script, not a re-shoot.** #1072 says the Play captures that shipped in #1085 need captions. They are already in the repo at 1432x2856. `compose.py --raw fastlane/metadata/android/en-US/images/phoneScreenshots --size 1432x2856` is the whole job.

## The other three questions

**How the files leave the runner: a workflow artifact.** Not an auto-commit and not an auto-opened PR. Partly the precedent — the iOS `Podfile.lock` auto-commit puts a bot commit on the branch tip and dependabot then refuses to rebase, recoverable only with `@dependabot recreate` — but mostly because six images going on the front page of a store listing is a judgement call. Somebody has to look at them. The artifact carries the raw captures alongside the composited ones, so a caption problem can be fixed locally without re-running a macOS job.

**Where they land:** `fastlane/metadata/ios/en-US/screenshots/iphone-6.9/` and `.../ipad-13/`, per the map. One note for whoever wires up an upload later: `deliver`'s *default* screenshot path is `fastlane/screenshots/<locale>/`, so `ios/fastlane/Fastfile` will need an explicit `screenshots_path:` if it ever grows an `upload_to_app_store` call carrying images. It has none today — the `release` lane uploads the IPA and nothing else.

**Can this lane also do Android? The compositing yes, the capture no — and it should not.** `store_screenshots_test.dart` runs unchanged against an emulator (`convertFlutterSurfaceToImage()` is called unconditionally; it is a documented no-op on iOS). But an Android emulator job belongs on `ubuntu-latest`, and hanging one off a macOS runner recreates exactly the demand #1016 removed. Shared test, shared compositor, separate lane if one is ever wanted.

## Traps found that the ticket did not list

1. **`flutter test integration_test/` runs the whole directory, and two jobs do that.** `android-integration-tests` runs it on **every pull request** (ubuntu-latest), and `ios-integration-attempt.yml` runs it on every push to `develop` and `main` — the iOS one is excluded from pull requests, which is #1016's work. Dropped in without a guard this file would have joined both, seeded a year of demo data into each, and made them slower and flakier for nothing. Hence the `--dart-define=STORE_SCREENSHOTS` gate: without it the test is `skip`ped at compile time, and nothing but this lane passes it. (The initial commit's comment named the wrong job; corrected on the branch.)
2. **The debug-build ribbon, which turns out to be a non-problem — but only by luck.** `flutter test` builds a *debug* app, and #1082's complaint was a red `DEBUG` ribbon on the live Play listing. `lib/main.dart:287` sets `debugShowCheckedModeBanner: false`, so there is no ribbon. Worth knowing that this lane's safety here rests on one line of app code: if it ever flips, this lane starts shipping ribboned assets. `--flavor full` covers the other half of that defect, the `[Alpha]` app bar.
3. **A wrong-sized capture would pass the spec guard.** The compositor forces its output to the target size whatever it is handed, so a capture from an iPhone SE would be upscaled into a soft 1290x2796 asset that the Apple guard then waves through — the same class of failure as the 1242x2208 set currently live. The capture script therefore checks the **native** capture size, where the device that produced it is still known, and fails on anything smaller than the slot (warning-only on Apple's larger accepted size, which is a legitimate downscale).
4. **`convertFlutterSurfaceToImage()` must be called after the app is up**, not before. The package's own example does it after `app.main()`; calling it first has nothing to convert.
5. **`pumpAndSettle`'s first argument is the interval between frames, not a timeout**, and under the live binding that interval is real elapsed time — `app_boot_test.dart`'s `pumpAndSettle(Duration(seconds: 30))` means "wait 30 seconds", not "give up after 30". Worse, a *bare* `pumpAndSettle()` returns as soon as no frame is scheduled, which is exactly the state a page is in while its bloc awaits its first load — so it happily returns on an empty screen. Fixed in the second commit: each navigation and each capture pumps for a real interval before settling, and `_shoot` refusing to capture over a `CircularProgressIndicator` is the backstop.

## Known risks on the first dispatch

- **Simulator device types.** The candidate lists are `iPhone 16 Plus | iPhone 15 Plus | ...` (430x932 @3x = 1290x2796) and `iPad Pro 13-inch (M4)` (1032x1376 @2x = 2064x2752). Whether those exact names exist on the current `macos-15` image is unverified. On a miss the step prints every device the runner does have, which is the one thing that makes it a two-minute fix rather than a mystery.
- **The status bar.** `IntegrationTestPlugin.m` renders `scene.windows`; on modern iOS the status bar is not one of them, so the captures may come back with an empty status-bar strip. That is not a defect — Apple does not require one, and the caption band sits over the top 18% anyway — but it is why `simctl status_bar override` is best-effort rather than load-bearing. It costs one command and it is what stops the `Carrier` placeholder on two of the live assets coming back.
- **The VM-service hang.** Same one `ios-integration-attempt.yml` documents and the reason for the `macos-15` pin. Rarer there, not absent. A re-run on a fresh runner has historically fixed it; the summary step says so.
- **iPad framing.** With an 18% caption band, a native 2064x2752 capture scales to about 1662px wide inside a 2064px canvas, so the iPad shots carry visible side margins. That reads as a deliberately framed marketing image rather than a defect, but it is a real look and worth an opinion before the set is committed.

## How you actually run it, which is not today

GitHub only offers a `workflow_dispatch` workflow once its file is on the **default branch**, and this repo's default is `main`. This PR targets `develop`, so **merging it does not make the lane runnable** — the Run workflow button appears only after the file reaches `main`. That is the same path `play-screenshots.yml` took: merged to `develop` as #1085, then carried to `main` by the #1090/#1091 hotfix pair, and only then dispatchable.

So the first real execution of any of this is gated on a `main` landing, which is worth knowing before anyone schedules the Apple re-shoot against it.

## What this does not do

Nothing here publishes. Apple's metadata on the shipped 2.2.0 record is locked (#1067) and screenshots go via Product Page Optimization, which is #1080's console action. This job holds no credentials, selects no environment, and is deliberately not a deployment — unlike `play-screenshots.yml`, which needs `release-ship` because it writes to a live listing.

### DIFF
--- a/.github/scripts/ios_capture_screenshots.sh
+++ b/.github/scripts/ios_capture_screenshots.sh
@@ -1,0 +1,190 @@
+#!/usr/bin/env bash
+#
+# Capture one device class's raw store screenshots on an iOS simulator.
+#
+#   ios_capture_screenshots.sh <slot> <out-dir>
+#
+#     <slot>     iphone | ipad — only used for the expected native size and
+#                for log text.
+#     <out-dir>  where the raw PNGs are copied to.
+#
+# Environment:
+#
+#     CANDIDATES  '|'-separated simulator device names, in preference order.
+#     BUNDLE_ID   the app's bundle identifier (must match the built flavor).
+#     SCREENSHOT_FIXTURE  optional: dev (default) or onboarding.
+#
+# Called twice by .github/workflows/ios-screenshots.yml. Extracted into a
+# script rather than inlined because the two device classes run exactly the
+# same twelve steps and an inlined copy of them would drift.
+#
+# The two things here that are not obvious:
+#
+#   1. The device type is chosen from a preference list, not hardcoded.
+#      Runner images add and drop simulator device types between updates, and
+#      a hardcoded name turns an image refresh into a mystery failure. When
+#      none of the candidates is present, this prints every device the runner
+#      *does* have, which is the only thing that makes that diagnosable from
+#      a log.
+#
+#   2. The captures are read out of the simulator's app container rather than
+#      handed back over a `flutter drive` channel. That keeps the run a plain
+#      `flutter test -d <udid>` — the exact command shape
+#      ios-integration-attempt.yml already runs green on this image — instead
+#      of adding flutter_driver and a second, unproven tool path on top of
+#      everything else here that has never executed.
+
+set -euo pipefail
+
+SLOT="${1:?usage: ios_capture_screenshots.sh <iphone|ipad> <out-dir>}"
+OUT_DIR="${2:?usage: ios_capture_screenshots.sh <iphone|ipad> <out-dir>}"
+: "${CANDIDATES:?CANDIDATES must list simulator device names, '|'-separated}"
+: "${BUNDLE_ID:?BUNDLE_ID must be set}"
+FIXTURE="${SCREENSHOT_FIXTURE:-dev}"
+
+case "$SLOT" in
+  iphone) EXPECT_W=1290; EXPECT_H=2796; LABEL='6.9" iPhone' ;;
+  ipad)   EXPECT_W=2064; EXPECT_H=2752; LABEL='13" iPad' ;;
+  *) echo "Unknown slot '$SLOT' (want iphone or ipad)" >&2; exit 2 ;;
+esac
+
+echo "==> $LABEL: choosing a simulator"
+UDID=$(
+  CANDIDATES="$CANDIDATES" python3 -c "
+import json, os, subprocess, sys
+
+wanted = [n.strip() for n in os.environ['CANDIDATES'].split('|') if n.strip()]
+data = json.loads(
+    subprocess.run(
+        ['xcrun', 'simctl', 'list', 'devices', 'available', '-j'],
+        capture_output=True, text=True, check=True,
+    ).stdout
+)
+available = [
+    d for runtime in data['devices'].values()
+    for d in runtime if d.get('isAvailable')
+]
+by_name = {}
+for d in available:
+    by_name.setdefault(d['name'], d)
+
+for name in wanted:
+    if name in by_name:
+        print(by_name[name]['udid'])
+        sys.exit(0)
+
+sys.stderr.write(
+    'None of the candidate device types is available on this runner.\n'
+    'Wanted, in order: ' + ', '.join(wanted) + '\n'
+    'Available:\n'
+)
+for name in sorted(by_name):
+    sys.stderr.write(f'  {name}\n')
+sys.exit(1)
+"
+)
+NAME=$(xcrun simctl list devices -j | python3 -c "
+import json, sys
+udid = '$UDID'
+data = json.load(sys.stdin)
+for runtime in data['devices'].values():
+    for d in runtime:
+        if d['udid'] == udid:
+            print(d['name'])
+")
+echo "    $NAME ($UDID)"
+
+cleanup() {
+  xcrun simctl shutdown "$UDID" >/dev/null 2>&1 || true
+}
+trap cleanup EXIT
+
+xcrun simctl boot "$UDID"
+xcrun simctl bootstatus "$UDID"
+
+# Apple's canonical status bar. Two of the live iPhone assets show the
+# simulator's `Carrier` placeholder (#1072); this is what stops that coming
+# back, and it costs one command. Best-effort: if a future simulator drops
+# `status_bar override` the run should still produce screenshots.
+xcrun simctl status_bar "$UDID" override \
+  --time '9:41' \
+  --dataNetwork wifi --wifiMode active --wifiBars 3 \
+  --cellularMode active --cellularBars 4 \
+  --batteryState charged --batteryLevel 100 \
+  || echo "::warning::simctl status_bar override failed; the status bar will be whatever the simulator felt like"
+
+echo "==> $LABEL: running the capture test"
+# --flavor full so the app carries the production bundle id and display name.
+# A screenshot of a build whose app bar reads "[Alpha]" is the defect this
+# whole lane replaces.
+flutter test integration_test/store_screenshots_test.dart \
+  -d "$UDID" \
+  --flavor full \
+  --dart-define=STORE_SCREENSHOTS=true \
+  --dart-define="SCREENSHOT_FIXTURE=$FIXTURE" \
+  --reporter expanded
+
+echo "==> $LABEL: reading the captures out of the simulator"
+SRC=""
+if CONTAINER=$(xcrun simctl get_app_container "$UDID" "$BUNDLE_ID" data 2>/dev/null); then
+  SRC="$CONTAINER/Documents/store_screenshots"
+fi
+# Fallback for the case `get_app_container` cannot resolve the bundle id —
+# a flavor mismatch, or flutter_tools having uninstalled the app after the
+# run. The data container survives either way, so find it by name.
+if [ -z "$SRC" ] || [ ! -d "$SRC" ]; then
+  echo "    app container lookup failed for $BUNDLE_ID; searching the device's data dir"
+  SRC=$(find "$HOME/Library/Developer/CoreSimulator/Devices/$UDID/data/Containers/Data/Application" \
+        -maxdepth 3 -type d -name store_screenshots 2>/dev/null | head -n 1 || true)
+fi
+if [ -z "$SRC" ] || [ ! -d "$SRC" ]; then
+  echo "No captures were written. The test reported success but nothing reached" >&2
+  echo "the app's Documents directory, which means the capture step never ran." >&2
+  exit 1
+fi
+
+mkdir -p "$OUT_DIR"
+cp "$SRC"/*.png "$OUT_DIR/"
+
+# The compositor forces its output to the target size whatever it is handed,
+# so a capture taken on the wrong device would be silently up- or downscaled
+# and would still pass the Apple spec guard afterwards. That is precisely how
+# a listing ends up with soft, wrong-sized assets, so check the *native*
+# capture here where the device that produced it is still known.
+EXPECT_W="$EXPECT_W" EXPECT_H="$EXPECT_H" OUT_DIR="$OUT_DIR" \
+LABEL="$LABEL" NAME="$NAME" python3 -c "
+import glob, os, struct, sys
+
+want_w = int(os.environ['EXPECT_W'])
+want_h = int(os.environ['EXPECT_H'])
+files = sorted(glob.glob(os.path.join(os.environ['OUT_DIR'], '*.png')))
+label = os.environ['LABEL']
+name = os.environ['NAME']
+
+if not files:
+    sys.exit(f'{label}: no PNGs copied out')
+
+bad = []
+for f in files:
+    head = open(f, 'rb').read(26)
+    if len(head) < 26 or head[:8] != b'\x89PNG\r\n\x1a\n':
+        bad.append(f'{os.path.basename(f)}: not a readable PNG')
+        continue
+    w, h = struct.unpack('>II', head[16:24])
+    print(f'  {os.path.basename(f):24s} {w}x{h}')
+    if w < want_w or h < want_h:
+        bad.append(
+            f'{os.path.basename(f)}: {w}x{h} is smaller than the {want_w}x{want_h} '
+            f'{label} slot. Compositing would upscale it into a soft asset. '
+            f'The simulator that produced this was \"{name}\".'
+        )
+    elif (w, h) != (want_w, want_h):
+        print(f'::warning::{os.path.basename(f)} is {w}x{h}, not {want_w}x{want_h}; '
+              f'it will be downscaled to fit the {label} slot')
+
+if bad:
+    sys.exit(f'{label} captures rejected:\n  ' + '\n  '.join(bad))
+print(f'{len(files)} raw capture(s) OK for {label}.')
+"
+
+echo "==> $LABEL: done"
--- a/.github/workflows/ios-screenshots.yml
+++ b/.github/workflows/ios-screenshots.yml
@@ -205,7 +205,14 @@ jobs:
           .github/scripts/ios_capture_screenshots.sh \
             ipad "build/screenshots/raw/ipad"
         env:
-          CANDIDATES: 'iPad Pro 13-inch (M4)|iPad Pro (12.9-inch) (6th generation)'
+          # Only the 13-inch M4 — 1032x1376pt at @2x is exactly 2064x2752.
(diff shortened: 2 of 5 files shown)


simonoppowa:
Addressed in `eb08da60`. Three findings applied, one pushed back on.

## The accent bug was real and would have shipped

`runAppWithChangeNotifiers` forwards its sixth argument straight to `Color(accentColor)` in `lib/main.dart:271`, so the preset **index** `7` became `Color(0x00000007)` — transparent near-black. Every capture would have rendered with a broken accent, and the pin that exists specifically to stop Material You turning the app blue would have produced something worse.

Now reads `accentPresetColors[7].toARGB32()` from the app's own palette, so the constant and the colour cannot drift.

## The visibility guards were satisfied by invisible widgets

Correct, and it defeated the guard's whole purpose. `MainScreen` keeps every tab alive in an `IndexedStack`, so an inactive tab's widgets stay in the tree and `findsWidgets` passed for a screen that was not on screen — the exact failure mode this guard exists for, since the Android pass once produced six shots of the Android launcher.

Both the subject check and the spinner check now use `hitTestable()`. The spinner one also stops false-failing on a loader in a tab nobody is looking at.

## The advertised Play second pass could not run

compose.py looked captions up by filename stem, but the committed Play set is `1_en-US.png`..`6_en-US.png` while the caption keys are `01-home`..`06-profile`. Every file failed. It now resolves by stem first, then by leading shot number.

Verified by running the documented invocation over the real six PNGs — six captioned 1432x2856 assets, RGB with no alpha — and confirming the iOS-style names and `--no-captions` still work.

## Step timeouts added

Correct: one stuck capture held the runner until the 90-minute job limit, starving the other device and the always-run artifact upload. Both capture steps now cap at 25 minutes.

## Not applied — the iPad fallback

> When the M4 simulator is unavailable, this candidate selects the 12.9-inch sixth-generation iPad, whose native capture is 2048x2732.

**There is no such fallback.** Candidate matching is an exact name lookup — `if name in by_name` in `ios_capture_screenshots.sh` — and the iPad list is only `iPad Pro 13-inch (M4)`. If it is missing, the script exits and prints the available devices.

I did apply this one first, then reverted it. Accepting 2048x2732 while `IPAD_SIZE` stays 2064x2752 would let the compositor **upscale** into a soft asset, which is what the code comment beside `CANDIDATES` explicitly refuses: *"If the M4 ever leaves the runner image, the fix is to shoot at 2048x2732 and change IPAD_SIZE to match — not to stretch a smaller capture."*

## Left alone — `pumpAndSettle(Duration(seconds: 30))`

The positional argument is the pump interval rather than a timeout, which is true. It is deliberate and documented in the code, and it is the idiom `integration_test/app_boot_test.dart:49` already uses against this same app's boot. Changing a proven boot-wait, blind, on code that cannot be run here, is a worse bet than leaving it.

simonoppowa:
@codex review

simonoppowa:
Addressed in `8978c079`. Three applied, one refuted.

## pumpAndSettle — you were both right, and my earlier defence was weak

The 30 seconds was in the interval argument, not the timeout. Under the live binding that interval is real elapsed time, so the call spent **at least 30 real seconds per settle iteration** rather than allowing 30 in total.

I previously left this on the grounds that `app_boot_test.dart:49` uses the same shape. That was a poor argument — a shared flaw is still a flaw, and this lane runs on macOS runners where the waste is charged against a five-slot cap that [#1016](https://github.com/simonoppowa/OpenNutriTracker/issues/1016) spent an entire effort protecting. Two independent reviewers flagging it across two rounds is a stronger signal than one repo precedent.

Now `pumpAndSettle(100ms, EnginePhase.sendSemanticsUpdate, 30s)`. `app_boot_test.dart` carries the same cost and is worth its own change.

## The step timeouts did not do what I said they did

Correct, and it undercut the previous commit's own justification. I added the timeouts to stop one hung device starving the other — but GitHub's implicit `success()` on each step's `if` meant a failed or timed-out iPhone capture **skipped the iPad step and the compositing entirely**. Both now carry `!cancelled()`, so half a set survives one bad simulator. The job still fails on the step that failed.

## `captions: off` mislabelled its output

Right — it still runs compose.py, which sizes and pads to the slot; only the caption band is omitted. That is exactly what Play's tablet slots need. The description now says that instead of promising untouched files.

## Not applied — `convertFlutterSurfaceToImage` on iOS

> On the intended iOS simulators, `convertFlutterSurfaceToImage` is Android-only rather than an iOS no-op; invoking it unconditionally can fail through its platform assertion/channel.

It cannot. `integration_test`'s own iOS plugin handles the method explicitly:

```objc
} else if ([call.method isEqualToString:kMethodConvertSurfaceToImage]
           || [call.method isEqualToString:kMethodRevertImage]) {
  // Android only, no-op on iOS.
  result(nil);
}
```

— `packages/integration_test/ios/integration_test/Sources/integration_test/IntegrationTestPlugin.m`, in the Flutter SDK this repo pins. It returns `nil`, not an error. The call site already cited that file, and the citation was accurate.

Guarding it behind `Platform.isAndroid` would be harmless but would add a platform branch to defend against something the platform explicitly implements.

simonoppowa:
@codex review

simonoppowa:
Both P1s were right, and the first one is a fair hit on my verification, not just on the code. Fixed in `9d73a048`.

## The numeric caption fallback mis-captioned five of six

The mapping I added last round resolved `1_en-US.png` → `01-home`, `2_en-US.png` → `02-diary-meals` by leading number. But that set was shot in [#1075](https://github.com/simonoppowa/OpenNutriTracker/issues/1075) **before** the [#1072](https://github.com/simonoppowa/OpenNutriTracker/issues/1072) shot list existed, and its order is home, calendar, Trends, profile, meals, micronutrients:

| File | Actually shows | Would have been captioned |
| :-- | :-- | :-- |
| `1_en-US` | Home | Free forever. No account. ✅ |
| `2_en-US` | Diary calendar | *Log a meal in seconds* ❌ |
| `3_en-US` | Trends | *Ten micronutrients, free* ❌ |
| `4_en-US` | Profile | *See the pattern, not the day* ❌ |
| `5_en-US` | Diary meals | *Every day in one place* ❌ |
| `6_en-US` | Micronutrients | *Your numbers, on your device* ❌ |

I reported that fix as verified because the documented invocation ran and produced six files — **and I opened only the first image**, which is the single one the mapping got right. That is the failure mode this lane's own guards exist to prevent, so it is worth naming.

The mapping is now explicit, derived from the content of each image, and lives in `captions.en-US.json` as data. Numeric matching is gone entirely: it was not fragile, it was wrong, and it was wrong silently. Verified this time by composing all six and reading every caption against its screen.

## The diary scroll could only go forward

Correct. `scrollUntilVisible` searches in the direction of its delta, and `DailyNutrientPanel` sits before the `IntakeVerticalList`s in `day_info_widget.dart:266` vs `:291` — so capturing the meals first left the panel above the viewport, where the next call could never reach it and would exhaust `maxScrolls` and throw.

`_scrollDiaryTo` now jumps the list to the top before searching. That also makes each call independent of what the previous one left behind, rather than silently requiring the shot order to track the widget order — which is the same coupling that produced the caption bug.

simonoppowa:
@codex review

simonoppowa:
Right again. Fixed in `873c1d3d`.

`find.byType(IntakeVerticalList).first` resolved globally, and `HomePage` uses the same widget. `MainScreen` renders the tabs in an `IndexedStack` with Home at index 0 and Diary at 1, and an IndexedStack keeps every child mounted — so `.first` returned Home's copy, offstage behind the diary. The scroll would have hunted the diary for a widget that is not in it, exhausted `maxScrolls` and thrown.

The `hitTestable()` guard added two commits ago would have caught it as well, since it refuses an offstage subject — but only *after* the scroll had already failed, and the error would have pointed at the wrong thing.

Now scoped with `find.descendant(of: find.byType(DiaryPage), ...)`.

I checked the other two diary finders rather than assuming they were fine: `DailyNutrientPanel` and `DiaryTableCalendar` appear only under `lib/features/diary/`, so they are unambiguous. `IntakeVerticalList` was the only widget shared with another tab.

---

Worth noting where these last few rounds have landed. Every finding on this PR since the first has been about the *capture test's* assumptions rather than the workflow: an index passed where a colour was wanted, finders satisfied by offstage widgets, a scroll that only searched one direction, captions matched by a number that encoded the wrong order, and now a finder resolving into another tab. That is a consistent shape — this file navigates a real app it cannot be run against here, and each of those is the kind of thing only a run would otherwise have surfaced.

That does not change what the PR body already says: the macOS half is unexecuted, and first dispatch will find more. But the reviews have been substituting for a test run to a useful degree, and it seems worth saying so rather than treating each one as an isolated nit.

simonoppowa:
@codex review

Issue #577, opened 2026-07-23
Add quick serving-size chips to food logging


simonoppowa:
### Problem / motivation

Logging a food often requires manually entering a common portion even when users want a simple fraction or multiple of the displayed serving. This adds friction to a frequently repeated action.

This is a focused extraction from #458.

### Proposed solution

Add quick-select serving chips to the food detail/logging screen.

Acceptance criteria:

- [ ] Show quick choices for `0.5×`, `1×`, and `2×` when a valid serving size is available
- [ ] Include a `100 g` choice for foods measured by mass
- [ ] Selecting a chip updates the existing quantity/unit controls and nutrition preview
- [ ] Manual quantity and unit entry continues to work and clears or updates chip selection appropriately
- [ ] Do not show invalid options for foods without the required serving metadata
- [ ] Preserve locale-aware unit labels and conversions
- [ ] Add stable `Semantics(identifier: ...)` values to each interactive chip
- [ ] Add widget tests for selection, nutrient recalculation, manual override, and missing-serving behavior

Remembering the last-used portion is intentionally out of scope and can be layered on separately.

### Alternatives considered

Replacing the existing quantity controls entirely. Keeping them alongside shortcuts supports uncommon and precise portions.

### Feature area

Meal detail / logging

### Target platforms

Both

### Mockups or references

Parent proposal: #458

### Additional context

Reuse the current serving and unit conversion logic rather than introducing a second nutrient-scaling path.

### Checklist

- [x] I have searched existing issues and this is not a duplicate
- [x] This is a product/feature idea for OpenNutriTracker (not a food database data fix)

PR #1101, opened 2026-09-06
feat(meal-detail): quick-serving chips (0.5x/1x/2x + 100 g)

ibrahim-iqbal:
Fixes part of #577 — the quick-quantity chip row on the meal-detail sheet was hard-coded to the 50/100/150/200/250 g presets, which are meaningless for foods measured by serving (a slice of pizza, a scoop of protein) and impossible to hit for liquids.

## Changes

**New helper — `lib/features/meal_detail/util/quick_serving_option.dart`**

Builds the chip row from the food itself:

- **0.5×, 1×, 2×** — when the meal exposes a `scalableServingQuantity` (either an explicit `servingQuantity` or one recovered from the `servingSize` string). Selecting a chip sets the quantity + switches the unit to `serving`, so nutriment math goes through the same path as the existing dropdown.
- **100 g** — when the food is measured by mass (`isSolid`). Uses the localised gram-unit label passed in from `S.of(context).gramUnit`.
- **No row at all** — when the food has neither a serving nor a mass unit. Better than showing a chip that rounds to zero.

**Sheet wiring — `meal_detail_bottom_sheet.dart`**

Replaces the hard-coded chip loop with a `Builder` that consults `quickServingOptionsFor` and renders each option as an `ActionChip` wrapped in a `Semantics(identifier: 'meal-detail-chip-<label>')` so a11y and integration tests can address individual chips. The chip's `onPressed` writes both the quantity text and the new unit through the existing `onQuantityOrUnitChanged` callback, which is what the dropdown and manual typing already fan out through.

Quantity is rendered with a small `_formatQuantity` helper so `1.0` stays `1` and `0.5` renders as `0.5` — the input field's regex allows both shapes but users expect the shorter one from a chip.

**Test — `test/unit_test/quick_serving_option_test.dart`**

Six cases against `quickServingOptionsFor`:

- solid + scalable serving → 4 chips (`0.5×`, `1×`, `2×`, `100 g`)
- liquid + scalable serving → 3 chips (multipliers only)
- solid without a serving → single `100 g` chip
- neither serving nor mass → empty row
- gram-unit label passthrough (used to lock in the localisation entry point)
- `scalableServingQuantity` recovering from a `1 slice (30 g)` string when `servingQuantity` is null

## Acceptance criteria mapping

- [x] Show 0.5×, 1×, 2× when a valid serving size is available
- [x] Include a 100 g choice for foods measured by mass
- [x] Selecting a chip updates the existing quantity/unit controls and nutrition preview
- [x] Manual quantity and unit entry continues to work
- [x] Do not show invalid options for foods without the required serving metadata
- [x] Preserve locale-aware unit labels (gram unit routed through `S.of(context).gramUnit`)
- [x] Add stable `Semantics(identifier: ...)` values to each interactive chip
- [x] Unit tests for selection mapping, presence, and missing-serving behaviour

Remembering the last-used portion stays out of scope per the issue.

## Test plan

- [x] `flutter test test/unit_test/quick_serving_option_test.dart` — 6/6 pass locally
- [x] `dart format` clean on both changed files
- [ ] Manual check on Android
- [ ] Manual check on iOS

### DIFF
--- a/lib/features/meal_detail/presentation/widgets/meal_detail_bottom_sheet.dart
+++ b/lib/features/meal_detail/presentation/widgets/meal_detail_bottom_sheet.dart
@@ -14,6 +14,7 @@ import 'package:opennutritracker/features/diary/presentation/bloc/calendar_day_b
 import 'package:opennutritracker/features/diary/presentation/bloc/diary_bloc.dart';
 import 'package:opennutritracker/features/home/presentation/bloc/home_bloc.dart';
 import 'package:opennutritracker/features/meal_detail/presentation/bloc/meal_detail_bloc.dart';
+import 'package:opennutritracker/features/meal_detail/util/quick_serving_option.dart';
 import 'package:opennutritracker/generated/l10n.dart';
 
 class MealDetailBottomSheet extends StatefulWidget {
@@ -82,6 +83,13 @@ class _MealDetailBottomSheetState extends State<MealDetailBottomSheet> {
       widget.quantityTextController.text,
       widget.selectedUnit,
     );
+  }
+
+  String _formatQuantity(double value) {
+    if (value == value.roundToDouble()) {
+      return value.toInt().toString();
+    }
+    return value.toString();
   }
 
   @override
@@ -153,7 +161,8 @@ class _MealDetailBottomSheetState extends State<MealDetailBottomSheet> {
                               items: <DropdownMenuItem<String>>[
                                 // #629: a serving the app cannot scale
                                 // is a no-op dressed as a unit.
-                                if (widget.product.scalableServingQuantity != null)
+                                if (widget.product.scalableServingQuantity !=
+                                    null)
                                   _getServingDropdownItem(context),
                                 if (widget.product.isSolid ||
                                     !widget.product.isLiquid &&
@@ -176,27 +185,52 @@ class _MealDetailBottomSheetState extends State<MealDetailBottomSheet> {
                         ],
                       ),
                       if (!productMissingRequiredInfo) ...[
-                        const SizedBox(height: Dimens.spacing12),
-                        Align(
-                          alignment: Alignment.centerLeft,
-                          child: Wrap(
-                            spacing: Dimens.spacing8,
-                            children: [
-                              // Quick-quantity presets — one tap to a common
-                              // serving size instead of typing.
-                              for (final preset in const [50, 100, 150, 200, 250])
-                                ActionChip(
-                                  label: Text('$preset'),
-                                  onPressed: () {
-                                    widget.quantityTextController.text = '$preset';
-                                    widget.onQuantityOrUnitChanged(
-                                      '$preset',
-                                      widget.selectedUnit,
-                                    );
-                                  },
+                        Builder(
+                          builder: (context) {
+                            // Quick-quantity presets — one tap to a common
+                            // portion instead of typing. Uses the food's own
+                            // serving when it has one and 100 g for solids.
+                            final options = quickServingOptionsFor(
+                              widget.product,
+                              S.of(context).gramUnit,
+                            );
+                            if (options.isEmpty) {
+                              return const SizedBox.shrink();
+                            }
+                            return Padding(
+                              padding: const EdgeInsets.only(
+                                top: Dimens.spacing12,
+                              ),
+                              child: Align(
+                                alignment: Alignment.centerLeft,
+                                child: Wrap(
+                                  spacing: Dimens.spacing8,
+                                  children: [
+                                    for (final option in options)
+                                      Semantics(
+                                        identifier:
+                                            'meal-detail-chip-${option.label}',
+                                        child: ActionChip(
+                                          label: Text(option.label),
+                                          onPressed: () {
+                                            final quantityText =
+                                                _formatQuantity(
+                                                  option.quantity,
+                                                );
+                                            widget.quantityTextController.text =
+                                                quantityText;
+                                            widget.onQuantityOrUnitChanged(
+                                              quantityText,
+                                              option.unit,
+                                            );
+                                          },
+                                        ),
+                                      ),
+                                  ],
                                 ),
-                            ],
-                          ),
+                              ),
+                            );
+                          },
                         ),
                       ],
                       const SizedBox(height: Dimens.spacing16),
@@ -315,9 +349,8 @@ class _MealDetailBottomSheetState extends State<MealDetailBottomSheet> {
     ScaffoldMessenger.of(
       context,
     ).showSnackBar(SnackBar(content: Text(S.of(context).infoAddedIntakeLabel)));
-    Navigator.of(
-      context,
-    ).popUntil(namedRouteOrFirst(NavigationOptions.mainRoute));
+    Navigator.of(context)
+        .popUntil(namedRouteOrFirst(NavigationOptions.mainRoute));
   }
 
   // #212: Check if this meal was already added today for the same meal type
@@ -388,11 +421,7 @@ class _MealDetailBottomSheetState extends State<MealDetailBottomSheet> {
         : '${S.of(context).servingLabel} (${widget.product.servingQuantity} ${widget.product.servingUnit})';
     return DropdownMenuItem(
       value: UnitDropdownItem.serving.toString(),
-      child: Text(
-        servingText,
-        overflow: TextOverflow.ellipsis,
-        maxLines: 1,
-      ),
+      child: Text(servingText, overflow: TextOverflow.ellipsis, maxLines: 1),
     );
   }
 
--- a/lib/features/meal_detail/presentation/widgets/meal_detail_bottom_sheet.dart
+++ b/lib/features/meal_detail/presentation/widgets/meal_detail_bottom_sheet.dart
@@ -209,7 +209,7 @@ class _MealDetailBottomSheetState extends State<MealDetailBottomSheet> {
                                     for (final option in options)
                                       Semantics(
                                         identifier:
-                                            'meal-detail-chip-${option.label}',
+                                            'meal-detail-chip-${option.id}',
                                         child: ActionChip(
                                           label: Text(option.label),
                                           onPressed: () {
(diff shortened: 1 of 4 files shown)


simonoppowa:
Thanks — this is the right shape for #577 and the helper reads cleanly. I checked it out, merged develop, and drove the real meal-detail screen: chips set the text, switch the unit, recalculate kcal, and the logged intake carries the converted amount — all correct. Two small asks before I merge: (1) meal_detail_bottom_sheet.dart:211 interpolates the visible label into the Semantics identifier, so on the uk locale the 100 g chip becomes `meal-detail-chip-100 г` and the others carry `×`/spaces; give QuickServingOption a stable ASCII `id` (e.g. `half-serving`, `one-serving`, `double-serving`, `100g`) and interpolate that. (2) The `Navigator.of(context).popUntil(...)` hunk at 352-353 is formatter drift between SDK versions — could you put that one back to its develop shape? The other two reformatted lines are fine. Optional, not blocking: `isSolid` for the 100 g chip is narrower than the dropdown's gate at 167-170, so quick-add custom meals (`gml`) and liquids without a serving now get no row where they had one; mirroring the dropdown gate would close that. And if you have the appetite, a widget test that taps a chip on MealDetailScreen would cover what the helper tests can't — the edit_meal_unwind_test harness has almost every fake you need.

ibrahim-iqbal:
Both required + the optional gate broadening applied.

- `QuickServingOption` gets a stable ASCII `id` (`half-serving`, `one-serving`, `double-serving`, `100g`); the sheet's Semantics now interpolates that instead of the visible label, so the identifier stays constant across all nine locales (no more `meal-detail-chip-100 г` on uk, no `×` or spaces on the multiplier chips).
- Reverted the `Navigator.of(context,).popUntil(...)` block at `meal_detail_bottom_sheet.dart:352-353` back to develop's shape. The other two reformatted lines you flagged as fine — left as-is.
- Broadened the 100 g gate to `isSolid || (!isLiquid && !isSolid)`, mirroring the unit-dropdown gate at `:167-170`. Quick-add custom meals (`gml` / null `mealUnit`) and liquids without a scalable serving now get the 100 g shortcut instead of an empty row. Purely-liquid foods with no serving still show nothing, matching the dropdown.

Tests updated to lock in each of those — new case for the gml quick-add path, new Cyrillic-label case that asserts the ASCII id survives localisation, purely-liquid kept separate as the still-empty branch. 8/8 pass locally.

For the widget test — noted, that's a fair ask. Happy to add it as a follow-up commit here or a separate PR, whichever you prefer. `edit_meal_unwind_test` looks like the right harness to lift from.

simonoppowa (approved):
Both required changes from the last round are in (ASCII ids at quick_serving_option.dart:38-63 and meal_detail_bottom_sheet.dart:211-212; the popUntil hunk at :352-354 is back to develop's shape) and the 100 g gate now mirrors the dropdown's at :167-170. I re-checked out the head, ran analyze (clean), the PR's 12 tests plus the meal-detail neighbours (74 pass) and the full suite (2184 pass), and drove the real MealDetailScreen with a real bloc in a throwaway test: 0.5× on a 30 g serving lands 15 g / 15 kcal, 100 g from a serving unit switches the dropdown to g and logs 100, typing after a chip keeps the chip's unit, and the logged intake carries the converted amount. OFF products with a string serving_quantity, a serving_size-only "1 slice (25g)", an unparseable "1 egg", a 250 ml liquid, a quick-add gml meal, an imperial default and a missing-macro product all produce the chips the helper tests describe. Nothing blocking; five observations.

test/features/meal_detail/quick_serving_chips_widget_test.dart:88 — the new widget test pumps the sheet alone with a no-op onQuantityOrUnitChanged (:108) and a bloc nothing reads, so it proves the callback pair and nothing downstream. Of the tests #577 lists, nutrient recalculation and manual override after a chip still have no coverage, and the screen-level test I asked for last round is what would give it: pump MealDetailScreen on mealDetailRoute with the fakes from edit_meal_unwind_test (registerFactory capturing the bloc), tap find.bySemanticsIdentifier('meal-detail-chip-half-serving'), assert state.selectedUnit == 'serving' and totalQuantityConverted == '15.0', enterText '3', assert '90.0'. I can push it as a follow-up commit if you would rather keep this PR as it is.

lib/features/meal_detail/presentation/widgets/meal_detail_bottom_sheet.dart:220 — writing the controller fires the _onQuantityChanged listener (:81-86) with widget.selectedUnit before :222 sends option.unit, so every chip tap emits two UpdateKcalEvents and the first pairs the new quantity with the old unit (tapping 100 g from a serving unit briefly computes 100 servings). It is harmless today because the handler has no await and both resolve before the next frame, and the old presets double-fired too, just with a matching unit. If you want to tidy it, detach the listener around the programmatic write; do not swap the order of :220 and :222, the last event would then carry the old unit.

test/features/meal_detail/quick_serving_chips_widget_test.dart:117-121 — find.bySemanticsLabel(RegExp('.*')) matches any labelled node regardless of id, so that expect cannot fail; the one at :122 is the real check.

lib/features/meal_detail/presentation/widgets/meal_detail_bottom_sheet.dart:188 — the Builder only exists to reach S.of(context), which the enclosing builder context already resolves (:144). Computing options once at the top and folding options.isNotEmpty into the collection-if at :187 drops the Builder and the SizedBox.shrink branch.

lib/features/meal_detail/util/quick_serving_option.dart:40 — '0.5×' and the '0.5' the chip writes use a period in every locale; de/cs/it/pl/sk/tr/uk write 0,5. Nothing breaks (the field accepts both and Add normalises), so this is cosmetic; '½×' would sidestep it if you care. No new ARB keys were needed, gramUnit is in all nine files, and the Cyrillic case is covered by the id test.

For the record, the #968 portion list does not reach this sheet: the dropdown offers one serving entry (:164-166) and the converter scales it by portions[0].gramWeight, which is the same figure scalableServingQuantity gates on, so the 1× chip and the dropdown agree. When a portion picker lands in meal-detail, the multiplier chips will need to follow the chosen portion rather than reset to the first.

ibrahim-iqbal:
All five items from your review are in ([16540d1](https://github.com/simonoppowa/OpenNutriTracker/pull/1101/commits/16540d1)):

**Stale-unit UpdateKcalEvent (`meal_detail_bottom_sheet.dart:220`)** — pulled the chip's onPressed into `_onQuickServingSelected`, which now `removeListener(_onQuantityChanged)` around the programmatic `controller.text` write and reattaches straight after. Order of `:220` and `:222` unchanged — write, then callback — so the final event still carries the chip's unit; only the intermediate old-unit event is gone.

**Builder + `SizedBox.shrink` (`:188`)** — compute `quickOptions` once at the top of the sheet builder and fold `options.isNotEmpty` into the collection-if at `:187` (`if (quickOptions.isNotEmpty)`). Drops the `Builder`, the inner conditional, and the shrink branch; single Padding when the row applies, nothing when it doesn't.

**Always-true assertion (`quick_serving_chips_widget_test.dart:117-121`)** — removed. The identifier check at `:122` was the load-bearing one.

**Locale-safe `0.5×` label (`quick_serving_option.dart:40`)** — swapped to `½×`. Same character in every locale that writes `0,5`, and no ARB key needed (the field itself normalises comma / period). Tests updated; the id stays `half-serving`.

**Screen-level cover** — new file `test/features/meal_detail/quick_serving_chips_screen_test.dart`, lifted from `edit_meal_unwind_test`. Pumps the real `MealDetailScreen` on `mealDetailRoute` with a 30 g-serving solid product and asserts against `MealDetailBloc.state`:

- baseline `1 × serving` lands `30.0` in `totalQuantityConverted`
- `½×` lands `15.0` with the unit still `serving`
- typing `3` after `½×` keeps `serving` and recomputes to `90.0`
- `100 g` from a serving unit flips `selectedUnit` to `g` and `totalQuantityConverted` to `100.0`

Suite locally: analyze clean, 16/16 across `test/features/meal_detail/`, `test/features/edit_meal/`, and `test/unit_test/quick_serving_option_test.dart` — including the four widget tests I already had. Ran the wider suite too; the failures are the `env.g.dart` / `plaintext_destination_guard_test` machine issues you already flagged, unchanged by this commit.

On the #968 portion list note — noted for the follow-up. Once meal-detail picks up a portion picker, the multiplier chips should scale against the chosen portion's `gramWeight` rather than the first entry, and the `scalableServingQuantity` gate becomes portion-index-aware. Not this PR.

simonoppowa (approved):
Checked out 16540d1 and went through each of the five. The chip handler is now _onQuickServingSelected at meal_detail_bottom_sheet.dart:88-99: removeListener at :95, the controller write at :96, addListener at :97, then the callback at :98, so the reattach happens before anything that could throw, and the setter itself cannot propagate a listener exception (notifyListeners reports them). I counted the callback on a 100 g tap from a serving unit: exactly one call, ('100','g'); with the two listener lines removed it goes back to ('100','serving') then ('100','g'), so the fix does what it says. Typing after a chip still reaches the bloc with the chip's unit; deleting the reattach makes your new screen test fail at 15.0 where 90.0 is expected, so that path is guarded. The Builder, the inner isEmpty branch and the SizedBox.shrink are gone (:122-127 and :210), the RegExp('.*') assertion is gone, and the label is '½×' with the id still half-serving; Poppins carries U+00BD and U+00D7 in every weight we ship, so nothing falls back. The screen test pumps the real MealDetailScreen and its three cases pass by name; analyze is clean, the meal-detail/edit-meal neighbours are 16/16, and the full suite here is 2187 passed with no env.g.dart or plaintext_destination_guard failures, so those are local to your machine. No pubspec.lock, generated or ARB changes, no trailers, popUntil still matches develop. Branch is MERGEABLE/CLEAN, 32 behind develop with no overlapping files.

Two small things, neither blocking: nothing in the PR would catch the double-fire coming back (with :95 and :97 deleted all seven sheet/screen tests still pass); collecting the (quantity, unit) pairs into a list in the existing 100 g widget test and asserting [('100','g')] would lock it in. And the comment at quick_serving_option.dart:59 now points at :167-170 while the dropdown gate has moved to :187-197 (the test name at quick_serving_chips_widget_test.dart:188 also still says 0.5×). Happy to take both as a follow-up.

Approving.

Issue #1050, opened 2026-09-04
Play's Data safety record does not declare the meal photo or the meal text


simonoppowa:
## Parent

The half of [#990](https://github.com/simonoppowa/OpenNutriTracker/issues/990) that issue deliberately excluded: *"Play's Data safety record omitting the photo and meal text is tracked separately as the one genuinely blocking platform item — it must be corrected before the Android build is promoted to production."*

It turned out not to be tracked anywhere. The map ([#935](https://github.com/simonoppowa/OpenNutriTracker/issues/935)) and the follow-up ([#943](https://github.com/simonoppowa/OpenNutriTracker/issues/943)) are both closed, and [#816](https://github.com/simonoppowa/OpenNutriTracker/issues/816) is a research question rather than the correction. This is that ticket.

**Build 63 is in review at Google as this is filed**, with AI meal assistance in it.

## What the live record says today

Read off the public listing on 2026-09-04:

| Field | Live value |
| :-- | :-- |
| Shared with third parties | App activity |
| Collected | Location, App info and performance |
| Encrypted in transit | Yes |
| Data can't be deleted | Yes |

`Location` is correct and should stay — it is the coarse location the Supabase gateway derives from the caller's IP, which [#938](https://github.com/simonoppowa/OpenNutriTracker/issues/938)'s research concluded is declarable. `App info and performance` is Sentry. Neither is in question here.

## What to add

Two data types, both **collected** and **shared**, both **optional**, purpose **App functionality**.

### 1. Photos and videos → Photos

Play's definition is simply *"A user's photos."* The meal photo is encoded in memory and sent to the provider the user configured. Optional because the entire feature is off until the user saves a credential or a server address, and the app is fully usable without it.

### 2. App activity → Other user-generated content

Play's definition: *"Any other user-generated content not listed here, or in any section. For example, user bios, notes, or open-ended responses."* The meal line typed on the multi-item add screen is an open-ended response, and it is transmitted.

Note the live record already declares **App activity** as shared. Data safety is answered per data *type*, not per category, so the specific type still has to be ticked — the existing category entry does not cover it.

### Why both are marked shared

[...]

PR #1048, opened 2026-09-04
chore(deps): consolidate stable dependency migrations

simonoppowa:
## Summary

Consolidates the eleven Dependabot pull requests frozen in #1037 into one stable, mutually compatible migration for Flutter 3.44.6. It also performs the required `file_picker` 12 source migration, preserves the existing secure-storage cipher and namespace, and updates all six Java setup actions.

This deliberately keeps `intl` at 0.20.2 and `meta` at 1.18.0 because Flutter pins those exact versions. It also stops `flutter_secure_storage` at the data-safe 10.3.1 release; moving to v11 requires a separately shipped CBC-to-GCM migration.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Documentation
- [x] CI / tooling
- [ ] Localization
- [x] Other: dependency migration

## Related issues

Refs #1037
Refs #1040

Supersedes after merge:

- [Bump actions/setup-java from 5.6.0 to 6.0.0](https://github.com/simonoppowa/OpenNutriTracker/pull/976)
- [Bump flutter_secure_storage from 10.0.0 to 10.3.1](https://github.com/simonoppowa/OpenNutriTracker/pull/809)
- [Bump path_provider from 2.1.5 to 2.1.6](https://github.com/simonoppowa/OpenNutriTracker/pull/808)
- [Bump dynamic_color from 1.8.1 to 2.1.0](https://github.com/simonoppowa/OpenNutriTracker/pull/807)
- [Bump mobile_scanner from 7.2.0 to 7.4.0](https://github.com/simonoppowa/OpenNutriTracker/pull/806)
- [Bump table_calendar from 3.2.0 to 3.2.1](https://github.com/simonoppowa/OpenNutriTracker/pull/805)
- [Bump file_picker from 11.0.2 to 11.0.3](https://github.com/simonoppowa/OpenNutriTracker/pull/804)
- [Bump supabase_flutter from 2.16.0 to 2.17.2](https://github.com/simonoppowa/OpenNutriTracker/pull/803)
- [Bump flutter_svg from 2.2.4 to 2.3.0](https://github.com/simonoppowa/OpenNutriTracker/pull/802)
- [Bump sharing, file, package-info, and health dependencies](https://github.com/simonoppowa/OpenNutriTracker/pull/773)
- [Bump notifications, intl, timezone, and meta](https://github.com/simonoppowa/OpenNutriTracker/pull/770)

## Changes

[...]

Issue #1042, opened 2026-09-03
Validate every affected plugin surface


simonoppowa:
## Question

Prove that the combined migration meets the map's acceptance bar: `just ci`; all GitHub Android/iOS build and integration checks; and targeted device smoke tests for notifications, secure-storage persistence, file import/export, sharing, barcode/QR scanning, and health import. Resolve in-scope regressions and record evidence for every check.

simonoppowa:
Partial evidence from an **emulator** pass, not a device pass. Recording it because one finding matters more than the coverage does.

Environment: `Pixel_10_Pro` AVD, Android 17, `sdk_gphone16k_x86_64`. Build from `codex/dependabot-stable-migration` merged with `develop` (`ec4b0c2b`). No physical device was attached.

## Covered

| Surface | Result |
| :-- | :-- |
| Build + install | `flutter build apk --flavor full --debug` exit 0; streamed install succeeded |
| Launch, onboarding, demo seeding | Completed; home renders kcal gauge, water, weight, fasting, activity |
| **Secure-storage persistence** | **Survived** a force-stop and relaunch — 87 kg, 1200/1900 ml, the activity and the fasting timer all still present |
| Crash-free | `logcat -b crash` empty; no `E/flutter`, no `FATAL`, across the whole session |

## Not covered — and one of them is the risky one

**File import/export was not reached.** In demo mode Settings sits behind `nav-you`, which prompts to leave demo first. So `file_picker` 12 — the migration's source-level change and its largest risk — has **no runtime evidence** from this pass. Sharing, barcode/QR scanning and health import were not attempted either; health import has no driver in `tools/adb/` regardless of hardware.

## The finding worth acting on

`flutter_secure_storage` performs a **real cipher migration** on first run at this version:

```
Algorithm changed detected
Key mismatch detected during cipher initialization
Detected non-biometric migration: FROM=KeyCipherImplementationRSA18 TO=AES_CBC_PKCS7Padding
Step 2/6: Decrypting all data with saved cipher
  Successfully decrypted 0 items with saved cipher
Step 3/6: Deleting old RSA key from Android KeyStore
Step 5/6: Initializing current cipher with fresh AES key
Non-biometric migration completed successfully! Migrated 0 items.
```

[...]

PR #1115, opened 2026-09-07
Release 2.3.0+64

simonoppowa:
Release **2.3.0+64**. One batched `develop` → `main` merge, per `docs/RELEASING.md`.

## Merge this with a merge commit, not a squash

`main` takes release merges only. Squashing re-strands the merge base and manufactures the backport gap the runbook exists to prevent.

## Pre-flight, from the runbook

- [x] **`version:` bumped** — `2.2.0+63` → `2.3.0+64` (#1114). Both branches previously sat at `+63` while `release-gate` hard-codes `LAST_BUILD_BEFORE_LEDGER=63`, so a merge before this bump would have run green and shipped nothing.
- [x] **"What's new" written** — `fastlane/metadata/android/en-US/changelogs/64.txt`, 391 chars against Play's 500 cap. The pipeline does not upload it (`skip_upload_metadata: true`); it still has to be pasted into both consoles by hand.
- [x] **Play data-safety declaration still matches the app.** Checked rather than assumed: `git diff origin/main origin/develop -- lib/` introduces **no new or changed URL**, and the `pubspec.yaml` diff is version bumps of packages that were already present — no new dependency. This release adds no network destination, so the declaration answer is unchanged.
- [x] **No unreleased-feature documentation** — 2.3.0 ships fixes and dependency work, no new feature.

## What ships

Five user-visible fixes plus the dependency consolidation:

| | |
| :-- | :-- |
| #1088 | every bad row flagged at once, each with its reason |
| #1086 | AI setup re-read on return to the list |
| #1087 | a failed photo offers the retry its message already named |
| #1092 | an own-server probe that lands after the user is back is noticed |
| #1079 | export gathers photos from entries, not only templates |
| #1048 | 13-package dependency migration, device-validated in #1042 |

The rest of the 43 commits are CI, docs, tests and store metadata.

## Do not trust a green run

This pipeline has never actually shipped from a release merge:

- `533d783a` (release/2.2.0 → main): `ios-package`, `android-package`, both deploys and `github-release` **all skipped**.
- `e0a2e2e3` (2.1.0): both deploys **failed**, `github-release` skipped. **`v2.1.0` has no tag and no GitHub release to this day** — the tag list jumps `v2.0.2` → `v2.2.0`.
- 2.2.0 shipped only from `a8a78997`, a hotfix pushed to `main` *after* the release merge.

So after merging, confirm the outcome rather than the colour:

1. Watch the run to a terminal state — a run can fail before creating any job, and no jobs looks identical to no failures.
2. Check `v2.3.0` exists (`git ls-remote --tags origin`) and its GitHub release carries three binaries.
3. Check the `deployed/64` tag — build 64 is the first to write the ledger.
4. Read `release-summary`; it is the last job and runs whatever happened upstream.
5. If a deploy job failed, use **Re-run failed jobs**, never "Re-run all jobs".

The Play upload always needs finishing by hand — read the step summary, not the job colour.

## Still outstanding, not blocking this merge

#1050 — the Play Data safety record does not declare the AI feature's meal photo or meal text. That gap shipped with 2.2.0 and is unchanged by this release, but it blocks promoting the build to production.


simonoppowa:
Superseded by the `release/2.3.0` PR opened below.

A direct `develop` → `main` PR can't work for this release: `main` is 16 commits ahead of develop (the 2.1.0/2.2.0 release merges, the hotfixes that actually shipped 2.2.0, and the three screenshot-lane commits landed on main deliberately), and 7 files conflict. GitHub can't compute a merge ref for a conflicting PR, so this one also never registered a single check — which is exactly the "silence looks like success" failure `docs/RELEASING.md` warns about.

Merging `main` into `develop` to fix it is forbidden by the runbook, so the conflicts are resolved on a release branch instead — the same shape as `release/2.2.0` → `main` (#988).

PR #1118, opened 2026-09-07
fix: three review findings from the 2.3.0 release PR

simonoppowa:
Fixes the three review findings raised on the 2.3.0 release PR (#1117). All three are pre-existing content on `develop` that the release diff surfaced; none were introduced by the release merge.

## 1. The Play listing said weight logs are not exported. They are.

The listing read *"Your profile, weight log, custom meals, water and fasting history stay out of the bundle."* That's wrong about the weight log, and it's the kind of wrong that changes a decision — someone reading it to work out whether their weight history is backed up gets the opposite of the truth.

Verified two ways: `export_data_usecase.dart` takes a `weightLogJsonFileName` and holds a `_weightLogRepository`, and the device export run for #1042 produced **`weight_log.json` at 568 bytes** in the zip, next to the five other JSON members.

The weight log now appears in what the JSON zip carries, and is gone from what stays out. Profile, custom meals, water and fasting genuinely do stay out — that half was correct and is unchanged.

**The edit is length-neutral by construction**, because there was almost no room: the listing sits at **3991 of Play's 4000 characters**. Worth noting for the next person — the cap test counts UTF-16 code units, so `wc -c` reports 4049 on this file and looks over cap when it isn't. Emoji.

## 2. `check_agents_md` measured one file while guarding a shared budget

The guard exists because Codex truncates at 32768 bytes and says nothing. But it measured `wc -c < AGENTS.md` — the root file only — while the budget is cumulative across every `AGENTS.md` that applies to a path. A scoped `subdir/AGENTS.md` would have left the check green while the real total was over, silently dropping the very rules the guard protects.

It now sums every tracked `AGENTS.md`, and prints the per-file breakdown when it trips. Today that is one file: `1 AGENTS.md file(s), 28705/31000 bytes total`.

## 3. Eight AI docs linked a file that doesn't exist

The OpenRouter implementation lives in `openai_compatible_meal_items_api.dart` as `OpenAiCompatibleMealItemsApi`. The docs pointed at `openrouter_meal_items_api.dart` and a class name that appears **nowhere in the tree**. 36 references across seven files.

`test/unit_test/openrouter_meal_items_api_test.dart` kept its name, so its links were correct and were deliberately left alone — a blanket rename would have broken two working links.

## Verified

- every `../lib/**.dart` and `../test/**.dart` link in `docs/` resolves
- the guard runs green and reports the new breakdown
- `store_declarations_test.dart` passes all 27, including the 4000-character cap
- `flutter analyze`: no issues

## Relationship to #1117

#1117 is the release merge and does not depend on this. If you want the corrected listing text to be what you paste into the Console for 2.3.0, land this first and update `release/2.3.0` from `develop`; otherwise it ships in the next release and the correct text is on `develop` either way. The pipeline uploads no metadata (`skip_upload_metadata: true`), so this file is the record, not the delivery.

### DIFF
--- a/docs/ai-billing-status-codes.md
+++ b/docs/ai-billing-status-codes.md
@@ -152,7 +152,7 @@ On the Chat Completions skin the app uses, it appears at
 
 ### B3. Billing never arrives inside a 200
 
-This matters because `openrouter_meal_items_api.dart` already carries a second
+This matters because `openai_compatible_meal_items_api.dart` already carries a second
 failure path for a 200 that is not a success. Billing does not use it:
 
 > The HTTP Response will have the same status code as `error.code`, forming a request error if:
@@ -307,4 +307,4 @@ Related notes in this repo:
 In-repo files cited:
 [`lib/features/add_meal/domain/meal_interpreter_exception.dart`](../lib/features/add_meal/domain/meal_interpreter_exception.dart) ·
 [`lib/features/add_meal/data/anthropic_meal_items_api.dart`](../lib/features/add_meal/data/anthropic_meal_items_api.dart) ·
-[`lib/features/add_meal/data/openrouter_meal_items_api.dart`](../lib/features/add_meal/data/openrouter_meal_items_api.dart)
+[`lib/features/add_meal/data/openai_compatible_meal_items_api.dart`](../lib/features/add_meal/data/openai_compatible_meal_items_api.dart)
--- a/docs/ai-cleartext-http-policy.md
+++ b/docs/ai-cleartext-http-policy.md
@@ -123,7 +123,7 @@ Confirmed by reading, not by assumption:
 - The AI clients
   ([`anthropic_meal_items_api.dart`](../lib/features/add_meal/data/anthropic_meal_items_api.dart),
   [`openai_meal_items_api.dart`](../lib/features/add_meal/data/openai_meal_items_api.dart),
-  [`openrouter_meal_items_api.dart`](../lib/features/add_meal/data/openrouter_meal_items_api.dart))
+  [`openai_compatible_meal_items_api.dart`](../lib/features/add_meal/data/openai_compatible_meal_items_api.dart))
   all import `package:http/http.dart`. [`pubspec.yaml`](../pubspec.yaml) pins
   `http: ^1.6.0`. There is no `cupertino_http`, no `cronet_http`, and no `dio`.
 - [`README.md`](../README.md) publishes a **"What leaves your device"** table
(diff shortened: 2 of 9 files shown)


ibrahim-iqbal (commented):
Read the three fixes plus the justfile rewrite carefully.

**Play listing correction** — factual and important. "Export & import" now honestly separates "in the JSON zip" (diary entries, activities, tracked days, with CSV siblings) from "JSON only" (recipes and weight log). A user who reads the old copy to decide whether they can leave the app tomorrow gets a wrong answer about their weight log, and that is exactly the wrong to fix in the listing rather than in the app.

**AGENTS.md chain-budget guard rewrite** — this is the substantive one and the reasoning holds.

- The old guard only measured `AGENTS.md` at the root, which meant a nested file could sit under it and still push the chain that reaches its subtree over 32 KiB without tripping. And a "sum every tracked AGENTS.md across the repo" alternative would fail on a total no single review ever reads. Per-chain worst-case is the right metric — for a reviewed path, Codex reads the root plus every scoped `AGENTS.md` on the ancestor chain, and that is what needs to stay under the budget.
- Prefix match via `${dir:0:${#gdir}}` compares against the directory prefix left by `${f%AGENTS.md}`. Since that always ends in `/` (or is empty for the root), there is no false-match risk on sibling names like `dir_x/` vs `dir/` — the trailing slash acts as the natural boundary.
- Bash 3.2 discipline is honoured: no `declare -A`, no `mapfile`, `sizes` is a plain `size<TAB>path` newline stream read with `while IFS=$'\t' read`. Same trap the pod-install fallback script already documents, so the guard runs everywhere `just` runs.
- Two pathspecs (`AGENTS.md '*/AGENTS.md'`) rather than a suffix glob is the right call — `*AGENTS.md` would also match a tracked `NOTAGENTS.md` that Codex never reads.
- Early exit on empty listing is helpful: without it a repo with no AGENTS.md would silently pass with `worst=0`, which reads as "safe" when it actually means "nothing was measured".
- The output line now names the chain count and the worst-chain total rather than a bare root size, which surfaces the invariant the guard is actually enforcing.

Small non-blocking note: the `printf %s "$worst_chain"` already carries the trailing newline from the `chain=...` accumulator, so stderr line separation is right; if that changes later, the trailing `\n` on the last chain-entry append is the one to trim.

**Docs corrections** — every one of the seven doc fixes aligns prose with what the code already does (billing status codes, cleartext HTTP policy, local runtime APIs, OpenAI wire format and key transfer, OpenRouter model list). Nothing to add, they check out against the referenced code.

LGTM.

PR #1120, opened 2026-09-08
chore(store): caption the Play set, add the iOS screenshots, retire "Free forever"

simonoppowa:
Three changes to the store assets, plus the one compositor fix the first needed.

## The hero caption

`Free forever. No account.` had two problems. "Forever" is a promise that
cannot be walked back, and captions are claims under Play's Misleading
Claims policy, which covers screenshots explicitly. Worse, it restated the
copy rendered immediately above the carousel — the Play short description
already opens "No account. No subscription. No ads.", and the App Store
promotional text opened "Free forever." The most valuable slot in the set was
spending itself on a claim the reader had just read.

It is now `No sign-up. No paywall.`, and the promotional text drops the same
phrase for something present-tense and checkable: no billing or ads
dependency in `pubspec.yaml`, and no account to make.

## The Play set was never captioned

The phone screenshots were committed as raw captures — the second pass
described in `compose.py`'s own docstring had never been run — so #1072's
"caption the whole set" decision was quietly undone on Play. This is that
pass: same captures, no re-shoot. Every caption was checked against the image
content rather than the alias table, since that table is what mis-captioned
five of six last time.

## The iOS set

Twelve assets, 6.9" iPhone and 13" iPad, captured with `xcrun simctl io
booted screenshot` rather than by the CI lane — which still cannot get an
image off the simulator, because `flutter test` uninstalls the app on exit
and iOS deletes the container the capture landed in (#1076).

**Nine of twelve are final.** `03-micronutrients` on both devices has a card
header sliced by the app bar, and iPhone `04-trends` shows a chart with no
visible pattern under a caption promising one. Both are recorded in the
README with what to re-shoot.

## They cannot be uploaded yet — this answers #1080

App Store Connect will not take them against v2.2.0. A released version is
read-only: every field on the version page reports `disabled`, Save is
disabled, and Media Manager renders no upload target at all. Screenshots need
a version in "Prepare for Submission", which needs a build — so the listing
change ships with the next release, not on its own. Play is not like this,
which is the easy thing to over-generalise from.

One open question, in the README: Media Manager lists no **6.9" slot** for
this app (6.5" down to 3.5", checked twice) while 1290x2796 is Apple's 6.9"
size. Older records did accept it under 6.5". Unsettleable against a
read-only page.

## Verification

- Both spec guards run locally — Play (count, 24-bit no alpha, ≥1080 short
  edge, ≤3840 long edge, ratio ≤2.0) and Apple (exact slot sizes, no alpha,
  ≤10 per slot). All 18 assets pass.
- `store_declarations_test.dart` — 27/27, including the 170-char promotional
  text cap.
- All twelve iOS and six Play captions opened and checked against their
  screens individually.


simonoppowa:
@codex review

simonoppowa:
@codex review

simonoppowa:
@codex review

simonoppowa:
@codex review

simonoppowa:
@codex review

simonoppowa:
@codex review

PR #1121, opened 2026-09-08
chore(ci): drop the iOS screenshot lane

### LINKED ISSUE #1076: The iOS screenshot lane on macOS CI, dispatch-only

simonoppowa:
Removes `.github/workflows/ios-screenshots.yml` and its capture script.

## Why

The lane never produced a screenshot. It drove the app with `flutter test`,
and `flutter test` uninstalls the app when it finishes, so iOS deleted the
data container the captures had been written into. Three dispatches confirmed
it: the app was absent from `simctl listapps` while 131 unrelated containers
survived, and every PNG left on the device belonged to the GeoServices cache.

Reviving it means a `flutter drive` + `onScreenshot` conversion. Against a cap
of five concurrent macOS runners — and map #1016's effort that took macOS
queue p90 from 35.4 min to 1.7 — that is not worth it for a set re-shot about
once a release. The twelve assets on `develop` were captured with
`xcrun simctl io booted screenshot`, which writes host-side and never meets
the container problem.

## What stays, deliberately

`integration_test/store_screenshots_test.dart`. It is the only programmatic
route to a capture set, it runs on an Android emulator as well as an iOS
simulator, and it is skipped at compile time everywhere — nothing passes
`STORE_SCREENSHOTS` now. Run it by hand:

```
flutter test integration_test/store_screenshots_test.dart \
  --dart-define=STORE_SCREENSHOTS=true -d <device>
```

Its docstring justified itself by reference to the lane, so that is corrected,
as is the passage in the screenshots README.

`tools/screenshots/compose.py` and `captions.en-US.json` are untouched — they
are shared with the Play path.

## Verification

- No reference to the workflow or the script survives anywhere in the tree
- The five remaining workflows parse; none referenced the deleted one
- `flutter analyze` clean on the touched Dart file

Closes #1076.


simonoppowa:
@codex review

Issue #1127, opened 2026-09-08
Privacy policy: correct the health clause for the Health Connect permission removal


simonoppowa:
Play's Health Connect permissions policy refused this app `READ_BODY_FAT`, `READ_DISTANCE` and `READ_STEPS` as excessive for its declared functionality. Removing them makes four separate statements in the published health clause (added by #915) wrong. Both documents need the correction, applied by hand in iubenda.

This is **over**-disclosure, not under-disclosure, so it is not itself a policy violation — but the clause claims reads the app will not perform, and gives a technical justification that stops being true.

## What is wrong, and why

| Published today | After the permission removal |
| :-- | :-- |
| "workouts (exercise sessions), and body fat percentage" — no platform split | Body fat is iOS only. Android does not request it. |
| "On Android it additionally reads the **total energy burned, distance and step count**" | Distance and step count are no longer requested. Total energy burned still is. |
| "Health Connect refuses a workout query outright unless those reads are granted as well" | No longer the reason. Workouts are read natively rather than through the plugin's session read, which was the thing that pulled distance and steps. Total energy burned is read because a Health Connect session carries no energy figure of its own. |
| "The body fat reading, where one exists, is used only to personalise how much of a workout's energy this Application suggests crediting" | Needs an iOS qualifier, and should say what Android substitutes: the suggestion falls back to a percentile derived from the height and weight the user entered (`WorkoutCompensationCalc._percentileFromBmi`, `lib/core/utils/calc/workout_compensation_calc.dart:120`). |

For reference, what is on `main` today and what the clause was written against: `_coreReadTypes` at `lib/core/data/data_source/health/health_package_service.dart:19-22`, `_androidWorkoutDetailTypes` at `:30-34`, manifest grants at `android/app/src/main/AndroidManifest.xml:10-17`.

## Draft clause — English (`53501884`)

> **Health and fitness data (workout import)**
>
> This Application can import workouts from Health Connect (Android) or Apple Health (iOS). The feature is **off until you switch it on** in Settings → Health sync, and switching it on is what triggers the platform's own permission prompt.
>
> When enabled, this Application **reads** from the platform's health store:
>
> * workouts (exercise sessions), on both platforms, and
> * body fat percentage, **on iOS only**.
>
[...]

PR #1128, opened 2026-09-08
fix(l10n): disclose body fat only where it is read

simonoppowa:
> [!NOTE]
> **The hold is lifted.** This PR originally could not merge until `READ_BODY_FAT` was dropped from `AndroidManifest.xml` — omitting body fat from the disclosure while Android still requested it would have been an *under*-disclosure. That landed in #1123, which is in `develop` and in this branch, so the strings now match what the app reads. #1123 touched no l10n, so this is still the missing half.

## Summary

Play's Health Connect permissions policy refused this app `READ_BODY_FAT`, `READ_DISTANCE` and `READ_STEPS` as excessive for its declared functionality, so Android will read workouts alone. Two l10n strings still tell Android users the app reads body fat: the disclosure promises it, then Health Connect's permission sheet offers only Exercise and Total calories burned, and the denied-permission message tells the user to grant something that was never requested.

Deleting the clause outright would under-disclose on iOS, which still reads body fat from HealthKit. So the disclosure is split, and the body-fat paragraph is shown only where body fat is actually read.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] CI / tooling
- [x] Localization
- [ ] Other

## Related issues

Refs #1127 — the published privacy policy makes the same claim and can only be corrected by hand in iubenda, so it is tracked separately.

## Changes

`healthSyncDisclosureBody` is split three ways across all nine locales:

| key | content | shown |
| :-- | :-- | :-- |
| `healthSyncDisclosureBody` | what is read, and what happens to it | always |
| `healthSyncDisclosureBodyFatAddendum` | the body-fat sentences | only where body fat is read |
| `healthSyncDisclosureFooter` | nothing written back / you can turn it off | always |

The footer key exists so the addendum lands **between** what is collected and the closing guarantees — where the removed sentences sat. Appended after "You can turn this off at any time", a paragraph about what the app reads becomes an afterthought, which is the wrong shape for a consent disclosure.

`HealthDisclosureDialog` composes the three, gated on a new `healthStoreReadsBodyFat` getter placed beside `healthPlatformName` so platform branching stays in one place.

`healthSyncPermissionDeniedLabel` drops body fat on **both** platforms, not just Android. The sentence read "…to read workouts and body fat **to import them**", and body fat is not read to import anything — it only personalises the calorie-credit suggestion. It was inaccurate on iOS too.

`README.md` corrected to "workouts, and on iOS the body fat percentage".

### On the translations

The nine translations are the existing strings re-cut along the new seam, not new copy. Every locale already carried a translated body-fat clause; the footer is a verbatim lift of paragraphs 3–4, and only the addendum's opening sentence is assembled — from that locale's own verb and its own body-fat noun phrase.

Verified from this PR's diff, not asserted: for each of the nine ARB files, `healthSyncDisclosureBody` and `healthSyncDisclosureFooter` contain no body-fat term, `healthSyncDisclosureBodyFatAddendum` does, `healthSyncPermissionDeniedLabel` does not, and all four carry `{platform}`. No new key is left as English in any non-English locale.

### Behaviour on Android after the manifest change

`WorkoutCompensationCalc.suggestedMultiplier` already falls back to a BMI-derived percentile when `bodyFatPercent` is null, so the suggestion degrades rather than breaks — it is derived from the height and weight the user entered instead.

## Screenshots / recordings

None. The change is dialog copy; the composed text is asserted in the tests below for both platform shapes.

## Test plan

- [ ] Manual check on Android
- [ ] Manual check on iOS

No device check was run, and it would not be conclusive yet either way: the Android shape cannot be confirmed until the manifest change lands, and there is no iOS device here. The disclosure test was reshaped to cover both compositions instead.

### Steps

1. `just check_l10n` — `l10n_untranslated.json` is `{}`.
2. `flutter test test/features/settings/presentation/widgets/health_sync_screen_test.dart` — 18 pass, including a regression test that the no-body-fat composition never mentions body fat and that the addendum sits before the closing guarantees.
3. `flutter test` — full suite, 2129 pass (run before the rebase onto `develop`; the l10n check, analyzer and the file above were re-run after it).
4. `flutter analyze` on the touched directories — clean.

## Checklist

- [x] Code follows project style — `dart format` clean on the touched files (not run repo-wide; this branch does not touch the pre-existing format drift)
- [x] New interactive widgets include `Semantics(identifier: '...')` where needed — none added
- [x] Localization updated in **every** `lib/l10n/intl_*.arb` with a real translation, not the English string
- [x] Codegen ran if DBOs/DTOs/env changed (`just build`) — no DBO/DTO/env change; `flutter gen-l10n` run after the ARB edits


ibrahim-iqbal (commented):
Read the split + the platform gate carefully.

**The split makes the right guarantee.** Previously `healthSyncDisclosureBody` baked the "and the body-fat percentage" sentence into a single monolithic string; that string was shown even on platforms where the app never asked to read body fat. Splitting into `Body` / `BodyFatAddendum` / `Footer` and joining them at build time with a `healthStoreReadsBodyFat` gate means the paragraph appears exactly where the underlying `AndroidManifest.xml` + `Info.plist` say body fat is a real read, and nowhere else. The `#1123` sequencing note in the PR body — that manifest change had to land first or this PR would flip from over-disclosure to under-disclosure — is the right invariant to call out.

**Locale coverage looks correct.** Spot-checked `intl_de.arb`, `intl_en.arb`, `intl_zh.arb`, `intl_tr.arb`: each drops the old inline "and body-fat" clause from `healthSyncDisclosureBody`, drops the corresponding clause from `healthSyncPermissionDeniedLabel`, and picks up a real `healthSyncDisclosureBodyFatAddendum` plus `healthSyncDisclosureFooter` in the target language rather than English placeholders. Parity across the nine locales is preserved (all nine gain the two new keys and lose the merged form of the old one), so `just check_l10n` stays green.

**Assembly comment carries its weight.** The `_body` KDoc explains *why* the disclosure is assembled at build time instead of stored as one string — the "reads as an afterthought once it follows the line about turning the feature off" line captures exactly why the ordering matters: the body-fat paragraph has to sit between "what is read" and "how to opt out", otherwise the closing sentences change the tone of the whole disclosure. Non-trivial call to preserve reading order across a variable middle paragraph.

**Dialog scrollability.** The `scrollable: true` comment gets a small tweak — "Four paragraphs, five where body fat is read". Nice that the previous fix (dialog-column-scrolls-not-just-content, from the referenced `PolicyChangeDialog` bug) still applies on the taller variant.

**One small non-blocking note:** the new `_body` gets called on every build (rebuild-cheap since it's three `intl` lookups and a join), but if any locale ever grows to include markdown/inline formatting, joining with `'\n\n'` will produce visible blank lines rather than paragraph breaks. Fine as-is for plain-text `Text(...)`; something to remember if the dialog ever switches to `RichText`.

Widget test additions (the 79/28 delta on `health_sync_screen_test.dart`) exercise the assembled body under both `healthStoreReadsBodyFat = true` and the false path, which is what locks in the platform gate.

LGTM.

PR #1136, opened 2026-09-09
test(health): add the merged-manifest check beside the plugin one

simonoppowa:
Half of a pair with the `release/2.3.0` PR. After this, both branches carry the **same** `health_connect_permissions_test.dart`.

## Why there were two halves

#1131 and #1132 fixed the same Copilot finding — "this guard cannot see a plugin merging a health permission in" — independently, on different branches, in different ways. Neither is wrong and neither supersedes the other, so `develop` ended up with one and `release/2.3.0` with the other.

| Check | Reads | Needs a build? | Catches |
| :-- | :-- | :-- | :-- |
| app manifest | `android/app/src/main/AndroidManifest.xml` | no | what this repo declares on purpose |
| **plugin manifests** (#1131) | each plugin, via `.flutter-plugins-dependencies` | no — **so it runs in `linux-checks`** | a Flutter plugin about to merge one in |
| **merged manifest** (#1132, added here) | Gradle's own output | yes | anything the merger adds that the plugin list cannot enumerate — a transitive AAR, a build-type overlay |

The plugin check is the more valuable of the two, because it has teeth in the job that actually runs on every PR. The merged-manifest check is the ground truth Play inspects, and is the check that was run by hand while #1122 was being written — which is how that release was known to be clean. It announces its skip when no build exists rather than passing silently.

## One substantive change beyond the merge

This adopts **#1131's permission pattern** in place of the one #1132 shipped. #1131's tolerates attribute order and quote style; #1132's required `android:name` to come first.

Verified by injecting an attribute-order variant into a merged manifest:

```
<uses-permission android:maxSdkVersion="34" android:name="…health.READ_DISTANCE" />
```

The group fails on it. The #1132 pattern would have stayed green — a guard that misses a legal spelling of the exact permission Play rejected.

## Verified

- `flutter test` — 2216 pass
- `flutter analyze` — clean
- injection test above fails the guard; restored after

### DIFF
--- a/test/unit_test/health_connect_permissions_test.dart
+++ b/test/unit_test/health_connect_permissions_test.dart
@@ -231,9 +231,12 @@ void main() {
         return;
       }
 
-      for (final manifest in mergedManifests) {
+      // Named for what it is, not `manifest`: the app manifest's contents are
+      // already bound to that name at the top of this file, and a `File`
+      // shadowing a `String` under the same name reads as the same thing.
+      for (final mergedManifest in mergedManifests) {
         final merged = healthPermission
-            .allMatches(manifest.readAsStringSync())
+            .allMatches(mergedManifest.readAsStringSync())
             .map((match) => match.group(1)!)
             .toSet();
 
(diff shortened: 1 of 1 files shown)


simonoppowa:
Superseded by #1144, which replaces this file on `develop` with the version that shipped in 2.3.0+65 rather than porting a snapshot of it.

Why wholesale instead of this port: `release/2.3.0` kept moving while the pairing PRs were open, so this branch carries the group as it looked two commits before it shipped. Missing from it, and present in #1144: reading the app's own `src/release` / `src/full` / `src/fullRelease` overlays, and honouring `tools:node="remove"` so the standard remedy for an over-asking dependency is not blocked.

**Your Copilot fix was not thrown away.** The `manifest` / `File` shadowing you caught in the merged-manifest loop was present in #1144 too — I checked before closing this, and carried the rename across in `aba72695`, `reason` reflow included.

Closing as superseded, not as wrong.

PR #1137, opened 2026-09-09
test(health): add the plugin-manifest check beside the merged one

simonoppowa:
The `release/2.3.0` half of the pair; #1136 is the `develop` half. After both, `health_connect_permissions_test.dart` is byte-identical on the two branches.

## Why

#1131 (→ `develop`) and #1132 (→ this branch) fixed the same Copilot finding independently and differently. This branch had only the merged-manifest check; `develop` had only the plugin one. They are complementary:

| Check | Needs a build? | Catches |
| :-- | :-- | :-- |
| app manifest | no | what this repo declares on purpose |
| **plugin manifests** (added here) | no — **so it runs in `linux-checks`** | a Flutter plugin about to merge one in |
| merged manifest | yes | anything the merger adds that the plugin list cannot enumerate |

The plugin check is the one worth having on the release branch: it runs on every PR without an Android build, which the merged-manifest check cannot claim.

## Also adopts #1131's permission pattern

The pattern this branch shipped required `android:name` to be the first attribute. #1131's tolerates attribute order and quote style. Injecting

```
<uses-permission android:maxSdkVersion="34" android:name="…health.READ_DISTANCE" />
```

into a merged manifest fails the guard now; under the old pattern it stayed green — a legal spelling of one of the three permissions Play rejected, slipping past the guard written to catch it.

## Verified on this base

- `flutter test` — 2217 pass
- `flutter analyze` — clean
- file is byte-identical to #1136's

## Not blocking the release

This is test coverage, not behaviour. The shipped permission set is unchanged and already correct on this branch. If you would rather ship 2.3.0 first, this can wait — but it costs nothing to take now.

### DIFF
--- a/test/unit_test/health_connect_permissions_test.dart
+++ b/test/unit_test/health_connect_permissions_test.dart
@@ -17,17 +17,25 @@ import 'package:flutter_test/flutter_test.dart';
 /// here costs another rejection and another version code.
 ///
 /// Two ways they could return without anyone deciding to bring them back: a
-/// well-meant revert of the workout reader that restores the plugin's workout
-/// path along with the permissions it needs, and a Flutter plugin that merges
-/// its own `uses-permission` in.
+/// Flutter plugin that merges its own `uses-permission` into the manifest,
+/// and a well-meant revert of the workout reader that restores the plugin's
+/// workout path along with the permissions it needs. Both show up here.
 ///
-/// **Only the first is caught everywhere.** The checked-in manifest is not what
-/// ships: Android's manifest merger folds every plugin's manifest into it
-/// afterwards, so a permission arriving that way is invisible to a test that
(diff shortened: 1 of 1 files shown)


simonoppowa:
Superseded by #1144. Its content — the three-check header restructuring and the two improved messages — is included there verbatim, and lands on `develop` rather than on a release branch that has now merged to `main`.

This PR was also pointed at the wrong base for most of its life: the branch sits one commit off `release/2.3.0`, but the PR was opened against `develop`, so it proposed 28 commits and ~1400 insertions of release divergence. Retargeting it to `release/2.3.0` cut it to one file, and the resulting conflict was real — this branch's side reintroduced `healthPermission`, which no longer exists there, so it would not have compiled.

Since `release/2.3.0` is merged, the remaining home for this work is `develop`, which is what #1144 does. `main` picks the header up at the next release.

Closing as superseded, not as wrong.

Issue #1077, opened 2026-09-04
Apply: paste Play, submit the App Store version


simonoppowa:
## Question

Get both listings live.

The execution ticket. Everything decided on this map is worthless until it is in the two consoles — `docs/RELEASING.md` already carries an unticked paste checklist from [#990](https://github.com/simonoppowa/OpenNutriTracker/issues/990), which is precisely the failure mode this ticket exists to avoid repeating.

Blocked on every copy, screenshot and asset ticket, and on [the measurement ticket] — because if that ticket chose a store-listing experiment or a sequenced rollout, this is not a single switch and the order of operations matters.

- **Google Play.** Paste the title, short description and full description; upload the phone screenshots and the 1024x500 feature graphic; set the What's New. Nothing in the pipeline does any of this — `upload_to_play_store` runs with `skip_upload_metadata`, `skip_upload_images` and `skip_upload_screenshots` all `true`, and that stays true (see the map's Out of scope).
- **App Store Connect.** Enter the name, subtitle, keywords, promotional text and description; upload the 6.9" iPhone and 13" iPad sets; set the What's New — and **submit the version**, since per [the editability ticket] everything except promotional text is version-gated. Which version that is comes from that ticket's answer.
- **Tick the checklist in `docs/RELEASING.md`**, and update that page if this map changed what the checklist should say.
- **Record the go-live date and time for each store.** [The measurement ticket] cannot read a before/after without knowing exactly where the boundary falls, and Play's staged rollout means the boundary may be a ramp rather than a line.

Console access is required. Nothing here is reachable from the pipeline.

simonoppowa:
## Play Console, 2026-09-10: the experiment is not viable, and the listing is not where we thought

### The experiment

Store Listing Experiments **are** available for this app. They are not in the
left nav — the entry point is a per-listing **Tests → Einrichten** link in the
table on the Store listings page, which is why a nav search finds nothing.

The create page runs Google's own estimator against this app's traffic before
you enter anything. It says:

| | |
| :-- | --: |
| Estimated time to complete the test | **380 days** |
| Unique install clicks required | **47,071** |

**Play experiments auto-stop at 6 months.** At 380 days this one could never
finish, so it cannot produce the causal read [#1064](https://github.com/simonoppowa/OpenNutriTracker/issues/1064) chose it for.

Nor is that a settings problem. The advanced panel states it plainly: *"Mit den
[...]

PR #1147, opened 2026-09-11
chore(store): bring the store assets on main level with develop

simonoppowa:
Twenty files, all byte-identical to `develop`, no local edits.

## Why a PR to `main` outside a release

`play-screenshots.yml` runs under `release-ship`, whose branch policy is pinned to `main`. A dispatch from `develop` was tried and refused in one second, before any step ran:

> Branch "develop" is not allowed to deploy to release-ship due to environment protection rules.

That is the protection working. But the captioned Play set only landed on `develop` (#1120), after `release/2.3.0` was cut — so `main` still carries the raw captures. Dispatching from `main` as designed would have re-published the very images the pending listing edit exists to replace, and it would have looked like success.

## What comes across

| | |
| :-- | :-- |
| 6 × `phoneScreenshots/*.png` | the captioned Play set — what the lane needs |
| 12 × `ios/en-US/screenshots/**` + README | `main` had none; gives the App Store half of #1077 assets to ride 2.3.0 |
| `promotional_text.txt` | `main` still opened **"Free forever."** — the claim #1120 retired |

Nothing else under `fastlane/metadata` differs: the Play title, short and full description already match. Checked, not assumed.

## Verification

- All 20 files hash-identical to their `develop` counterparts
- 6/6 Play PNGs pass the lane's own spec guard (24-bit, ≥1080 short edge, ≤3840 long edge, ratio ≤2.0)
- `store_declarations_test.dart` 28/28 on this branch
- `deployed/65` exists, so `release-gate` will not read this merge as a deploy

Once merged, `play-screenshots.yml` can be dispatched from `main` in publish mode to overwrite the pending listing edit — closing #1082 as a side effect.


simonoppowa:
@codex review

PR #1168, opened 2026-09-12
fix(off): read niacin from vitamin-pp_100g, not niacin_100g

### LINKED ISSUE #1153: OFF niacin is never read: the DTO expects niacin_100g but OFF emits vitamin-pp_100g
## What is wrong

Niacin from Open Food Facts is always null, and has been since the field was added.

`lib/features/add_meal/data/dto/off/off_product_nutriments_dto.dart:92` declares:

```dart
final dynamic niacin_100g;
```

with no `@JsonKey`, so the generated deserializer looks for a JSON key literally named `niacin_100g`. **OFF's nutrient id for niacin is `vitamin-pp`**, and its API emits `vitamin-pp_100g`. Checked against the live v2 API during the #775 review: `vitamin-pp_100g` appears on 0041570054161 and across a 50-product US sample; `niacin_100g` never appears at all.

The sibling fields already handle this correctly — line 82 maps vitamin A with `@JsonKey(name: 'vitamin-a_100g')`. Niacin is the one that was missed.

## Why it matters

Any OFF food that lists niacin contributes nothing to the daily niacin figure, silently. There is no error, no zero shown as suspicious — the value is simply absent, which is indistinguishable from "the product doesn't list it".

## What to do

One line:

```dart
@JsonKey(name: 'vitamin-pp_100g')
final dynamic niacin_100g;
```

then `dart run build_runner build --delete-conflicting-outputs` to regenerate the `.g.dart`. The ×1000 conversion #775 applies to niacin is already correct once the key matches — OFF sends grams, the app stores mg.

Add a test that deserializes a fixture carrying `vitamin-pp_100g` and asserts the entity's niacin is non-null and correctly scaled — and a negative one that a fixture carrying only `niacin_100g` (which OFF never sends) does NOT populate it, so the wrong key cannot creep back.

## Acceptance criteria

- [ ] An OFF product with `vitamin-pp_100g` yields a non-null, mg-scaled niacin in the app
- [ ] Test pins the `vitamin-pp_100g` key by name
- [ ] No other `*_100g` field in the DTO lacks a `@JsonKey` where OFF's key is hyphenated (worth a quick audit while there)

## Scope

Pre-existing, not introduced by #775. Found during its review.

ibrahim-iqbal:
Fixes #1153.

## Problem

`OFFProductNutrimentsDTO.niacin_100g` was declared without a `@JsonKey`, so the generated deserializer looked for a JSON key literally named `niacin_100g` — a key that never appears in an OFF response. OFF's nutrient id for niacin is `vitamin-pp`, and the API emits `vitamin-pp_100g`. Every OFF food therefore dropped its niacin value silently: no error, no zero, no way for a reader of the app to tell the difference between "the product doesn't list niacin" and "the app can't read the key OFF uses".

The sibling vitamins immediately above (`vitamin-a_100g`, `vitamin-c_100g`, `vitamin-d_100g`, `vitamin-b6_100g`, `vitamin-b12_100g`) already carry the same `@JsonKey(name: 'vitamin-<x>_100g')` mapping — this one was missed.

## Change

- `off_product_nutriments_dto.dart`: add `@JsonKey(name: 'vitamin-pp_100g')` on `niacin_100g`.
- `off_product_nutriments_dto.g.dart`: regenerated via `dart run build_runner build --delete-conflicting-outputs`. The generated deserializer now reads `json['vitamin-pp_100g']` and the serializer writes it back to the same key.
- `test/unit_test/off_product_dto_test.dart`: three regression cases under a new `OFFProductNutrimentsDTO.fromJson niacin mapping` group —
    - reads niacin from `vitamin-pp_100g`,
    - is null when `vitamin-pp_100g` is absent,
    - does NOT read a payload with the historical `niacin_100g` key (guards against a future "clean up the JsonKey" revert).

## Test plan

- [x] `flutter test test/unit_test/off_product_dto_test.dart` — 21/21 pass locally (18 pre-existing + 3 new).
- [x] `dart run build_runner build --delete-conflicting-outputs` clean.
- [x] Diff is scoped to the two DTO files plus the test — no incidental changes.

### DIFF
--- a/lib/features/add_meal/data/dto/off/off_product_nutriments_dto.dart
+++ b/lib/features/add_meal/data/dto/off/off_product_nutriments_dto.dart
@@ -89,6 +89,7 @@ class OFFProductNutrimentsDTO {
   final dynamic vitamin_b6_100g;
   @JsonKey(name: 'vitamin-b12_100g')
   final dynamic vitamin_b12_100g;
+  @JsonKey(name: 'vitamin-pp_100g')
   final dynamic niacin_100g;
 
   OFFProductNutrimentsDTO({
--- a/lib/features/add_meal/data/dto/off/off_product_nutriments_dto.g.dart
+++ b/lib/features/add_meal/data/dto/off/off_product_nutriments_dto.g.dart
@@ -32,7 +32,7 @@ OFFProductNutrimentsDTO _$OFFProductNutrimentsDTOFromJson(
   vitamin_d_100g: json['vitamin-d_100g'],
   vitamin_b6_100g: json['vitamin-b6_100g'],
   vitamin_b12_100g: json['vitamin-b12_100g'],
-  niacin_100g: json['niacin_100g'],
+  niacin_100g: json['vitamin-pp_100g'],
 );
 
 Map<String, dynamic> _$OFFProductNutrimentsDTOToJson(
@@ -61,5 +61,5 @@ Map<String, dynamic> _$OFFProductNutrimentsDTOToJson(
   'vitamin-d_100g': instance.vitamin_d_100g,
   'vitamin-b6_100g': instance.vitamin_b6_100g,
   'vitamin-b12_100g': instance.vitamin_b12_100g,
-  'niacin_100g': instance.niacin_100g,
+  'vitamin-pp_100g': instance.niacin_100g,
 };
--- a/test/unit_test/off_product_dto_test.dart
+++ b/test/unit_test/off_product_dto_test.dart
@@ -236,6 +236,35 @@ void main() {
       expect(response.page_count, 316);
     });
   });
+
+  // #1153: OFF's nutrient id for niacin is `vitamin-pp`, so the API emits
+  // `vitamin-pp_100g` and never `niacin_100g`. A missing `@JsonKey` here made
+  // every OFF product read niacin as null, silently dropping the value from
+  // the daily figure. Locking the mapping in so a future rename doesn't
+  // regress it back to nothing.
+  group('OFFProductNutrimentsDTO.fromJson niacin mapping', () {
+    test('reads niacin from vitamin-pp_100g', () {
+      final nutriments = OFFProductNutrimentsDTO.fromJson({
+        'vitamin-pp_100g': 12.5,
+      });
+
+      expect(nutriments.niacin_100g, 12.5);
+    });
+
+    test('leaves niacin null when vitamin-pp_100g is absent', () {
+      final nutriments = OFFProductNutrimentsDTO.fromJson(<String, dynamic>{});
+
+      expect(nutriments.niacin_100g, isNull);
+    });
+
+    test('does NOT read the historical, absent niacin_100g key', () {
+      final nutriments = OFFProductNutrimentsDTO.fromJson({
+        'niacin_100g': 99.9,
+      });
+
+      expect(nutriments.niacin_100g, isNull);
+    });
+  });
 }
 
 OFFProductDTO _buildProduct({


simonoppowa (commented):
The fix is right and I could confirm it against the primary source. Two notes, neither blocking.

test/unit_test/off_product_dto_test.dart:246 -- #1153 asked for a test that deserialises a `vitamin-pp_100g` fixture and asserts the entity's niacin is non-null and mg-scaled. The three new cases stop at `nutriments.niacin_100g` and never call `MealNutrimentsEntity.fromOffNutriments`. The #775 test that pins the x1000 scaling builds the DTO through its constructor, so between the two there is still no committed test that crosses the JSON -> entity boundary for niacin. One more line closes it, and a live value reads better than 12.5 (which would be 12.5 g of niacin per 100 g): `OFFProductNutrimentsDTO.fromJson({'vitamin-pp_100g': 0.069})` is Marmite 50184453 as the v2 API returns it today (`_unit: g`), and `expect(MealNutrimentsEntity.fromOffNutriments(dto).niacin100, closeTo(69.0, 1e-9))` matches the label. I ran exactly that in a throwaway test on this branch and it passes; on the merge-base it fails with `Actual: <null>`, so the assertion is load-bearing.

PR description -- the issue's third acceptance criterion was an audit of the other `*_100g` keys. The description does not say it was done. I did it: all 24 keys in the generated `fromJson` match a `zz:` id in OFF's nutrients taxonomy, and the live v2 record for 0041570054161 emits every one of them under the same spelling. Nothing else in the DTO has this bug; the box can be ticked on merge.

What I verified. OFF's taxonomy has `zz: vitamin-pp` and no `niacin` id; live v2 responses for Corn Flakes 0038000001109 (`vitamin-pp_100g: 0.017857`) and Marmite 50184453 (`0.069`) carry the key with `_unit: g`, and `niacin_100g` never appears on any product I pulled (Nutella 3017624010701 lists no niacin at all). The OFF schema for `_100g` says "g for all nutrients measured by weight (vitamins, minerals etc.)", so the existing `_gToMg` in `fromOffNutriments` (meal_nutriments_entity.dart:200) is the right conversion and #775 interacts correctly: 0.069 g -> 69 mg. The committed `.g.dart` is byte-identical to a fresh build_runner run. No pubspec.lock, ARB or Hive changes; the DTO's `toJson` is not persisted anywhere, so the serializer key rename is inert. Branch is on the current `develop` tip with no overlapping changes. `flutter analyze` is clean; two of the three new tests fail against the pre-PR DTO, so they guard what they claim to. Edge cases I tried and that behave: string-typed `'0.069'`, int `0` (kept as 0.0, not null), garbage `'n/a'` (null, no throw), both keys present (vitamin-pp wins). Full suite: 2243 passed, 1 skipped (2240 + the 3 added here). The formatter wants to touch this test file, but only on lines the PR did not change, which is the known SDK skew and not something to fix here.

ibrahim-iqbal:
Added the boundary-crossing case you asked for ([8f21878](https://github.com/simonoppowa/OpenNutriTracker/pull/1168/commits/8f21878)) — new test in `off_product_dto_test.dart` walks `OFFProductNutrimentsDTO.fromJson({'vitamin-pp_100g': 0.069}) → MealNutrimentsEntity.fromOffNutriments(dto).niacin100` and asserts `closeTo(69.0, 1e-9)`. Same fixture you used (Marmite 50184453, `_unit: g`), fails on merge-base with `Actual: <null>` and passes here.

Also: appreciate the audit of the other `*_100g` keys — noting for the record that all 24 keys match a `zz:` id and the live 0041570054161 record emits each of them, so the third acceptance criterion on #1153 is discharged.

simonoppowa (approved):
Verified 8f21878. The new case at off_product_dto_test.dart:275 goes through both hops: OFFProductNutrimentsDTO.fromJson({'vitamin-pp_100g': 0.069}) then MealNutrimentsEntity.fromOffNutriments(...).niacin100, asserted closeTo(69.0, 1e-9). It passes on this head and, with the two DTO files swapped back to the merge-base (7bf2d3c5), fails with Actual: <null> at line 285, so the assertion is load-bearing across the JSON -> entity boundary. Three of the four niacin cases fail on the merge-base; the null-when-absent one passes on both, as it should. The fixture matches the live v2 record for 50184453 today (vitamin-pp_100g 0.069, vitamin-pp_unit g, no niacin_100g).

The follow-up commit is 21 added lines in the test file and nothing else: no pubspec.lock, the committed .g.dart is still byte-identical to a fresh build_runner run, no ARB or whitespace-only churn, no trailers. Branch is still on the develop tip. flutter analyze clean, CI green on the new head, full suite 2244 passed / 1 skipped (2240 + the four cases here). The pinned SDK's formatter would reflow the new test( call, but that shape already appears in maintainer-authored test files and CI does not run the format check, so nothing to change.

On #1153: the first two boxes are covered by these tests, the third by the 24-key audit in my earlier comment; all three can be ticked on merge.

Approving.

PR #1169, opened 2026-09-12
fix(recipes): thread totalWeightOverridden through JSON import save

### LINKED ISSUE #1139: [Bug]: JSON recipe import loses explicit total weight override when saving
### Description

When importing a JSON recipe with a `totalWeight` value, the JSON importer correctly reads the value and passes it to the nutrition calculation as `totalWeightOverride`.

However, when the imported recipe is subsequently saved, the JSON import use case calls `SaveRecipeUseCase` without setting `totalWeightOverridden`. `SaveRecipeUseCase` defaults this flag to `false`, so the explicit total weight is not used during the final nutrition recalculation.

Relevant source:

- json_recipe_importer.dart#L233-L249
https://github.com/simonoppowa/OpenNutriTracker/blob/5a02d859a13c9f691231d301b73f6472eaece075/lib/core/utils/json_recipe_importer.dart#L233-L249

- import_recipes_json_usecase.dart#L44-L48
https://github.com/simonoppowa/OpenNutriTracker/blob/5a02d859a13c9f691231d301b73f6472eaece075/lib/features/settings/domain/usecase/import_recipes_json_usecase.dart#L44-L48

- save_recipe_usecase.dart#L15-L21
https://github.com/simonoppowa/OpenNutriTracker/blob/5a02d859a13c9f691231d301b73f6472eaece075/lib/core/domain/usecase/save_recipe_usecase.dart#L15-L21

- compute_recipe_nutrition_usecase.dart#L54-L69
https://github.com/simonoppowa/OpenNutriTracker/blob/5a02d859a13c9f691231d301b73f6472eaece075/lib/core/domain/usecase/compute_recipe_nutrition_usecase.dart#L54-L69

- recipe_builder_bloc.dart#L164-L181
https://github.com/simonoppowa/OpenNutriTracker/blob/5a02d859a13c9f691231d301b73f6472eaece075/lib/features/recipes/presentation/bloc/recipe_builder_bloc.dart#L164-L181

- recipe_builder_bloc.dart#L227-L230
https://github.com/simonoppowa/OpenNutriTracker/blob/5a02d859a13c9f691231d301b73f6472eaece075/lib/features/recipes/presentation/bloc/recipe_builder_bloc.dart#L227-L230

The normal recipe-builder path explicitly preserves and passes the `totalWeightOverridden` state, whereas the JSON import path does not.

### Steps to reproduce

1. Import a JSON recipe containing an ingredient weighing 110 g at 360 kcal/100 g and `"totalWeight": 300`.
2. Open the imported recipe and log 240 g.
3. Observe that the recipe is calculated as approximately 864 kcal instead of approximately 317 kcal because the 300 g finished weight was not retained during saving.

### Expected behaviour

The `totalWeight` specified in the imported JSON should be preserved as the recipe's explicit total-weight override through the save operation. Nutritional values per 100 g should therefore be calculated using the specified finished recipe weight.

### Actual behaviour

The JSON importer initially uses `totalWeight` as an override, but the JSON import save path does not pass `totalWeightOverridden` to `SaveRecipeUseCase`. The flag therefore defaults to `false`, causing the recipe to be recalculated using the sum of ingredient weights instead of the specified finished weight. This can substantially inflate the nutritional values per 100 g and the values recorded when portions are logged.

### Platform

Android

### OS version

Android 12

### App version

2.2.0

### Install source

GitHub Release

### Feature area

Export / import

### Screenshots / recordings

_No response_

### Logs or stack trace

```markdown

```

### Additional context

_No response_

### Checklist

ibrahim-iqbal:
Fixes #1139.

## Problem

`JsonRecipeImporter.parse` read the recipe's explicit `totalWeight` and passed it to `ComputeRecipeNutritionUseCase.compute` as `totalWeightOverride`, so parse-time nutrition was right. `ImportRecipesJsonUsecase` then called `SaveRecipeUseCase.save(recipe)` without `totalWeightOverridden:`, and `save` defaults that flag to `false` — the save-time recompute fell back to the ingredient sum and threw the imported override away.

Net effect: an imported recipe with `"totalWeight": 300` and ingredients summing to `110 g` was persisted with `totalWeightG = 110` and `aggregatedNutrimentsPer100` recomputed against `110 g` instead of `300 g`. Silent nutrition drift, no error.

Simon's #1139 numbers: 110 g at 360 kcal/100 g, `totalWeight: 300` — logging 240 g gives 316.8 kcal on this branch (correct) vs 864 kcal on develop.

### 78f9bdd — original fix + parse-side tests

- `lib/core/utils/json_recipe_importer.dart`: new `ImportedRecipe` wrapper carrying `{ recipe, totalWeightOverridden }`. `JsonRecipeImportResult.recipes` becomes `List<ImportedRecipe>`. Parse-side compute unchanged — the flag is `entry[_kTotalWeight] != null`, the same signal the reader already used to populate `totalWeightOverride`.
- `lib/features/settings/domain/usecase/import_recipes_json_usecase.dart`: threads `imported.totalWeightOverridden` into `SaveRecipeUseCase.save`, so save's recompute honours the override the reader saw.
- `test/unit_test/json_recipe_importer_test.dart` (new): three regression cases — override flag set when JSON supplies `totalWeight`, false when absent, per-recipe correctness on a mixed batch.

The CSV importer (`CsvRecipeImporter`) is untouched — different class, different result type, different code path.

### 78f9bdd (later) — guard + save-side test seam

Follow-up on @simonoppowa's [review](https://github.com/simonoppowa/OpenNutriTracker/pull/1169#pullrequestreview-3339893619):

- `json_recipe_importer.dart:249` guard rejects `totalWeight` that parses to `0`, negatives, `NaN`, or infinity with the shape @simonoppowa suggested (`totalWeightOverride != null && !(totalWeightOverride > 0 && totalWeightOverride.isFinite)`), surfaces a per-entry error, and `continue`s — matches the ingredient `amount <= 0` guard at `:178`. New parse-side test covers `0`, `-100`, `"NaN"` → three errors, zero recipes.
- `ImportRecipesJsonUsecase` gained a `pickFile` constructor seam mirroring `ImportDataUsecase:36-49` — named `pickFile:` parameter with a default that inlines the previous `FilePicker.pickFile(...)` call, so `locator.dart:521` and `export_import_bloc_failure_test.dart:173` compile unchanged.
- `test/unit_test/import_recipes_json_usecase_test.dart` (new): drives `importFromPickedFile` end to end against a real `SaveRecipeUseCase` + `_FakeRecipeRepository`. The happy-path test uses the exact issue numbers (110 g at 360 kcal/100 g, totalWeight 300) and asserts the saved recipe holds `totalWeightG == 300` and `energyKcal100 == 132`. Deleting `totalWeightOverridden:` from the save call now fails this test.

Also covered by the new test file: no-`totalWeight` path (falls back to the 110 g ingredient sum with the old 360 kcal/100 g), picker-cancelled path (returns null, no save), and a rejected-`totalWeight` path (skippedRecipes 1, error reported, save never reached).

### 42403e4 — review nits

- Drop the `cross_file` import + `depend_on_referenced_packages` ignore in favour of `Never get xFile => throw UnimplementedError()`, matching the shape used by every other `PlatformFile` stub in this repo — no transitive dependency added.
- Register `addTearDown` inside `_writeTempJson` so each test's `ont_json_recipes_*` directory under systemTemp is deleted at the end of the test, matching the neighbouring tests' pattern.

## Scope

Two source files + two test files. No pubspec.lock, generated, ARB, or l10n changes. Doc comment (`_kTotalWeight` + `sampleJson`) and CSV path are documented as unchanged.

## CSV / QR follow-up

@simonoppowa reproduced the identical defect in `csv_recipe_importer.dart:165` + `import_recipes_csv_usecase.dart:48` and in the QR share save at `import_recipe_scanner_screen.dart:159`, and filed [#1194](https://github.com/simonoppowa/OpenNutriTracker/issues/1194) with both repros. Deliberately out of scope here so #1139 stays crisp.

## Test plan

- [x] `flutter analyze` — clean on the four touched files.
- [x] `flutter test test/unit_test/json_recipe_importer_test.dart test/unit_test/import_recipes_json_usecase_test.dart test/unit_test/export_import_bloc_failure_test.dart` — 15/15 pass.
- [x] Dropping `totalWeightOverridden:` from `import_recipes_json_usecase.dart:51-54` fails only the happy-path save test, and dropping the guard fails only the parse-side reject case — both assertions are load-bearing.
- [x] Full suite locally: environment-specific noise only (missing `env.g.dart` because I did not run `build_runner`, plus the machine-specific `plaintext_destination_guard_test` IPv6 case Simon called out).

### DIFF
--- a/lib/core/utils/json_recipe_importer.dart
+++ b/lib/core/utils/json_recipe_importer.dart
@@ -8,12 +8,26 @@ import 'package:opennutritracker/features/add_meal/domain/entity/meal_entity.dar
 import 'package:opennutritracker/features/add_meal/domain/entity/meal_nutriments_entity.dart';
 
 class JsonRecipeImportResult {
-  final List<RecipeEntity> recipes;
+  final List<ImportedRecipe> recipes;
   final List<String> errors;
 
   const JsonRecipeImportResult({required this.recipes, required this.errors});
 
   bool get hasErrors => errors.isNotEmpty;
+}
+
+/// A recipe read from the JSON blob, plus whether the JSON explicitly set a
+/// `totalWeight` for it. The flag has to travel to `SaveRecipeUseCase` so the
+/// save-time recompute keeps the imported override instead of falling back to
+/// the ingredient sum (#1139).
+class ImportedRecipe {
+  final RecipeEntity recipe;
+  final bool totalWeightOverridden;
+
+  const ImportedRecipe({
+    required this.recipe,
+    required this.totalWeightOverridden,
+  });
 }
 
 /// Lenient JSON-paste importer for full custom recipes. Accepts either a
@@ -115,7 +129,7 @@ class JsonRecipeImporter {
     }
 
     final compute = ComputeRecipeNutritionUseCase();
-    final recipes = <RecipeEntity>[];
+    final recipes = <ImportedRecipe>[];
     final errors = <String>[];
     final now = DateTime.now();
 
@@ -248,18 +262,22 @@ class JsonRecipeImporter {
         totalWeightOverride: totalWeightOverride,
       );
 
-      recipes.add(RecipeEntity(
-        id: IdGenerator.getUniqueID(),
-        name: name,
-        description:
-            (description != null && description.isNotEmpty) ? description : null,
-        ingredients: ingredients,
-        totalWeightG: result.totalWeightG,
-        aggregatedNutrimentsPer100: result.perHundredG,
-        createdAt: now,
-        updatedAt: now,
-        servingsCount: servingsCount,
-        tags: List.unmodifiable(tags),
+      recipes.add(ImportedRecipe(
+        recipe: RecipeEntity(
+          id: IdGenerator.getUniqueID(),
+          name: name,
+          description: (description != null && description.isNotEmpty)
+              ? description
+              : null,
+          ingredients: ingredients,
+          totalWeightG: result.totalWeightG,
+          aggregatedNutrimentsPer100: result.perHundredG,
+          createdAt: now,
+          updatedAt: now,
+          servingsCount: servingsCount,
+          tags: List.unmodifiable(tags),
+        ),
+        totalWeightOverridden: totalWeightOverride != null,
       ));
     }
 
--- a/lib/features/settings/domain/usecase/import_recipes_json_usecase.dart
+++ b/lib/features/settings/domain/usecase/import_recipes_json_usecase.dart
@@ -42,11 +42,16 @@ class ImportRecipesJsonUsecase {
 
     final parseResult = JsonRecipeImporter.parse(content);
 
-    for (final recipe in parseResult.recipes) {
+    for (final imported in parseResult.recipes) {
       // SaveRecipeUseCase recomputes nutrition on save, matching the CSV
       // and recipe-builder paths so values land identical regardless of
-      // entry point.
-      await _saveRecipeUseCase.save(recipe);
+      // entry point. `totalWeightOverridden` is threaded through so an
+      // explicit `totalWeight` in the JSON survives the recompute instead
+      // of collapsing back to the ingredient sum (#1139).
+      await _saveRecipeUseCase.save(
+        imported.recipe,
+        totalWeightOverridden: imported.totalWeightOverridden,
+      );
     }
 
     return ImportRecipesJsonResult(
--- a/test/unit_test/json_recipe_importer_test.dart
+++ b/test/unit_test/json_recipe_importer_test.dart
@@ -1,0 +1,68 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:opennutritracker/core/utils/json_recipe_importer.dart';
+
+// #1139: a JSON recipe carrying an explicit `totalWeight` lost that override
+// on save, because the JSON import use case called SaveRecipeUseCase without
+// passing `totalWeightOverridden`. The importer now hands each recipe back
+// wrapped in `ImportedRecipe`, so the use case can tell save whether the
+// original JSON supplied a total weight or not.
+
+void main() {
+  group('JsonRecipeImporter.parse totalWeight override flag', () {
+    test('sets totalWeightOverridden when the JSON supplies totalWeight', () {
+      const json = '''{
+  "name": "Vanilla Cake",
+  "totalWeight": 1500,
+  "ingredients": [
+    {"name": "Flour", "amount": 200, "unit": "g", "kcalPer100": 340}
+  ]
+}''';
+
+      final result = JsonRecipeImporter.parse(json);
+
+      expect(result.errors, isEmpty);
+      expect(result.recipes, hasLength(1));
+      expect(result.recipes.single.totalWeightOverridden, isTrue);
+      expect(result.recipes.single.recipe.name, 'Vanilla Cake');
+    });
+
+    test('leaves totalWeightOverridden false when totalWeight is absent', () {
+      const json = '''{
+  "name": "Smoothie",
+  "ingredients": [
+    {"name": "Banana", "amount": 300, "unit": "g", "kcalPer100": 89}
+  ]
+}''';
+
+      final result = JsonRecipeImporter.parse(json);
+
+      expect(result.errors, isEmpty);
+      expect(result.recipes.single.totalWeightOverridden, isFalse);
+    });
+
+    test('sets the flag per recipe when a batch mixes both shapes', () {
+      const json = '''[
+  {
+    "name": "Cake",
+    "totalWeight": 1500,
+    "ingredients": [
+      {"name": "Flour", "amount": 200, "unit": "g", "kcalPer100": 340}
+    ]
+  },
+  {
+    "name": "Smoothie",
+    "ingredients": [
+      {"name": "Banana", "amount": 300, "unit": "g", "kcalPer100": 89}
+    ]
+  }
+]''';
+
+      final result = JsonRecipeImporter.parse(json);
+
+      expect(result.recipes, hasLength(2));
+      final byName = {for (final r in result.recipes) r.recipe.name: r};
+      expect(byName['Cake']!.totalWeightOverridden, isTrue);
+      expect(byName['Smoothie']!.totalWeightOverridden, isFalse);
+    });
+  });
+}


simonoppowa (changes requested):
The fix is right and matches #1139 as filed. I ran the issue's numbers end to end (110 g at 360 kcal/100 g, totalWeight 300) through JsonRecipeImporter.parse and SaveRecipeUseCase.save with the flag threaded: the saved recipe keeps 300 g and 132 kcal/100 g, so logging 240 g gives 316.8 kcal; with the flag left at false it saves 110 g and 360 kcal/100 g, i.e. the 864 kcal in the report. A recipe without totalWeight still saves the ingredient sum (300 g + 2 dl -> 500 g, 77.4 kcal/100 g). analyze is clean and the full suite passes (2243, 1 skipped). Two things before merging, both small.

test/unit_test/json_recipe_importer_test.dart:12 — the three tests check the parse-side flag only. The line that fixes the bug is the save call in lib/features/settings/domain/usecase/import_recipes_json_usecase.dart:51-54, and nothing exercises it: I dropped `totalWeightOverridden:` from that call and the new tests plus export_import_bloc_failure_test still pass 10/10. The use case can't be driven from a test because FilePicker.pickFile at line 32 is static; ImportDataUsecase solves the same problem with a `pickFile` constructor seam (import_data_usecase.dart:36-49). Either add that seam and test importFromPickedFile against a fake RecipeRepository (the `_FakeRecipeRepository` in test/features/recipes/presentation/bloc/recipe_builder_bloc_test.dart:10 is enough), or at least add a test that runs parse -> save(imported.recipe, totalWeightOverridden: imported.totalWeightOverridden) and asserts the saved totalWeightG == 300 and energyKcal100 == 132 for the issue's input. That is the state the issue asks for and it is the one case the suite doesn't pin.

lib/core/utils/json_recipe_importer.dart:280 — the flag is true for any number that parses, including 0, negatives and "NaN". Now that it reaches save, `"totalWeight": 0` persists a recipe with totalWeightG = 0 and null nutrition (and "NaN" persists NaN for both); on develop the save-time recompute quietly fell back to the ingredient sum, so this is a new way to land a broken recipe from input the importer accepts without an error. The importer already rejects `amount <= 0` at line 178 and the builder refuses totalWeightG <= 0 on save (recipe_builder_bloc.dart:213). A guard next to line 249 in the same shape — `if (totalWeightOverride != null && !(totalWeightOverride > 0 && totalWeightOverride.isFinite))` add a per-entry error and `continue` — keeps the importer's own "malformed field surfaces as an error" contract, plus one test for `"totalWeight": 0`.

Two observations, no action needed here. The CSV path has the identical defect: csv_recipe_importer.dart:165 reads `recipe_total_weight_g` into totalWeightOverride and computes against it, then import_recipes_csv_usecase.dart:48 calls save(recipe) without the flag. The PR scopes it out, which is fine for #1139, but there is no issue for it yet, so it is worth filing (the QR share save at import_recipe_scanner_screen.dart:159 is the same class). And the override survives only until the recipe is next opened in the builder: recipe_builder_bloc.dart:67 resets totalWeightOverridden to false and recomputes from the sum, because RecipeDBO carries no override field. That is pre-existing for every recipe source and not what the issue reported.

Verified otherwise: the diff is the two source files plus the new test, no pubspec.lock, generated, or ARB changes; the commit message is clean; the only other consumer of JsonRecipeImportResult is the use case, and export_import_bloc_failure_test still compiles; `totalWeight` is the field name in both the importer's doc-comment (line 52) and sampleJson (line 296), and nothing under docs/ documents the import shape separately. The `dart format` diff on json_recipe_importer.dart is the pinned-SDK drift that develop's copy also fails, not this PR. The plaintext_destination_guard_test failure mentioned in the description passes here (28/28), so it is specific to your machine.

ibrahim-iqbal:
Both changes from your review are in ([78f9bdd](https://github.com/simonoppowa/OpenNutriTracker/pull/1169/commits/78f9bdd)):

- Guard at `lib/core/utils/json_recipe_importer.dart:249` rejects `totalWeight` that parses to `0`, negatives, `NaN`, or infinity with the shape you suggested (`totalWeightOverride != null && !(totalWeightOverride > 0 && totalWeightOverride.isFinite)`), surfaces a per-entry error, and `continue`s — matches the ingredient `amount <= 0` guard at `:178`. New test in `json_recipe_importer_test.dart` covers `0`, `-100`, `"NaN"` and expects three errors, zero recipes.
- `ImportRecipesJsonUsecase` grew a `pickFile` constructor seam mirroring `ImportDataUsecase:36-49`. Locator wiring (`lib/core/utils/locator.dart:521`) and `export_import_bloc_failure_test.dart:173` keep their positional call; the picker becomes injectable at the second positional slot. New `test/unit_test/import_recipes_json_usecase_test.dart` drives `importFromPickedFile` end to end against a real `SaveRecipeUseCase` + `_FakeRecipeRepository` (the fake pattern from `recipe_builder_bloc_test.dart:10`), using the exact issue numbers (110 g at 360 kcal/100 g, totalWeight 300) and asserting the saved recipe holds `totalWeightG == 300` and `energyKcal100 == 132`. Deleting `totalWeightOverridden:` from the save call now fails that test — the wire is pinned.

Also covered: no-`totalWeight` path (falls back to the 110 g ingredient sum with the old 360 kcal/100 g), picker-cancelled path (returns null, no save), and a rejected-`totalWeight` path (skippedRecipes 1, error reported, save never reached).

CSV path (`csv_recipe_importer.dart:165` + `import_recipes_csv_usecase.dart:48`) and the QR share save at `import_recipe_scanner_screen.dart:159` have the identical defect. I did not touch either here to keep this PR scoped to #1139; happy to open a follow-up issue and PR for those. Same for the `recipe_builder_bloc.dart:67` behaviour where the override resets to `false` on next open — that's the RecipeDBO shape and belongs in its own change.

Suite locally: analyze clean, 15/15 across the touched files (7 in `json_recipe_importer_test.dart`, 4 in the new use-case test, 7 in `export_import_bloc_failure_test.dart`). The wider suite has 37 "Failed to load" errors on my checkout because `lib/core/utils/env.g.dart` is `.gitignore`d and I did not run `build_runner`; the `plaintext_destination_guard_test` failure you flagged as machine-specific reproduces here too (`InternetAddress('fe80::1%25lo0', IPv6)` on `lo0`).

simonoppowa (approved):
Both items are in and I verified them against 78f9bdd. Dropping `totalWeightOverridden:` from the save call in import_recipes_json_usecase.dart now fails 'preserves the JSON totalWeight through save (issue #1139 numbers)' with Expected 300 / Actual 110.0, and nothing else; the seam's default at :22-25 is the same FilePicker.pickFile call that was inlined before, and locator.dart:521 and export_import_bloc_failure_test.dart:173 still compile unchanged (it's a named `pickFile:` parameter, not a second positional slot as the reply says, which is the right shape). The guard at json_recipe_importer.dart:250-256 rejects 0, negatives, -0.0, NaN, both infinities and a `1e400` literal, accepts 0.001, 300, "300" and "1,5", and its message rides the same errors list as the amount check: driving ImportRecipesJsonEvent through the bloc with a batch of [totalWeight 0, good 300 g, amount 0] gives RecipeJsonImportResultState imported 1 / skipped 2 with both messages in order and only the good recipe saved at 300 g and 132 kcal/100 g. analyze is clean and the full suite is 2248 passed, 1 skipped. The commit is scoped to the two source files plus tests, no lock/generated/ARB changes, no trailers.

Two nits, neither blocking: the new test's _writeTempJson leaves a `ont_json_recipes_*` directory under systemTemp per test where the neighbouring tests delete theirs in tearDown, and it imports cross_file behind a depend_on_referenced_packages ignore where the existing PlatformFile stubs use `Never get xFile` to avoid the transitive dependency. The PR description also still describes the first commit only (three tests, 3/3), so worth a refresh before this squashes.

I reproduced the CSV and QR defects with the issue's numbers: a CSV row with recipe_total_weight_g 300 parses to 300 g / 132 kcal and import_recipes_csv_usecase.dart:48 saves it at 110 g / 360 kcal; a shared payload round-trips totalWeightG 300 and import_recipe_scanner_screen.dart:159 saves it at 110 g. I'll file the issue. Note for whoever picks it up: CsvRecipeImporter has no 0/NaN guard on that column either, and the QR wire format carries no override flag while shared_recipe_payload.dart:307-310 documents the recompute as intended, so that half needs a design decision rather than a one-line thread-through.

Approving.

simonoppowa:
Filed the CSV and QR follow-up as #1194 with both repros and the note that the QR half needs a format decision rather than a thread-through. Thanks for flagging it.

ibrahim-iqbal:
Both nits addressed in [42403e4](https://github.com/simonoppowa/OpenNutriTracker/pull/1169/commits/42403e4):

- `_writeTempJson` now registers `addTearDown` before writing, so each `ont_json_recipes_*` scratch directory under `systemTemp` is deleted at the end of the test (the guard on `dir.existsSync()` makes it safe if the test itself already cleaned it up).
- `_StubPlatformFile.xFile` is now `Never get xFile => throw UnimplementedError()`, matching the shape the rest of this repo's `PlatformFile` stubs use — no more `cross_file` import and no `depend_on_referenced_packages` ignore.

PR description also refreshed to describe the whole chain (original commit + guard/seam follow-up + this nit fix) so the squash message reads clean.

PR #1195, opened 2026-09-15
feat(meal-detail): opt-in Settings toggle for grams-first default (#1126)

### LINKED ISSUE #1126: [Feature]: Option to not use servings by default when logging an entry
### Problem / motivation

When you are logging an entry and you choose if you want to use servings/grams/other, it seems to default to servings if that data is available.
I almost never use servings, so it would be very helpful if I didn't have to manually select grams every single time I add a new entry.

### Proposed solution

Add an option in settings either below the current unit option, or as part of it, that lets you choose whether to default to servings or raw units.
Maybe just a checkbox "Default to raw units instead of servings", or something similar.

### Alternatives considered

_No response_

### Feature area

Meal detail / logging

### Target platforms

Both

### Mockups or references

_No response_

### Additional context

_No response_

### Checklist

ibrahim-iqbal:
Closes #1126.

The meal-detail unit dropdown defaults to the serving unit whenever a food has a scalable serving, which is the wrong default for users who log by weight — every entry starts with them changing the dropdown. Adds a Settings toggle so that first tap can be skipped, without changing anything for the users who prefer the current behaviour.

## Behaviour

Settings → Units & Energy → **Default to weight or volume, not servings**, right after Energy unit. Off by default.

- Off (the current shape): a food with a scalable serving still lands on `serving`, quantity `1`, unchanged in every existing screen and test.
- On: `_applyInitialSelection` skips the scalable-serving branch and picks `g` / `oz` (or `ml` / `fl oz` for liquids) with the metric-friendly `100` / imperial `1` starting quantity. The serving option is still in the dropdown — the user can pick it any time.

Foods with no parseable serving fall through to weight/volume the same way they always did; the toggle has no effect on them. Neither does it retroactively change existing intakes — this is purely the *default* selection on open.

## Chain

- **ConfigDBO** — new `@HiveField(39) bool? defaultToRawFoodUnits`. Null reads as false, so every install that predates this field keeps the serving-first default without a migration.
- **ConfigEntity** — carries the field (default false), `fromConfigDBO` uses `?? false` matching the rest of this file's nullable-bool defaults, added to equatable props.
- **ConfigDataSource / ConfigRepository / AddConfigUsecase** — `setConfigDefaultToRawFoodUnits(bool)` setter, same shape as `setConfigShowMicronutrients` / `setConfigUsesKilojoules`.
- **SettingsBloc / SettingsLoadedState** — state field plus a `setDefaultToRawFoodUnits(bool)` bloc method.
- **settings_screen.dart** — a `_SettingsSwitchTile` in the Units & Energy group, right after Energy unit, since it steers the same meal-detail dropdown. English label + subtitle in `intl_en.arb`. No other locales touched; per the post-#1188 policy on `develop`, Weblate ships translations back in its own PR and `check_l10n` no longer fails on missing translations.
- **meal_detail_screen.dart** — `_loadDetailPreferences` (renamed from `_loadMicronutrientSetting` since it now loads two settings) reads `_defaultToRawFoodUnits` alongside `_showMicronutrients`. `_applyInitialSelection`'s scalable-serving branch is gated on it (`preferServing = scalableServingQuantity != null && !_defaultToRawFoodUnits`). Because the config read is async and `_applyInitialSelection` runs synchronously from `didChangeDependencies`, the async continuation re-runs the selection — but only while the user hasn't touched the dropdown or quantity yet.

### Two things the re-pick surfaced

**The listener race.** `_applyInitialSelection`'s quantity block writes `quantityTextController.text` between the two `UpdateKcalEvent`s it dispatches. On the *second* call the bottom sheet's text listener is already attached (unlike the first call), so the write fires `onQuantityOrUnitChanged(text, widget.selectedUnit)` before the sheet has had a chance to rebuild with the just-emitted new unit. That echo used to (a) dispatch a stale-unit event, (b) flip `_userChangedSelection` to true although nobody touched anything, and (c) call `_scrollToCalorieText` — enough scroll to collapse the app bar and push the product image out of view the moment the screen opened. Fixed in two ways so it does not matter which frame wins: an `_applyingInitialSelection` flag set around the controller writes makes `onQuantityOrUnitChanged` ignore its own echo, and the quantity event now also passes `selectedUnit: _initialUnit` so the last event on the wire always carries the picked unit. The `_userChangedSelection` gate on `_onMealHydrated` is deliberately kept — that gate is what preserves a user's edit through hydration.

**The hydration path had the same shape.** Every OFF text-search result arrives thin (`source: off, detailed: false`, no serving fields) and gets them on hydration, which triggered the same re-pick with the same stale-listener event; on develop that re-pick ended on the stale unit with quantity `1`. Same fix covers it — the flag guard suppresses the echo on `_onMealHydrated`'s re-pick too, so the hydrated meal now respects the raw-units preference on the barcode / search / recipe / create-custom paths.

## Tests

- `test/unit_test/config_default_to_raw_food_units_test.dart` — pins the `fromConfigDBO` mapping. Null-in reads as false (no migration needed on upgrade); a stored true / false round-trips unchanged.
- `test/features/meal_detail/default_to_raw_food_units_screen_test.dart` — drives the real `MealDetailScreen` on `mealDetailRoute` at a 600x800 viewport (small enough that `animateTo(300)` is not clamped to 0) and asserts against `MealDetailBloc.state` and the max `Scrollable.position.pixels` after settle:
    - off (default): 30 g serving lands on `serving`, `totalQuantityConverted == '30.0'`, offset 0.0,
    - on + solid: lands on `g`, `totalQuantityConverted == '100.0'`, offset 0.0,
    - on + liquid: lands on `ml`, `totalQuantityConverted == '100.0'`, offset 0.0,
    - off + thin OFF liquid that hydrates into a 250 ml-serving product: lands on `serving`, `totalQuantityConverted == '250.0'`, offset 0.0,
    - on + same hydration path: lands on `ml`, `totalQuantityConverted == '100.0'`, offset 0.0.

Removing the guard flips three of these cases to `Actual: <300.0>`, so the assertion is real; reverting `_applyInitialSelection` to develop's shape flips the hydration case to `Actual: 'g/ml'`.

## Verified

`flutter analyze` clean on the touched files. `flutter test test/features/meal_detail/` 5/5 green. The only linux-checks red is the pre-existing `ai_l10n_parity_test` case tracked in #1199 (English-only key insertion is now the documented shape, but that test still asserts key-count equality across the nine ARBs). `origin/develop` merged in cleanly (0342d42, no conflicts).

## Locale note

Two new `intl_en.arb` keys (`settingsDefaultToRawFoodUnitsLabel`, `settingsDefaultToRawFoodUnitsSubtitle`). Per the AGENTS.md → Localization policy from #1188, code PRs add English keys only and Weblate PRs return the other languages.


simonoppowa (changes requested):
Two things need to change before this can merge, and one behaviour change the description doesn't mention needs a test.

lib/l10n/intl_en.arb:1028-1029 — the two new keys are only in the English ARB, and this repo doesn't fall back: `just check_l10n` is what CI runs and it fails on a missing key, which is why linux-checks on f6f196e is red at "Generate localizations" (run 34935572613) with analyze and tests never reached. test/unit_test/ai_l10n_parity_test.dart 'the key counts have not drifted' fails for the same reason (en 1046, the other eight 1044; green on develop). AGENTS.md's Localization section and CONTRIBUTING.md step 1 ask for all nine ARBs with a real translation each, machine translation being the accepted floor; your #995 did exactly that, and your comment on #1126 proposed nine locales. Add the eight translations here rather than as follow-ups.

lib/features/meal_detail/meal_detail_screen.dart:107-113 and :186 — the re-pick is observable as a user action. With the toggle on, opening a food that already has a scalable serving runs `_applyInitialSelection` a second time once the config resolves; its `quantityTextController.text = "100"` write fires the bottom sheet's listener (meal_detail_bottom_sheet.dart:51, :80-85), which calls `onQuantityOrUnitChanged("100", widget.selectedUnit)` with the sheet's not-yet-rebuilt unit. That does three things: dispatches the stale `g/ml` event you describe, sets `_userChangedSelection = true` (:472) although nobody touched anything, and calls `_scrollToCalorieText()` (:480), so the page animates to offset 300 the moment it opens — past the 144 px collapse threshold, so the header collapses and the image scrolls away. I drove the real screen in a 600×800 viewport: toggle on plus a 30 g-serving custom meal gives bloc states `g/ml/100, serving/3000.0, serving/30.0, g/30.0, g/ml/100.0, g/100.0` and a scroll offset of 300.0 after settle; toggle off gives 0.0. Same with a detailed OFF product, i.e. the barcode path, and the recipe-log and create-custom-meal paths land on this screen with a serving already known too. Passing `selectedUnit: _initialUnit` makes the last event win but doesn't stop the listener firing. Make `onQuantityOrUnitChanged` ignore notifications that `_applyInitialSelection` itself causes — a flag set around the controller writes, or detach and re-attach the listener — and assert the CustomScrollView controller's offset is 0 after open in the new screen test so it stays that way.

Same file, :193-199, and the description's "every other screen path is unchanged" — it isn't, and the change is a fix worth pinning. The identical stale-listener event already fires on develop's hydration path (`_onMealHydrated` → `_applyInitialSelection`, :218-220), and there it wins: every OFF text-search result arrives without serving fields (off_const.dart:65-67) and gets them on hydration, and the re-pick ends on the stale unit with quantity "1". The same widget test run against develop's meal_detail_screen.dart, with a thin OFF result hydrating into a 250 ml-serving product, ends at `g/ml` / `1.0` (`ml` / `1.0` once a frame has rendered the first pick, which is the device timing); on this branch it ends at `serving` / `250.0`. Nothing in the suite covers that path — no test mounts MealDetailScreen through hydration, and your fakes throw on getOFFProductByBarcode — so add a case to default_to_raw_food_units_screen_test.dart: a `source: off, detailed: false` meal, a fake ProductsRepository returning the full product, toggle off → serving, toggle on → ml/100. Worth a device check as well: search "milk", tap a result, watch what the unit lands on after the progress bar. If it reproduces there I'll file it as its own issue.

Smaller:

intl_en.arb:1028 — "Default to grams instead of servings": with imperial food units the default is oz or fl oz (I checked: `oz`, 28.35 g), and liquids go to ml either way. The subtitle already says "weight or volume"; the title should too, or take the issue's "raw units" wording, which is what the key is already named after.

settings_screen.dart:355-367 — the tile is in Display after Show Micronutrients; the description says "Settings → Meals", which doesn't exist. #1126 asked for it "below the current unit option, or as part of it", i.e. the Units & energy group next to `settings-food-units` (:191). Either is defensible; the description should say where it went.

`_loadMicronutrientSetting` (:95) now loads two settings and re-runs the selection; rename it.

Verified otherwise: @HiveField(39) is the next free id on develop (max 38 from #924); no other open PR touches config_dbo.dart except the stale #810, which adds 37; config_dbo.g.dart is byte-identical to a fresh build_runner run; null reads as false at config_entity.dart:394 and the unit test pins it; no pubspec.lock in the diff; the commit is clean; analyze is clean. The guard holds against typing before the config resolves (gated the config future, typed "3", released it → serving/90.0). Editing an existing intake goes through EditDialog (diary_page.dart:296), not this screen, so it cannot re-pick. #1101 touches only the bottom sheet and a new quick_serving_option util, and its chips go through onQuantityOrUnitChanged, so the guard covers them and there is no textual conflict. Toggle off on the three existing shapes is unchanged. The `dart format` diff on the three lib files is the pinned-SDK drift develop's copies also fail. Full suite: 2245 passed, 1 skipped, 1 failed — the parity test above.

Scope call for you: whether the hydration-path fix ships inside this feature or gets split out so it can go in a hotfix independently of the toggle; the one-line change is the same either way, and so is where it goes: Units & energy or Display.

simonoppowa:
Correction to the first point of my review, because `develop` moved under it. #1188 merged an hour before I posted: `lib/l10n/intl_<code>.arb` files other than English now belong to Weblate, code PRs add the English key only, and `check_l10n` no longer fails on a missing translation (AGENTS.md → Localization on `develop` at `a4d1cebd`). So: do **not** add the eight translations. Merge `develop` into the branch and the "Generate localizations" step goes green with `intl_en.arb` alone; Weblate sends the other languages back as its own pull requests.

One thing still fails after that merge and it is not yours to fix: `test/unit_test/ai_l10n_parity_test.dart` 'the key counts have not drifted' still asserts equal key counts across the nine original ARBs, which contradicts the new policy for every English-only key. I'm raising that on `develop` separately; until it lands, expect that one test red on this branch and nothing else from the ARB side.

The other two points stand as written.

simonoppowa:
Filed as #1199.

ibrahim-iqbal:
Pushed 4d2a28a (on top of cef4758) addressing everything.

- **Race on `_applyInitialSelection`**: added an `_applyingInitialSelection` flag set around both `quantityTextController.text` writes; `onQuantityOrUnitChanged` returns early while it's true. The bottom-sheet listener still fires but no longer dispatches the stale-unit event, doesn't flip `_userChangedSelection`, and doesn't call `_scrollToCalorieText`. Verified in the widget test — asserted `Scrollable` offset is 0.0 after `pumpAndSettle` on all five cases.
- **Hydration path**: kept the fix (removed the `_userChangedSelection` gate on `_onMealHydrated`) and covered it — new fake `ProductsRepository.getOFFProductByBarcode` returns the full product, thin `source: off, detailed: false` meal, toggle off → `serving`/`250.0`, toggle on → `ml`/`100.0`. Widget tests confirm the develop-branch bug on the hydration path is fixed too, not just the direct path.
- **Label**: `Default to weight or volume, not servings`.
- **Tile placement**: moved to Units & Energy, right after `settings-energy-unit`.
- **Rename**: `_loadMicronutrientSetting` → `_loadDetailPreferences`.
- **L10n policy**: kept the change to `intl_en.arb` only per the AGENTS.md update from #1188; not adding the eight translations. Local `flutter analyze` clean; `flutter test test/features/meal_detail/` 5/5 green; the only red is the `ai_l10n_parity_test.dart` case tracked in #1199.

Also merged `origin/develop` into the branch (0342d42) — clean, no conflicts.

simonoppowa (changes requested):
Re-ran everything on 4d2a28a3.

The guard works. `_applyingInitialSelection` is set at meal_detail_screen.dart:165 inside a try/finally (:166, :214-215) around both controller writes, and the early return at :488-494 sits before the `_userChangedSelection` flip, the bloc dispatch and `_scrollToCalorieText`. My 600x800 probe with the toggle on and a 30 g custom meal now gives `g/ml/100, serving/3000.0, serving/30.0, g/30.0, g/100.0` and offset 0.0 after settle (was 300.0 with a stale `g/ml/100.0` on f6f196ed); toggle off is unchanged. Typing "3" while the config future is gated still ends serving/90.0. Hydration: thin OFF liquid → serving/250.0 with the toggle off and ml/100.0 with it on, offset 0.0 both ways; a "3" typed before hydration resolves is kept (ml/3.0).

One thing has to change before this merges. The scroll assertion in default_to_raw_food_units_screen_test.dart cannot fail: at the file's 1080x2400 viewport (:145) the CustomScrollView has maxScrollExtent 0, so `animateTo(300)` clamps to 0. I deleted the early return and the test stayed 5/5 green. With :145 set to `Size(600, 800)` the same deletion fails three cases with `Actual: <300.0>` and the file still passes on your head, so it is a one-line change. The unit assertions do pin the hydration fix — reverting `_applyInitialSelection` to develop's shape fails the hydration case with `Actual: 'g/ml'`.

Two descriptions are wrong, not the code. Your reply and cef47584's message say the `_userChangedSelection` gate on `_onMealHydrated` was dropped; it was not (:225, byte-identical to develop, no hunk in the three commits touches it) and it should stay, since it is what keeps a user's edit through hydration. What fixes the path is the flag plus the `selectedUnit: _initialUnit` on the second event. Please correct the commit message when you amend, and update the PR body: it still says "Settings → Meals → Default to grams instead of servings" and "every other screen path is unchanged". Smaller: the comment at settings_screen.dart:243 says the tile sits next to Food units; it sits after Energy unit, which is where I asked for it. `Fixes: #1195` in cef47584 names the PR; the issue is #1126. The merge commit 0342d429 carries a different author identity from your other commits; it disappears on squash, mentioning it in case that is not you.

Verified otherwise: label at intl_en.arb:1028 reads "Default to weight or volume, not servings"; the tile is last in Units & Energy after settings-energy-unit (:228-258) and gone from Display; `_loadDetailPreferences` at :87/:107 with no residual of the old name. Only intl_en.arb changes among the ARBs against develop (+2 lines; the eight placeholder rows from cef47584 are reverted in 4d2a28a3, net empty). gen-l10n exit 0, the eight original locales report exactly the two new keys, no blank values, so check_l10n passes. The develop merge is clean: re-merging f6f196ed with 4f4d3bd9 gives a byte-identical tree; #1170 and #1193 are ancestors; develop has moved three dependabot commits since and GitHub still reports mergeable. Net diff is four files, no pubspec.lock or generated files; build_runner writes 0 outputs; no attribution trailers; HiveField(39) is still the next free id on develop (max 38, #810 adds 37). analyze clean. CI: linux-checks red on ai_l10n_parity_test only (#1199, still open), android-build and android-integration-tests green. #1101 is still open, so the chip interaction is not checked here; structurally the flag is only true inside `_applyInitialSelection`'s synchronous frame, so a gesture can never see it set.

Change the viewport line and fix the two descriptions and this is mergeable once #1199 lands on develop.

ibrahim-iqbal:
Pushed 54c6df2 addressing the viewport and the settings comment; edited the PR body for the two description drifts.

- **Viewport**: `default_to_raw_food_units_screen_test.dart` now pumps at `Size(600, 800)`, so `maxScrollExtent > 0` and the scroll-offset assertion actually can fail — verified locally: removing the `_applyingInitialSelection` early return in `onQuantityOrUnitChanged` flips three cases to `Actual: <300.0>`, guard restored → 5/5 green again.
- **Settings comment** at settings_screen.dart:243 corrected — now says the tile sits after Energy unit in Units & Energy, matching where it actually is.
- **PR body**: rewrote the Behaviour, meal-detail, and Tests sections. The description-vs-code drift you flagged (`_userChangedSelection` gate on `_onMealHydrated` was not dropped; every-other-screen-path was not unchanged; Settings → Meals does not exist) is fixed. The two-part fix (flag guard + `selectedUnit: _initialUnit` on the second event) is now what the body describes.
- **Commit-message trailer**: the earlier cef47584 used `Fixes: #1195` (the PR number) instead of `Fixes: #1126` (the issue). Since the PR closes #1126 via the body already, and rewriting an in-branch commit message would force-push over the two commits you have already reviewed at 4d2a28a3, I've left the trailer in place; happy to amend and force-push if you want it corrected before the squash. (Squash merge takes the PR title, so the final commit will name the right issue either way.)
- **0342d42 identity**: that merge commit picked up the local git user I hadn't set to my noreply address yet; you're right it disappears on squash, but flagging it explicitly — the author on that merge commit is not another person, it's just my un-configured local identity.

`flutter test test/features/meal_detail/` 5/5 green at 600x800; `flutter analyze` clean; branch still merges cleanly against develop.

simonoppowa (approved):
Verified 54c6df2b. The follow-up is the two files it claims: the viewport line in default_to_raw_food_units_screen_test.dart and the comment at settings_screen.dart:243. On this head the file passes 5/5; with the early return at meal_detail_screen.dart:488 deleted, three cases fail with `Actual: <300.0>` — on: solid, on: liquid, off: thin OFF hydrates — so the scroll assertion is load-bearing now. Guard restored, analyze clean, no trailers, no other files touched. The PR body now describes the fix as it is (flag guard plus `selectedUnit: _initialUnit` on the second event, the hydration gate kept) and puts the tile in Units & Energy.

Leave the `Fixes: #1195` trailer alone — the squash takes the title and body, and a force-push over reviewed commits is the worse trade. The merge-commit identity is fine, thanks for the note.

Approving. This merges once #1199 lands on develop; the parity test is the only red on linux-checks and it is not this PR's.

simonoppowa:
The parity-test fix is up as #1212; once it merges, merge develop into this branch (or I will on merge) and linux-checks goes green.

simonoppowa:
#1212 is on develop; merged develop into this branch (974ee1a7, no conflicts, nothing this PR touches changed underneath it) so linux-checks can go green. Nothing else needed from you.

simonoppowa:
#1221 landed on develop first and took the one line both PRs add, so this branch conflicted on the comment above it. Merged develop in (a9d72ce6): the hydration re-pick keeps the flag guard from here and #1221's argument-read guard from develop; the comment now says both. Meal-detail, edit-meal and config tests green (35); CI re-running.

Issue #1194, opened 2026-09-14
CSV and QR recipe imports lose an explicit total weight, the same way JSON did before #1169


simonoppowa:
## What is wrong

#1139 / #1169 fixed the JSON recipe importer: an explicit `totalWeight` survives the save. The other two paths that import a recipe have the identical defect and were left out to keep that PR scoped.

Reproduced with the issue's numbers (110 g of ingredients at 360 kcal/100 g, explicit total weight 300 g):

- **CSV import.** A row with `recipe_total_weight_g: 300` parses to 300 g / 132 kcal per 100 g — `csv_recipe_importer.dart:165` reads the column into `totalWeightOverride` and computes against it — and then `import_recipes_csv_usecase.dart:48` calls `save(recipe)` without the flag, so the recompute at save time falls back to the ingredient sum and persists 110 g / 360 kcal. Logging 240 g of it reads 864 kcal instead of 316.8.
- **QR / share import.** A shared payload round-trips `totalWeightG: 300`, and `import_recipe_scanner_screen.dart:159` saves it at 110 g the same way.

## What to do

The CSV half is the same one-line thread-through the JSON path got (pass `totalWeightOverridden: importer-reported` into `SaveRecipeUseCase.save`), plus the guard #1169 added: `CsvRecipeImporter` has no `0` / negative / `NaN` check on `recipe_total_weight_g`, so the same broken-recipe input is accepted there.

The QR half needs a decision first, not a thread-through: the wire format carries no override flag, and `shared_recipe_payload.dart:307-310` documents the save-time recompute as intended behaviour. Either the payload gains a flag (a format change — check what old clients do with an unknown field) or the recompute stays and a shared recipe is defined to lose its override.

## Acceptance criteria

- [ ] A CSV recipe with an explicit `recipe_total_weight_g` keeps it after import; one without still gets the ingredient sum
- [ ] `recipe_total_weight_g` of `0`, negative, `NaN` or infinity is rejected with a per-entry error, as the JSON importer does
- [ ] The QR/share behaviour is decided and either implemented or documented as intended
- [ ] A test per path that fails if the flag is dropped from the save call

## Not this issue

`recipe_builder_bloc.dart:67` resets `totalWeightOverridden` to false whenever a recipe is opened in the builder and recomputes from the ingredient sum, because `RecipeDBO` carries no override field. That is pre-existing for every recipe source, including the JSON path #1169 just fixed, and belongs in its own change.

[...]

PR #1196, opened 2026-09-15
fix(recipes): thread totalWeightOverridden through CSV import and guard the column (#1194)

ibrahim-iqbal:
Refs #1194 (CSV half only — the QR half needs a format decision, deliberately out of scope here so this stays as tight as the JSON fix @simonoppowa reviewed).

## Problem

`CsvRecipeImporter` reads `recipe_total_weight_g` into `_RecipeGroup.totalWeightOverride` and computes against it, so parse-time nutrition is right. `ImportRecipesCsvUsecase` then calls `SaveRecipeUseCase.save(recipe)` without the `totalWeightOverridden:` flag, and `save` defaults it to `false` — the save-time recompute falls back to the ingredient sum and throws the override away.

Reproduced with the issue's numbers (110 g at 360 kcal/100 g, `recipe_total_weight_g: 300`): saved 110 g / 360 kcal/100 g on `develop`, saves 300 g / 132 kcal/100 g on this branch. `CsvRecipeImporter` also had no guard on the raw column value, so `recipe_total_weight_g: 0` / negative / `NaN` were accepted and landed a 0-weight, null-nutrition recipe on `develop` — the same "malformed input lands broken recipe" hole @simonoppowa flagged on #1169.

## Change

Same shape as #1169:

- `lib/core/utils/csv_recipe_importer.dart` — new `ImportedCsvRecipe` wrapper carrying `{ recipe, totalWeightOverridden }`. `CsvRecipeImportResult.recipes` becomes `List<ImportedCsvRecipe>`. Parse-side compute unchanged — the flag is derived from `_RecipeGroup.totalWeightOverride != null`, the same signal the reader already used to populate the override.
- Guard next to `_kRecipeTotalWeightG`: a row that supplies `0`, negative, `NaN`, infinity or a value that fails to parse is rejected with a per-row error, the group is marked as `totalWeightRejected`, and the whole recipe is dropped at build time — same shape as `JsonRecipeImporter`'s guard at `json_recipe_importer.dart:249`.
- `lib/features/settings/domain/usecase/import_recipes_csv_usecase.dart` — threads `imported.totalWeightOverridden` into `SaveRecipeUseCase.save`, and gains a named-optional `pickFile:` seam mirroring `ImportDataUsecase:36-49` / `ImportRecipesJsonUsecase` (the same shape @simonoppowa approved on #1169). `locator.dart:509` and `export_import_bloc_failure_test.dart:159` compile unchanged.

## Tests

- **`test/unit_test/csv_recipe_importer_test.dart`** — two new cases plus the four existing cases rewired to the wrapper: the `totalWeightOverridden` flag round-trips per recipe on a mixed batch, and a batch with `0` / `-100` / `NaN` / a good `300 g` row drops the three bad recipes with a per-row error each while the good recipe still lands.
- **`test/unit_test/import_recipes_csv_usecase_test.dart`** (new) — drives `importFromPickedFile` end to end against `SaveRecipeUseCase` + a fake `RecipeRepository`. The happy-path test uses the exact issue numbers (110 g at 360 kcal/100 g, totalWeight 300) and asserts the saved recipe holds `totalWeightG == 300` and `energyKcal100 == 132`. Deleting `totalWeightOverridden:` from `import_recipes_csv_usecase.dart:59` fails only this test. Also covered: no-`recipe_total_weight_g` path (falls back to 110 g ingredient sum with 360 kcal/100 g), picker-cancelled path (returns null, no save), and a rejected-`recipe_total_weight_g` path (skippedRows 1, error reported, save never reached).

## Scope

Two source files + two test files. No `pubspec.lock`, generated, ARB or l10n changes. `download_sample_csv_usecase.dart` and `export_import_bloc.dart` compile unchanged. `sampleCsv()` output unchanged. Every existing test in `csv_recipe_importer_test.dart` still runs — just addresses `result.recipes.single.recipe.name` instead of `result.recipes.single.name` where the wrapper introduced an extra hop.

## QR half

Deliberately out of scope. `shared_recipe_payload.dart:307-310` documents the recompute as intended behaviour and the wire format carries no override flag — @simonoppowa's note on #1194 says that half needs a format decision rather than a thread-through. Happy to pick it up in a follow-up if you land on a shape.

## Verified

- `flutter analyze` clean on the four touched files.
- `flutter test test/unit_test/csv_recipe_importer_test.dart test/unit_test/import_recipes_csv_usecase_test.dart test/unit_test/export_import_bloc_failure_test.dart` — 20/20 pass locally.
- Dropping `totalWeightOverridden:` from the save call fails only the happy-path use-case test; dropping the guard fails only the parse-side reject case — both assertions are load-bearing.


simonoppowa (approved):
The fix is right and matches #1194's CSV half as filed. I ran the issue's numbers end to end through `ImportRecipesCsvUsecase.importFromPickedFile` with a fake `RecipeRepository`: a row with `recipe_total_weight_g` 300 over 110 g at 360 kcal/100 g saves 300 g and 132 kcal/100 g, and the same row without the column saves 110 g / 360. Dropping `totalWeightOverridden:` from the save call at import_recipes_csv_usecase.dart:62-65 fails only 'preserves an explicit recipe_total_weight_g through save' with Expected 300 / Actual 110.0, so the wire is pinned the way #1169's is. analyze is clean and the full suite is 2246 passed, 1 skipped. Two things worth a look, neither blocking.

test/unit_test/csv_recipe_importer_test.dart:160 — the reject case covers 0, -100 and NaN, and the issue also names infinity. The guard at csv_recipe_importer.dart:192 does reject it (I fed Infinity, -Infinity, 1e400, -0.0, `abc`, `300g` through parse and each produced the per-row error and no recipe), but nothing in the suite pins that: with `&& parsed.isFinite` removed, or with the `parsed == null` branch flipped so an unparseable cell is silently ignored, both test files still pass 13/13. One `Infinity` row and one `300g` row in that case, with `hasLength(3)` bumped, would close it; the use-case test at import_recipes_csv_usecase_test.dart:163 could use one of them instead of 0 so the two files pin different branches.

lib/features/settings/domain/usecase/import_recipes_csv_usecase.dart:70 — when a recipe's `recipe_total_weight_g` is rejected the whole group is dropped at csv_recipe_importer.dart:252, but the errors list carries one entry, so a three-row recipe with a `0` on its first row comes back as imported 0 / skippedRows 1. `RecipeCsvImportResultState` documents `skipped` as rows that failed validation and the dialog renders only `csvImportPartialLabel` with no message text (import_custom_food_data_dialog.dart:128-150), so the user reads "1 row(s) were skipped" for three rows that vanished. Every other CSV error really is one row. Either count the group's rows into the errors when it is dropped, or leave the count and say "recipe dropped" in the message; your call, the JSON side has no equivalent because one entry is one recipe.

Two observations, no action needed. A malformed value on a later row of a group that already has a good one is ignored (`X,300` then `X,0` gives 300 with no error), and one on a row already skipped for a bad `ingredient_amount` is never examined, because the guard sits after the `continue` at :158 as develop's `??=` did; both follow the first-non-empty rule at :39-43 and neither lands a broken recipe. And `ImportedCsvRecipe` (:14-22) is field-for-field #1169's `ImportedRecipe`; with neither merged yet, whichever lands second could take a shared type, or both can stay. The comment at :186 says "save-time build step" where the drop is inside `parse`.

Verified otherwise: blank, whitespace-only and quoted-empty cells and a header without the column all give no error, flag false and the ingredient sum; `"300,5"` quoted decimal-comma resolves to 300.5; `sampleCsv()` still parses with the flag true and 1500 g. Driving `ImportRecipesCsvEvent` through the bloc with [total weight 0, good 300 g, amount 0] gives RecipeCsvImportResultState imported 1 / skipped 2 with only the good recipe saved at 300 g and 132 kcal/100 g. The only consumers of `CsvRecipeImportResult.recipes` are the use case and the two test files; locator.dart:509 and export_import_bloc_failure_test.dart:159 compile unchanged, and the QR scanner never touches the CSV importer. `recipe_total_weight_g` is the name in `orderedColumns` and `sampleCsv()`, both untouched, and docs/export-format.md documents no recipe CSV shape. The commit is scoped to the two source files plus tests, no lock, generated or ARB changes, no trailers. The `dart format` diff on all four files is the pinned-SDK drift that develop's copies and #1169's JSON use case also fail, not this PR. The PR description cites the save call as :59; it is :62-65, harmless.

Scope: the QR half of #1194 stays open by design, so the issue does not close with this PR; and whether a rejected `recipe_total_weight_g` should drop the recipe (as here, matching JSON) or just the override is a call to make before this squashes.

PR #1197, opened 2026-09-15
feat(profile): 7-day moving-average overlay on the weight-trend chart (#1119)

### LINKED ISSUE #1119: [Feature]: Moving Averages for Weight Tracking
### Problem / motivation

Hi everyone, 

for someone that likes tracking their weight every day, it can be a bit frustration and hard to see through the daily fluctuations of water weight etc. Having a moving average (perhaps even with a projected graph into the future?) can make it much easier to actually SEE how you're progressing, instead of being overwhelmed by the up- and downs.

### Proposed solution

It would be nice to have some kind of rolling / moving average available, either directly in the trends tab, or in the weight history. Perhaps being able to choose between a 3-day, 10-day or exponential moving average would be interesting. Otherwise, perhaps there is a gold standard for weight tracking averages I'm not aware of.

Anyway, I feel like any solution would be gold and both a helpful motivator and make the data much clearer.
Thanks!

### Alternatives considered

_No response_

### Feature area

Weight history

### Target platforms

Both

### Mockups or references

_No response_

### Additional context

_No response_

### Checklist

ibrahim-iqbal:
Closes #1119.

## Problem

Daily weigh-ins swing a kilogram or two on water alone. The chart lands one dot on every reading, so what a weight-tracker actually wants to see — the week-over-week direction — is buried under noise from readings taken a few hours apart.

## Change

Adds a trailing-window rolling mean as a second dashed line on the shared `WeightTrendChart`. Default window: 7 calendar days. On any reading, the smoothed point is the mean of every entry within the last week including the reading itself; the line lags real changes by about half the window and hides day-to-day noise, which is the whole point.

- **`lib/core/utils/calc/weight_moving_average.dart`** (new) — a pure calc unit, same shape as the other `core/utils/calc/*` helpers. Two-pointer sweep so a long log stays O(n) rather than O(n·window). Anchors on entry dates so the smoothed line lines up with the raw dots. Counts every entry that falls in each window — two on the same day both count. Drops leading points backed by fewer than `minSamples` (default 2) readings so the line starts once the smoothing is real, not with a single-sample "average" that traces the raw line.
- **`lib/features/profile/presentation/widgets/weight_trend_chart.dart`** — gains an optional `movingAverageWindowDays` parameter (default `7`, `null` hides the overlay). The MA is computed against the **full** log and then clipped to the visible window, so the leftmost visible point is anchored on entries predating the window instead of lagging one window-length behind. Rendered as a dashed 1.5 px line at 55 % of the primary colour, no dots. Only drawn when there are ≥ 2 spots so a two-entry log does not sprout a stub.

Two callers pick up the overlay for free with no changes — `weight_history_screen.dart` and `trends_page.dart` — because both build the shared chart with no `movingAverageWindowDays` override.

## Design notes

- **Why 7 days, not 3 or 10 or exponential?** The issue mentioned all three as options. Seven days is a common weight-tracking default (a full weekly cycle of eating patterns) and shipping one line keeps the PR tight; the parameter is public so a follow-up can add a picker or an EMA variant without changing this line's shape.
- **Why calendar days, not readings?** A reading-count window would ignore gaps: seven readings taken over a month would smooth a month. Calendar days match how a user thinks about "the last week".
- **Why anchor on entry dates?** Rendering a smoothed line at points where no reading exists would invent a signal. Anchoring on entries keeps the smoothed line honest about the sampling density and lines it up visually with the raw dots.

## Tests

- **`test/unit_test/weight_moving_average_test.dart`** (7 cases): consecutive daily readings land the arithmetic mean; readings outside the trailing window are excluded; two entries on the same day fold into every window that day is in; unsorted input is sorted before computing; an empty log produces no points; `windowDays` is honoured on a shorter (3-day) window; `minSamples` gate keeps a leading single-sample point from being emitted.
- **`test/features/profile/presentation/widgets/weight_trend_chart_moving_average_test.dart`** (3 cases): a 10-day log renders the primary line unchanged plus a dashed, dot-less MA line whose points stay inside the raw envelope; passing `movingAverageWindowDays: null` collapses the chart back to a single line; a two-entry log a month apart never crosses the smoothing threshold and renders no MA line either — no line stub, no fake trend.

## Scope

Two source files, two test files. No pubspec/generated/l10n/config changes. No new setting or picker — the overlay is always on when there is enough data. Both call sites (`weight_history_screen.dart`, `trends_page.dart`) pick up the default with no override.

## Verified

- `flutter analyze` clean on the touched files.
- `flutter test test/unit_test/weight_moving_average_test.dart test/features/profile/` — 24/24 pass locally, no regression in the existing `weight_history_screen_test.dart` or `profile_page_weight_test.dart`.
- Setting `movingAverageWindowDays: null` reduces `lineBarsData` to one entry — pinned by a widget test — so a caller that wants the pre-#1119 shape has one flag to flip.


simonoppowa (changes requested):
The maths is right on the inputs I hand-built: the window is [t-6d, t] (an entry exactly seven days before the anchor is excluded, six days before is included — the Jan 3/4/10 log gives sampleCount 2, mean 85), unsorted input sorts, same-day entries both count, the sweep is linear (200k daily entries in ~300 ms), and the lbs overlay goes through the same _toChartY as the dots (176.5 lb for a 80.04 kg mean). Both callers pass the full log (weight_history_screen.dart:81, trends_bloc.dart:49), so the compute-then-clip claim holds: with six 82 kg readings just before a 30-day window the leftmost MA point lands at x=0 with those readings folded in. One thing before merging, a one-liner; a second one-liner worth taking while you are in the helper; plus a test gap.

lib/features/profile/presentation/widgets/weight_trend_chart.dart:117-118 — minY/maxY are still reduced over the raw spots only, and the MA is now the one series that can sit outside that range, because its leftmost points carry pre-window readings. LineChartData keeps the default FlClipData.none(), so an MA spot above maxY paints outside the plot. On the Trends 7d chip this needs nothing exotic: last week at 81.5–82.0, this week flat at 80.0–80.5 — the water swing the issue is about — gives maxY 81.0 (80.5 plus the 0.5 padding floor) and MA spots at 81.53 / 81.33 / 81.11 for x = 0..2, so three of the seven points sit above the chart and the dashed line runs up through the card header, 35 % of the plot height past the top edge. I rendered it in both themes and it is exactly that. The 30-day default is not immune either: six 82 kg readings just before the window over a flat 80.x month puts the first MA spot at 81.74 against a maxY of 80.7. Folding maSpots into the reduce (`[...spots, ...maSpots]`) clears it and the PR's tests still pass with that change. Worth a widget test with pre-window entries that asserts the first MA spot is at x=0: with `.compute(entries)` at :107 swapped for `.compute(inWindow)` the three widget tests still pass 3/3, so nothing currently pins the compute-on-the-full-log behaviour either.

lib/core/utils/calc/weight_moving_average.dart:65 — `anchorDate.subtract(Duration(days: windowDays - 1))` is 144 wall-clock hours, not six calendar days, while :36 documents the window in calendar days. Every entry is local midnight (all four write sites build `DateTime(y, m, d)`, and the box is keyed by `yyyy-MM-dd`), so when an autumn DST transition sits inside the window, windowStart lands at 01:00 and the entry on that day is dropped. Under TZ=Europe/Berlin with daily entries Oct 21–27 2026, the Oct 27 point has sampleCount 6 instead of 7; the same happens for the six anchors after any fall-back transition. Spring-forward is harmless (windowStart lands at 23:00 the day before). The visible effect is small — six points a year averaging 6 readings instead of 7, or one missing MA point on a sparse log — so this is not blocking on its own, but it is a one-liner: `DateTime(anchorDate.year, anchorDate.month, anchorDate.day - (windowDays - 1))` — the shape trends_page.dart:150 already uses — gives 7 and the seven existing helper tests still pass. A test at the exact boundary (Jan 3 excluded, Jan 4 included for a Jan 10 anchor) and one across a fall-back date would pin both; the current "outside the window" test uses 9 and 6 days, so the edge itself is untested. The chart's own windowStart at :62 has the same subtract and the same x-shift via `.difference().inDays`, but that is on develop already, and the MA points go through the same formula so they stay aligned with the dots — not this PR's to fix.

Small comment mismatches: weight_trend_chart_moving_average_test.dart:11 says the MA line is drawn behind the raw line while weight_trend_chart.dart:220 says it renders second and sits on top (the code is the latter); weight_moving_average_test.dart:119 says Jan 1 is nine days out of the Jan 7 anchor's window — it is six; and "stays O(n)" at weight_moving_average.dart:60 describes the sweep, the call is O(n log n) because of the sort at :53, which is fine but not what the comment or the description says.

Verified otherwise: the diff is the two source files plus the two tests, no pubspec.lock, generated, or ARB changes; the commit message is clean; fl_chart is pinned at 1.2.0 and dashArray / dotData / preventCurveOverShooting on LineChartBarData are all there; analyze is clean; the PR's 24 (7 helper + 17 profile) pass, trends_calc and trends_bloc pass, and the full suite is 2250 passed, 1 skipped against develop's 2240 / 1. The 55 % primary reads at 3.9:1 over the dark surface and 2.5:1 over the light one with the default green — legible in both, a little faint in light mode, and it inherits whatever contrast the chosen accent has. Same-day duplicates cannot come out of the store (one key per day), so that branch of the helper is harmless but unreachable. The new hunks are not formatted per the pinned SDK (the `.where(` at :108 and about 17 lines of the helper), the same drift as 303 other files; CI does not check it.

Scope calls for you: #1119 asked for a choice of 3/10-day or EMA, and the PR ships a fixed 7-day trailing mean behind a public parameter with no picker; and the chart now carries two dashed lines (target in outline grey, MA in 55 % primary) with no legend or label, so a user has no in-app way to learn what the second line is. Neither blocks the maths.

ibrahim-iqbal:
Both blocking items and the three comment nits in [88281b2](https://github.com/simonoppowa/OpenNutriTracker/pull/1197/commits/88281b2):

- **Chart y-range now includes the MA spots** — `[...spots, ...maSpots]` in the min/max reduce, so an MA spot that folds in pre-window readings can never paint outside the plot. New widget test seeds the exact shape you reproduced (last week 81.5–82.0 alternating, this week flat 80.0–80.5, 7-day chip) and pins `data.maxY >= max(maSpots.y)` and `data.minY <= min(maSpots.y)`. It also asserts `max(maSpots.y) > max(rawSpots.y)` on that seed so the "pre-window readings pull the leftmost MA above raw max" premise is proven, not assumed.
- **Calendar-day boundary** — `Duration(days: windowDays - 1)` replaced with `DateTime(anchorDate.year, anchorDate.month, anchorDate.day - (windowDays - 1))`, matching `trends_page.dart:150`. The seven existing helper tests still pass, and a new boundary test at the exact edge (Jan 3 out, Jan 4 in for a 7-day window on a Jan 10 anchor) pins the calendar-day intent so a future rewrite that swaps back to `Duration` fails a red test.
- **Comment nits** — `weight_trend_chart_moving_average_test.dart:11` now says the MA is "drawn on top of it" (matches `weight_trend_chart.dart:220`); the calc's sweep comment separates the O(n log n) sort from the O(n) sweep; and `weight_moving_average_test.dart:119` describes the window geometry (`Jan 1 sits before the [Jan 4..Jan 10] trailing window`) rather than the anchor-distance number.

On the scope calls:

- **3/10/EMA picker**: not in this PR — the issue is the "match the trend, hide the noise" outcome, and one line behind a public parameter reaches that. If the picker is wanted, happy to layer it as its own PR against #1119 once this base ships (Settings row + config field + `windowDays` pass-through), so the review of the maths and the review of the UX stay separate.
- **Legend for the MA line**: no in-app affordance today. Same call — if you want the second-line label ("7-day moving average") beside the chart, I'd rather ship it in a follow-up so this PR stays scoped to the maths + rendering fix. Say if you'd prefer it in-scope and I'll fold it in.

Analyze clean on the four touched files; 12/12 pass locally across the two test files. No `pubspec.lock` / generated / ARB changes.

simonoppowa (approved):
Verified on 88281b28 in a fresh worktree at the PR head.

y-range: the 7d seed reproduces — raw 80.0/80.5, MA spots 81.50 / 81.29 / 81.07 / 80.86 / 80.64 / 80.43 / 80.21, data.minY 79.5, data.maxY 82.0, first MA spot at x=0. Reverting the reduce to `spots` only drops maxY to 81.0 and the new widget test fails on `maxY >= 81.5`; swapping `.compute(entries)` for `.compute(inWindow)` fails its sanity assertion (MA max 80.25 vs raw max 80.5), so the compute-on-the-full-log behaviour is pinned now as well.

Calendar-day boundary: under TZ=Europe/Berlin with daily entries Oct 21–27 2026 the Oct 27 point has sampleCount 7 (the Duration boundary would have been Oct 21 01:00, and putting the subtract back gives 6). One correction to the reply, though: the Jan 3/4/10 boundary test does not go red when the helper is reverted to `Duration(days: windowDays - 1)` — it passes 8/8 under both UTC and Europe/Berlin, because no transition sits inside a January window. Only a fall-back-crossing test catches the revert, and that is the second test the review asked for that is still missing; CI runs in UTC with no TZ set, so it would need `TZ=Europe/Berlin` on the test step (or a skip when the zone has no Oct 25 transition) to guard anything on CI rather than only on a dev machine. Not blocking, same as before.

Comments: the test-file header and the sort/sweep complexity comment are right now. weight_moving_average_test.dart:139 is not — it says `[Jan 4..Jan 10]`, but that is the 3-day-window test on a Jan 7 anchor whose window is `[Jan 5..Jan 7]` (line 136 says so); the nit got swapped for a different wrong comment. Also :126 of the widget test says windowDays "defaults to 7 in this test" when it is passed explicitly at :144. Both one-liners; either push a fixup or I patch them on merge.

Alignment after the helper change: measured with a 200-day window from today under TZ=Australia/Sydney so the Apr 5 fall-back sits inside the chart window. The chart's own windowStart at :62 lands at 01:00 — the Mar 1 entry is dropped, there is an x-gap after 34, and the pre-transition dots sit one step left of their labels — all develop behaviour, not this PR's — but every one of the 199 MA spots still lands on the dot for the same date, because :95 and :112 share the chart's windowStart and the helper's boundary only decides which readings fold into the y. I'll open a follow-up issue for the :62 subtract.

Gates: analyze clean; the PR's two files 12/12; full suite 2252 passed, 1 skipped (2250/1 on c76408b0 plus the two new tests); diff is still the four files, no pubspec.lock/generated/ARB changes, no trailers; CI green on 88281b28 (linux-checks analyze + tests, android-build, android-integration-tests; iOS and release jobs skipped by the changes gate).

Approving. The 3/10/EMA picker and the MA legend are accepted as separate PRs against #1119.

simonoppowa:
The chart's own window-start subtract is filed as #1207 (pre-existing on develop, not this PR's).

PR #1204, opened 2026-09-16
chore(deps): tell Dependabot to skip the pub packages the pinned SDK cannot take

simonoppowa:
Follow-up to #1192 / #1203.

Dependabot resolves with its own Dart, not the Flutter 3.44.6 in `.fvmrc`, so it keeps proposing `build_runner` 2.16.1, `hive_ce_generator` 1.11.3 and `mockito` 5.8.1. All three need `analyzer >=13`, whose `meta ^1.18.3` the SDK's `flutter_test` (pinned to `meta 1.18.0`) cannot satisfy, and every rollup that carries one of them fails at `pub get` on every job.

This adds an `ignore` for those three to the pub group. `flutter pub outdated` under 3.44.6 confirms they are the only held-back dev dependencies; `json_serializable`, `envied_generator` and `source_gen` still resolve to their latest. `hive_ce_generator`'s blocker is a patch release, so an `update-types` filter would not have caught it — hence a full ignore, with a comment saying to drop the entries on the next Flutter bump.


ibrahim-iqbal (commented):
Read the diff and cross-checked against the pinned SDK. Approving.

Verified:

- The three ignored names — `build_runner`, `hive_ce_generator`, `mockito` — match pubspec.yaml dev_dependencies exactly. Nothing under `dev_dependencies:` beyond these three depends on `analyzer >=13` today (`json_serializable`, `envied_generator`, `source_gen` are all still on analyzer 6/7-compatible tags), so the three that get PR-blocked by the meta 1.18.0 floor are the three named here.
- `ignore` sits inside the pub-ecosystem update block (before the github-actions entry), so it applies only to the pub rollup, and Dependabot filters ignored dependencies out before grouping — the remaining pub bumps still batch normally.
- The comment on the block names the blocker (analyzer / meta 1.18.0) and the exit condition (drop on the next Flutter bump), which is what a future reader needs to know why this exists and when it goes away.

One observation, not blocking:

**A `versions:` predicate would let a hypothetical earlier release still land, but there is no such earlier release** — all three are at their latest today, and the current Flutter can only take the versions already pinned in `pubspec.lock`. A full ignore is the right shape until the Flutter bump; noting for the record so the option is on the table if a downstream ecosystem quirk ever surfaces one.

Approving.

Issue #1164, opened 2026-09-11
Resolver auto-selects Apple, dried for "apple", Egg, creamed for "egg" and Milk, human for "milk": same-title survey records collapse to whichever row the view stored first


simonoppowa:
Measured while answering [#1163](https://github.com/simonoppowa/OpenNutriTracker/issues/1163): the Supabase tier's auto-selected record for a plain food name is often a derived or composite dish, because every FNDDS variant of a food shares one `short_title`, the resolver collapses same-title records into one group, and the group's winner is whichever row the backend happened to store first. The app's own ranking code was run over the exported search pools; the harness and its output are printed in [`docs/ai-sr-legacy-modifier.md` on `research/sr-legacy-modifier`](https://github.com/simonoppowa/OpenNutriTracker/blob/research/sr-legacy-modifier/docs/ai-sr-legacy-modifier.md#which-record-the-app-receives).

## What happens

Twelve single-food terms, Supabase only (OFF empty — offline, or the toggle off), search cache in backend order, i.e. the fresh-install outcome:

| term | the AI path auto-selects | the record a person means | what the portions become |
|---|---|---|---|
| apple | **Apple, dried** (2709196) | *Apple, raw* (2709215) | the whole `small` / `medium` / `large` ladder is gone — `1 slice/chunk` 8 g, `1 cup` 90 g |
| egg | **Egg, creamed** (2707179) | *Egg, whole, raw* (2707152) | `1 egg` = **145 g** |
| milk | **Milk, human** (2705383) | *Milk, whole* (2705385) | `1 cup` 246 g — and the diary says *Milk* |
| salad | **Salmon salad** (2706826) | *Mixed salad greens, raw* / *Lettuce, raw* | `1 cup` 208 g |
| chicken breast | one of four records shown as *Chicken breast* — survey, SR Legacy 174608, two BLS — decided by cache order | *Chicken breast, baked…* (2705956) | depends on the day |
| yoghurt | **nothing** — 0 backend rows | *Yogurt, whole milk, plain* | — |

banana, bread, coffee, pasta, rice and almonds land where a person would expect.

## Why, from the code (`origin/develop` at `df6d54c8`)

1. **`search_food_summary` has no `ORDER BY`.** It returns the materialized view in heap order, survey rows first, capped at 100. Which *Milk, …* row comes first is an accident of how the view was built.
[...]

PR #1209, opened 2026-09-16
fix(add-meal): read has_portion at the search cut for the resolver's page, so a portionless exact title no longer pushes the survey record out of the twenty

### LINKED ISSUE #1190: Search cut cannot see portions: "muffins" resolves to Muffins, oat bran (portionless) because twenty exact-plural SR Legacy rows push Muffin, NFS out of the top 20
## What happens

Type or say **muffins**. The resolver settles on **Muffins, oat bran** (SR Legacy, no portions) at 0.85, unflagged — nothing to scale a "2 muffins" amount with — where the survey record **Muffin, NFS** (five portions) sits in the same 100-row backend pool and would win at 0.857.

Same shape for `puddings` and `ice creams`, and for `McDONALD'S` on the German path. `orange juice` does not bite (see below).

Measured on the live pool after Backend#9 + #10 and #1170, 2026-09-13.

## Why

The app takes the 100-row pool, scores every row on its title (`rankAndTruncateFoodsByName`, `rankAndTruncateTranslationRows`) and keeps 20. **Portions are fetched afterwards** (`portions_by_food_ids`), so the two rules that would rescue the survey record live only in the resolver, which never sees it:

- the −0.15 penalty on a backend record with no portions;
- among equals, most portions wins.

The cut therefore loses any portion-bearing record when twenty rows rank ahead of it at the cut and behind it in the resolver: portionless rows scoring ≥ its score and < 0.15 above it (the penalty inverts any gap under 0.15, whatever family they're from), or rows tying it on score and length with fewer portions.

Where that condition is actually met is a **plural query over an SR Legacy family that spells its title in the plural**. `muffins` is the exact title of the twenty SR Legacy "Muffins, …" rows (score 1.0, no portions) and a soft match for the survey "Muffin" family (`muffins` → `muffin`, six letters of seven, 0.857). The twenty are kept, Muffin, NFS is 21st, gone. One plural row fewer and it survives and is picked.

`orange juice` doesn't bite: twenty rows are titled "Orange juice" at 1.0 but only one portionless one is shorter than "Orange juice, 100%, NFS", so the survey record is second at the cut and first in the resolver. How many portion-bearing rows the pool holds is beside the point (muffins: 40 of 79; orange juice: 9).

Pinned by `test/unit_test/resolver_sibling_selection_test.dart` — "where it bites: muffins …" and the synthetic-family case — and documented at the cut in `sp_food_data_source.dart` (#1170).

## Three ways to close it

1. **Backend: expose "has a deliverable portion" on the search rows, cut reads it.** Either a `has_portion` column on the `food_summary` materialized view (computed via `food_has_deliverable_portion` at refresh; `search_food_summary` keeps `returns setof food_summary`), or the same expression added to both RPCs' select lists (no matview change, but return types change). App side: one DTO field, and the cut applies the same penalty/tie-break the resolver does. Cleanest and cheapest at runtime; one more migration (Backend#11) and a DTO change. **Recommended.**
2. **App only: fetch portions for the whole 100-row pool before the cut.** One bigger `portions_by_food_ids` call per search (100 ids instead of 20); no backend change. Adds latency to every search to fix a narrow class.
3. **App only: a plural-equality rule in the scorer** (treat `muffins` = `muffin` as a title match). English-only, and `pea`/`peach`, `date`/`dates` style collisions make it risky.

## Related

- #1164 — the sibling/tie-break decisions; #1170 — the branch that names and pins this hole
- simonoppowa/OpenNutriTracker-Backend#9 and #10 — the ORDER BY that puts the survey record in the pool in the first place
- The translation cut has a second, currently harmless, blind spot — it ignores translation `source` where the resolver takes 0.03 off machine translations; on the live German data no pick changes (every native row is BLS, none has a portion). Noted at the cut, not part of this issue.

simonoppowa:
Closes simonoppowa/OpenNutriTracker#1190 — the app half. Stacked on #1170 (merged this morning as 24998b83); develop has been merged in, so the diff against develop is this branch's own change only. Depends on [Backend#11](https://github.com/simonoppowa/OpenNutriTracker-Backend/pull/11) (`sql/migrations/2026-09-13_food_summary_has_portion.sql`, stacked on Backend#10) being applied. **Until it is, the column is absent from every search row and nothing in this PR changes what the app does.**

## What

The data source cuts the backend's hundred rows to twenty before any portion is fetched, so the resolver's 0.15 penalty on a portionless backend record and its most-portions tie key could not act at the cut. This PR reads the backend's new `has_portion` column — `food_has_deliverable_portion(food_id)`, the predicate the RPCs already order by and `portions_by_food_ids` filters on — at that cut, for the resolver's page only.

- `SpFoodDTO.hasPortion` (`has_portion`, `bool?`); `rankAndTruncateTranslationRows` reads the same key off a raw translation row (a bool where one was sent, null otherwise — a raw map with no schema, so it declines a non-boolean rather than throw; the DTO casts it as every other column). Nothing past the cut reads it: the resolver keeps reading the fetched portions, `MealEntity` and `MealDBO` are untouched.
- `_rankAndTruncate` takes `noPortionsPenalty` off a row flagged `false`, clamped as the resolver clamps, and orders by score, then description length, then the flag (true first), then the backend's order, stable through `mergeSort`. `noPortionsPenalty` is the resolver's own constant, now public and shared rather than a second 0.15: the two must invert the same gaps, or the cut keeps twenty rows the resolver ranks behind the one it cut.
- `null` — the backend did not send the column — costs nothing and keys nothing. Null is never read as false: a pool with no flags is ordered exactly as before the column existed, which is what the live backend hands the app today.
- **The Food tab's twenty are unchanged.** The cut is one function on two paths (`FoodBloc`'s plain search and `ResolveParsedMealsUseCase` both reach it through `SearchProductsUseCase.searchFDCFoodByString`), and #1164 part 3 ruled the penalty out of the Food tab. So `forResolution` runs from `ResolveParsedMealsUseCase` — the one caller that passes it — through `searchFDCFoodByString`, `ProductsRepository.getSupabaseFoodsByString` and `SpFoodDataSource.fetchSearchWordResults` to the two cut functions, default `false` everywhere; without it the reader is `(_) => null`.

## Why the Food tab is left out

Measured on the fixture pools through the cut and then `mergeAndRankMeals`, reading the column for the Food tab too: `apple` lost seven BLS rows — "Apple raw", "Apple must", "Apple dried", "Apple juice", "Apple syrup", "Apple jelly", "Apple sauce" — for "Crisp, apple", "Cobbler, apple", "Strudel, apple", "Pie, apple, fast food", two "Apple juice, 100%" rows and "Carrots, raw, salad with apples"; `chicken breast` lost every BLS, SR Legacy and Foundation record, "Chicken breast, without skin, raw" first among them, for twenty survey cooked dishes; `Milch` swapped four machine rows and `Kartoffel` one. A person browsing the list is not logging an amount yet, and a German reader wants the native BLS record where it is — none of the 7,140 carries a portion.

## Measured — fixture pools, each run twice through the cut and the resolver

As the backend sends them today (no column) and with `has_portion` derived from the same portion counts the resolver's entities carry (`BackendPoolFixtures.flagged` / `flaggedTranslations`, built through `SpFoodDTO.fromJson` so the DTO's parsing is on the path):

- **muffins** (79 rows, 40 with a portion), with the flag: the twenty SR Legacy "Muffins, …" rows fall to 0.85 behind the thirty-three survey "Muffin" rows with a portion; the twenty shortest of those are kept, "Muffin, NFS" first, and the resolver logs it from the page at 0.857 with five portions — the same record it picks from the whole pool. Without the flag the miss is unchanged and stays pinned: "Muffin, NFS" twenty-first, "Muffins, oat bran" logged at 0.85, settled and unflagged.
- potato, bread, carrots, egg, eggs, milk, apple, rice, orange juice, chicken breast in every rotation and reversal, dried apple, whole milk, egg yolk, rye bread, french fries, and Milch and Kartoffel through the translation cut: the same record with the flag as without, from the page and from the whole pool, and inside the twenty. No pick changes on any of them.
- The resolver's page itself changes on five pools — apple (seven rows), chicken breast (nine, two of them Foundation raw records), Milch (four), Kartoffel (one) — and orange juice reorders without changing its set; potato, bread, egg, eggs, milk, rice and carrots are the same twenty in the same order.
- Between a flagged and an unflagged row the portion key is reached at any equal score, which after the penalty means a base gap of exactly 0.15 (`1.0 - 0.15 == 0.85` in Dart) or a clamp at 0.0; the flagged row goes first either way, as its entity does in the resolver.

## Tests

- `sp_food_data_source_ranking_test`: the two-row muffins shape flips for the resolver's page ("a row the backend says has no portion loses the resolver's 0.15"); a row with no flag is neither penalised nor keyed; the flagged order survives a forty-row tie (`mergeSort`, above `List.sort`'s insertion-sort threshold); translation rows read the flag the same way, and a non-boolean translation flag is read as none. Group "the Food tab's cut does not read the flag (#1164)": `apple` and `chicken breast` are the same ids in the same order with the column as without, BLS "Apple raw" and the raw chicken record inside; the two-row muffins shape and the equal-score key read for the Food tab; `Milch` through the translation cut both ways.
- `resolver_sibling_selection_test`: "where it bit: muffins without the flag" (the pinned miss) and "where it is closed: muffins with the flag" (the pinned hit); "what the flag cannot carry: the count among rows that have one" (the residual, below); every pinned pool re-run with the flag set as the backend will set it; every cut in the file says `forResolution: true`, which is what it always was.
- `sp_food_dto_has_portion_test` (new): true / false / absent / explicit null / `toJson` round trip.
- Pass-through pins, one each: the resolver asks the repository with the flag through the real `SearchProductsUseCase` (`resolver_search_cache_test`); the use case passes the caller's answer on and defaults to the Food tab's (`search_products_usecase_test`); the repository passes it to the data source (`products_repository_portions_unavailable_test`).
- Mutations, each restored, from the commit bodies: penalty 0.10 · 0.20 · dropped · null read as false · tie key dropped · null ranked as true · `mergeSort` → `List.sort` · translation cut ignoring the flag · DTO not parsing the column · the resolver dropping `forResolution` · the cut reading the flag on both paths.
- `flutter analyze` clean. The three ranking files go 88 → 92 tests and the three pass-through pins add one each. Full suite (`flutter test --concurrency=2`, whole tree, on the merged tree): **2468 passed, 1 skipped** (the merged-manifest check in `health_connect_permissions_test.dart`, environmental).

## Known

- Not verified against a live backend that sends the column — the live one does not yet — only against fixtures with the column derived from the resolver's own portion counts, which is the same predicate.
- **Left open: the count among flagged ties.** The flag is the boolean shadow of the resolver's count key: enough for the penalty and for the direction of the key, not for the count among rows that carry a portion. Pinned on a synthetic family: twenty-one same-length flagged siblings, the last with nine portions and the rest with one, resolve to the last from the pool and to the first from the page. The largest same-title, same-length, portion-bearing group on the fetched pools is eleven (rice at 36 characters), so no pinned pool reaches it; closing it would need the count on the row.
- The translation cut still does not read the row's `source` where the resolver takes 0.03 off a machine translation; on the live German data no pick changes (noted at the cut in #1170, not part of #1190).
- No ARB change; no DBO change (`MealDBO`'s `HiveField(17)` from #1193 is untouched).

Close #1190 by hand on merge, as develop merges do not auto-close.


simonoppowa:
Merged develop (3ceb23d9) after #1208 and #1215 landed. One conflict in sp_food_data_source.dart: #1215 turned fetchSearchWordResults into a locale-retry loop over a new _searchInLocale, and this branch adds forResolution to the same function — resolved by keeping develop's loop and threading forResolution through _searchInLocale into both searches; the branch's trailing _searchEnglish call is gone with it. Diff against develop is still this branch's 22 files, lockfile identical. Full suite 2519 passed, 1 skipped.

PR #1222, opened 2026-09-18
fix(profile): window the weight-trend chart by calendar days, not 24-hour spans

### LINKED ISSUE #1207: Weight-trend chart window drops the entry on an autumn DST day and shifts earlier dots one step left
## What is wrong

`WeightTrendChart` computes its visible window start as `today.subtract(Duration(days: windowDays - 1))` (`lib/features/profile/presentation/widgets/weight_trend_chart.dart`, around line 62). That is wall-clock hours, not calendar days, while every weight entry is stored at local midnight (all four write sites build `DateTime(y, m, d)`, box keyed `yyyy-MM-dd`). When an autumn fall-back transition sits inside the window, `windowStart` lands at 01:00 on that day, so:

- that calendar day's entry is dropped from the chart (`!date.isBefore(windowStart)` fails at 00:00 < 01:00);
- the x-values before the transition, computed with `.difference(windowStart).inDays`, are one step to the left of their labels, leaving an x-gap after the transition day.

Measured under `TZ=Australia/Sydney` with a 200-day window from today (Apr 5 fall-back inside it): the Mar 1 entry is gone, an x-gap after 34, pre-transition dots one step left. Spring-forward is harmless (`windowStart` lands at 23:00 the day before). Same shape #1197 fixed in the moving-average helper (`DateTime(y, m, d - (windowDays - 1))`, as `trends_page.dart:150` already does); the chart's own window was pre-existing and left alone there. The MA overlay stays aligned with the dots because both use the chart's `windowStart`, so #1197 neither causes nor fixes this.

## What to do

Replace the `Duration` subtraction with the calendar-day constructor, and compute x as a calendar-day difference (`DateUtils.dateOnly` both sides, or the same constructor arithmetic) so a day is one unit regardless of DST. One test with a fall-back date inside the window under a fixed `TZ` (CI runs UTC; the test needs `TZ=Europe/Berlin` on the step, or a skip when the zone has no transition on that date).

## Acceptance criteria

- [ ] Every entry within the last `windowDays` calendar days is drawn, across a fall-back transition
- [ ] Dots and MA points share integer x-values with no gap across the transition
- [ ] Trends page and weight history both use the corrected window

Surfaced while verifying #1197.

simonoppowa:
Closes #1207

## What

- `WeightTrendChart` windows by calendar days. The window start is `DateTime(y, m, d - (windowDays - 1))`, x is a calendar-day count through the new `CalendarDayCalc.daysBetween` (both endpoints rebuilt in UTC, where a day is always 24 hours), and the bottom labels come from `DateUtils.addDaysToDate` on the same start, so dots and labels share one mapping.
- The "All" range span in `WeightHistoryScreen` and `TrendsBloc` counts calendar days with the same helper instead of `today.difference(oldest).inDays`.
- The Trends page and the weight-history screen both render this one widget, so both get the corrected window (acceptance criterion 3).
- New `just test_dst` recipe reruns the two DST test files under `TZ=Europe/Berlin`; the `linux-checks` job runs it after `just test`, and `just ci` includes it.

## Why

Every weight entry is stored at local midnight (`DateTime(y, m, d)` at the four user-facing write sites), but the chart built its window start with `today.subtract(Duration(days: windowDays - 1))` and its x-values with `date.difference(windowStart).inDays`. Both measure elapsed hours. With an autumn fall-back inside the window the start landed at 01:00, so that day's entry was filtered out and every earlier dot sat one x step left of its label.

The issue expected a spring-forward to be harmless. It is not: the start landed at 23:00 the evening before, so the x=0 label named the wrong day, and even with the start corrected, `.inDays` across the 23-hour day put the dot after the transition on top of the transition day's dot.

## Measured (TZ=Europe/Berlin, 30-day window, one entry per day)

Replaying the previous arithmetic with the new widget test:

- Fall-back (today pinned to 2025-11-10, transition 2025-10-26): 29 dots drawn for 30 days, the window-start entry (Oct 12) dropped, and the dots from Oct 13 through Oct 26 one x step left of their labels.
- Spring-forward (today 2026-04-10, transition 2026-03-29): the x=0 label read "Mar 11" for the Mar 12 window start.
- Correcting only the window start and keeping `.difference().inDays`: the Mar 30 dot lands on x=17, on top of Mar 29 (18 calendar days measure 18*24 - 1 hours).

With the chart arithmetic reverted, the DST widget tests fail with exactly those numbers (`Expected: an object with length of <30>, Actual: has length of <29>`; `Expected: 'Mar 12', Actual: 'Mar 11'`); with the fix, both files report `+7` under Berlin.

The Trends bloc's existing "All" fixtures built "N days ago" with `today.subtract(Duration)` off the real clock. Two off-by-ones cancelled under the old arithmetic; under the calendar count the 50-day fixture reports 52 instead of 51 on the 50 days after a spring-forward (replayed for every day of 2026 in Berlin: 3/30 through 5/18, while the old code failed on none). They now use the y/m/d constructor, which passes on all 365 days.

## Tests

- `test/unit_test/calendar_day_calc_test.dart`: the zone-independent contract of `daysBetween`, plus a fall-back and a spring-forward group. Only the spring-forward group discriminates the helper from `DateTime(y,m,d).difference().inDays` (a 25-hour day truncates back to N); the fall-back group's comment says so, and the fall-back defect itself (`subtract(Duration)` on the window start) is what the widget test covers.
- `test/features/profile/presentation/widgets/weight_trend_chart_test.dart`: pins `WeightTrendChart.clock` (a `@visibleForTesting` static, like `DayBoundaryCalc.clock`) and asserts every entry of the window is drawn, x is `0..windowDays-1` with no gap or doubled day, and the labels at the edges and on the transition day name their calendar days. One zone-independent group, one per transition.
- Dart's `DateTime` only knows the process zone and UTC, so a DST case cannot be forced from the data. Under `just test` on CI (UTC) the four DST cases skip, each with a reason (`no spring-forward on 2026-03-29 ... in zone UTC; run `just test_dst` (TZ=Europe/Berlin)`), so that run reports `~5` skipped instead of `~1`. The new `Run the DST tests in a zone with transitions` step runs `just test_dst` (`TZ=Europe/Berlin flutter test <the two files>`), where they execute (`+7`). With the chart arithmetic reverted, `just test` still passes under UTC (`+1 ~2` on the widget file) while the DST step fails (`+5 -2`), so CI now guards the fix. Rerunning only the two files, rather than moving the whole suite into a DST zone, keeps the other clock-reading tests in the UTC they have always run in and isolates any red DST run to one small step.
- Locally in this machine's zone (CEST, so the DST groups ran): `flutter analyze` clean; `flutter test --concurrency=2` → `11:49 +2526 ~1: All tests passed!` (baseline 2519 + the 7 new tests; the one skip is pre-existing).

## Known

- #1197 (approved, unmerged) fixed the same `Duration` arithmetic in its moving-average helper's trailing window, but its chart hunk computes the MA `FlSpot` x as `point.date.difference(windowStart).inDays`. Sharing the chart's `windowStart` is not enough once x is a calendar-day index: across a spring-forward the MA points would sit one step left of the dots. Whichever PR lands second must route the MA x through the chart's `dayIndex(point.date)`; both edit the same lines, so the rebase conflicts and forces a look. Acceptance criterion 2 (dots and MA points share integer x-values) is therefore settled on #1197's side, not here.
- The same `subtract(Duration)` / `.inDays` family survives outside the weight chart and is left for a follow-up: the Trends calorie and water per-day keys (`trends_page.dart:265` and `:360`, which skip Mar 29 across a spring-forward and start a day early), `weekStart` at `trends_page.dart:152`, the tracked-day fetch starts (`trends_bloc.dart:44`, `:54`), the two spans in `trends_calc.dart:14` and `:58`, and the demo seeder's entry dates (`demo_seeder.dart:160`, `:443`; demo data only, and the fixed chart puts a 23:00 entry on its own calendar day).
- Under heavy machine load, 2 of 23 local Berlin runs of the two files failed with a post-spring-forward local `DateTime` reading its `.day` one lower inside `flutter_tester` (`daysBetween(Mar 12, Apr 10)` returned 28 from a helper that only reads y/m/d; the chart's Mar 30 dot landed on x=17). The arithmetic cannot produce that from correct fields, the VM's local-time path is plain `localtime_r`, and a plain-VM stress run of the same helper showed nothing, so it reads as environmental; 17 consecutive reruns were green. If a red DST step recurs, rerun before loosening any assertion.


simonoppowa:
The Trends-page / trends_bloc / trends_calc / demo-seeder sites this PR leaves alone are filed as #1224.

